PR #695, opened 2015-11-26
privacy setting for sound/pack downloads

soycode:
Not tested as I don't see an obvious script to run a dev environment and don't have the time at the moment to figure it out. Should be a fairly simple change, though - the one thing I'm uncertain of is if the `hide_downloads` field needs to be populated elsewhere (e.g. some user data structure), and perhaps given a default ("false" is fine by me, but I just figure it's nice to give users the option).

Let me know your thoughts, thanks!

### DIFF
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -245,7 +245,12 @@ class ProfileForm(forms.ModelForm):
         label = "",
         required=False
     )
-    
+    hide_downloads = forms.BooleanField(
+        help_text="Hide sounds/packs downloaded on your Profile page",
+        label = "",
+        required=False
+    )
+
     def __init__(self, request, *args, **kwargs):
         self.request = request
         super(ProfileForm, self).__init__(*args, **kwargs)
--- a/templates/accounts/account.html
+++ b/templates/accounts/account.html
@@ -127,8 +127,10 @@
                 {% if user.profile.num_posts %}
                 <li>Number of forum posts: {{user.profile.num_posts}}</li>
                 {% endif %}
+                {% if not user.profile.hide_downloads %}
                 <li><a href="{% url user-downloaded-sounds user.username %}">Sounds downloaded</a> by {{user.username}}</li>
                 <li><a href="{% url user-downloaded-packs user.username %}">Packs downloaded</a> by {{user.username}}</li>
+                {% endif %}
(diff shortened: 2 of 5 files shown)


bdejong:
Thank you so much for your pul request. However... we consider all data on freesound to be free/open, this also includes the downloaded sounds and packs. Everything except private messages and users' email addresses are open on freesound.

From a technical point of view I had a look at your PR and it doesn't look like this is storing anything in the database for the profile. So, your PR doesn't really do what you might think it does ;-)

soycode:
Re: free/open, I wholeheartedly agree that open source and data is great, but would ask you to reconsider whether people's download records are part of that. You don't openly post complete server/clickstream logs, and I'd say that download records are essentially a high-signal subset of that. It's fine if people *want* to share, but it's potentially identifying information (to others who know/profile your tastes, be they friends or ads networks) that some may prefer keeping private. Perhaps it makes sense to still let the sound uploader know who downloaded (or at least how many), but just putting it out there for everyone without any options seems a bit much (it's also indexed by search engines so it's not just public but quite discoverable).

Re: technical, yes as I said I imagine I'm missing populating the field in some data structure, but there wasn't a clear way to run a dev environment so I decided to just send the more obvious changes to start the conversation. It may be nice to add a `CONTRIBUTING.md` file with some instructions on how to get started, regardless of your final thoughts on this pull request.

If after further consideration you stand by your decision that download logs should be publicly discoverable then I respect that - the culture of your project is up to you. I definitely appreciate the practical and philosophical difficulties in balancing freedom/openness with privacy - I just feel that, in this particular case, it'd be nice to at least give users the option to not be *that* open (you can of course keep the default as "open", which will probably steer most users that way anyway). All actual data of value (the sounds themselves, their total # of downloads, etc.) can and should still be open.

bdejong:
Hi! This is not the first time we discuss this particular data. We have discussed it on both the freesound forum and internally with the freesound crew. And... we stand by this cultural decision.

Regarding getting things up and running, yeah, we know, there is a lot of information missing there.

I actually already talked to the core guys about this, but I used this this opportunity to make a ticket regarding this: https://github.com/MTG/freesound/issues/696

soycode:
Thanks for your response - I'm glad the issue has been considered. I also appreciate the ticket for a getting started/contributing guide.

PR #700, opened 2016-01-21
Make syntax of audio processing compatible with python3

bretonium:


alastair:
Hi!
Thanks for the PR. Can you explain why you want to be able to run this file (and only this file) in python 3? We're definitely interested in improving py3 support for freesound, but it's quite a large project and we're not sure where to start. This could be a good place though.

One small change - 
Can you make sure you add 

    from __future__ import print_function

to the top of the file, as the syntax in python 2 is currently incorrect.

bretonium:
Hi,

> Can you explain why you want to be able to run this file (and only this file) in python 3?

Because we use this file in GNU MediaGoblin, https://mediagoblin.org. Right now we have an effort to make it py3-compatible.

> as the syntax in python 2 is currently incorrect.

it is correct for python2.7, but if you support python2.6, then OK, I will add it.

bretonium:
In fact, it works fine without the import you mention even on python2.6:

```
breton ~ % python2.6
Python 2.6.8 (unknown, Jan 29 2013, 10:05:44) 
[GCC 4.7.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> print(2)
2
>>> print('hello world')
hello world
>>>
```
Do you still want me to add the import?

alastair:
Thanks for the reply. The behaviour of `print` with brackets on python 2.x is a little tricky. In this case it's still treating it as a statement with an expression rather than a function. See http://stackoverflow.com/questions/12162629/using-print-in-python2-x for another description.

With the added import I'll be happy to merge this.

PR #708, opened 2016-02-11
{% url %} templatetag quoting from future, accounts edition

alastair:
This is one of the biggest changes we need to make for django 1.5, adding quotes to the url templatetags in django. I'll commit the templates for each app separately, so that someone can double-check them in smaller batches that I didn't miss, e.g. a missing quote, or change any words.

I'm not sure what the best way to deploy this is. One idea would be to update django at the same time, make all the template changes, and also every other change that has to be made.
An alternative is to use the future tag: `{% load url from future %}`. This is valid in django 1.4. We could add it to all templates, bit by bit, deploy with the current django, and then when we make the upgrade to 1.5, remove the future import.  Thoughts?

### DIFF
--- a/templates/accounts/account.html
+++ b/templates/accounts/account.html
@@ -2,6 +2,7 @@
 
 {% load display_sound %}
 {% load tags %}
+{% load url from future %}
 
 {% block title %}{% if home %}home{% else %}{{user.username}}{% endif %}{% endblock %}
 
--- a/templates/accounts/account.html
+++ b/templates/accounts/account.html
@@ -69,22 +69,22 @@
 
         <div class="account_information_and_home_options">
         <ul>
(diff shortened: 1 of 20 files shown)


ffont:
Because we'll probably need to start changing templates before upgrading django (for design changes etc) I think it makes sense to import url from future and make it work with that.

alastair:
Good idea, let's make this a smaller PR then, and just work per-module for now.

(Although, my hope is that we could upgrade before the design change!)

alastair:
I've added another commit to load url from future, now it should also work in 1.4. I'll do a test.
This is only for the accounts module, if we're happy with working like this we can either do a PR per module, or just do commits without a PR.

alastair:
I've verified all the changed tags in this PR. I think it's OK to merge.

PR #718, opened 2016-04-13
{% url %} quotes, forum & geotags edition

alastair:


alastair:
oops, I incorrectly pushed this to the wrong branch, though the number of changes in each commit is quite small. I'll leave it

PR #734, opened 2016-04-16
Support for multiple files upload and drag-and-drop

giuband:
Added the following things:

1. Support for multiple file uploads (requires a new handy python package)
2. Support for drag and drop
3. Added a upload progress bar

All of them on the HTML5 version.

### DIFF
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -33,7 +33,7 @@ from utils.spam import is_spam
 def validate_file_extension(audiofiles):
     for file_ in audiofiles:
         content_type = file_.content_type
-        if not content_type.startswith("audio") or not filename_has_valid_extension(str(file_)):
+        if not content_type.startswith("audio") and not filename_has_valid_extension(str(file_)):
             raise forms.ValidationError('Uploaded file format not supported or not an audio file.')
 
 
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -24,18 +24,21 @@ from django.contrib.auth.forms import PasswordResetForm, UNUSABLE_PASSWORD, Auth
 from django.utils.translation import ugettext as _
 from django.utils.safestring import mark_safe
 from django.core.urlresolvers import reverse
+from multiupload.fields import MultiFileField
 from accounts.models import Profile
 from utils.forms import RecaptchaForm, HtmlCleaningCharField, filename_has_valid_extension
 from utils.spam import is_spam
 
 
-def validate_file_extension(value):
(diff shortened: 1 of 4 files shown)


accounts/forms.py
```diff
@@ -24,18 +24,21 @@
...
 from utils.spam import is_spam
 
 
-def validate_file_extension(value):
-    if not filename_has_valid_extension(str(value)):
-        raise forms.ValidationError('Uploaded file format not supported or not an audio file.')
+def validate_file_extension(audiofiles):
+    for file_ in audiofiles:
+        content_type = file_.content_type
+        if not content_type.startswith("audio") and not filename_has_valid_extension(str(file_)):
```
bdejong: I find and/not combos always rather confusion. Do you mean:

```
if not (content_type.startswith("audio") or filename_has_valid_extension(str(file_))):
```

or rather

```
if not (content_type.startswith("audio") and filename_has_valid_extension(str(file_))):
```

accounts/views.py
```diff
@@ -832,23 +832,25 @@ def upload_file(request):
...
             if form.is_valid():
-                if handle_uploaded_file(request.user.id, request.FILES["file"]):
-                    uploaded_file = request.FILES["file"]
-                    success = True
-                else:
-                    error = True
+                submitted_files = request.FILES.getlist('files')
+                for file_ in submitted_files:
+                    if handle_uploaded_file(request.user.id, file_):
+                        uploaded_file = file_
```
bdejong: If there are multiple uploads, uploaded_file will only contain the last one...

bdejong:
@ffont how difficult is it right now to

- run automated tests on branches/forks in github
- get freesound to run on a local machine 😄 

@giuband COOL addition! Very minor comments... I suppose all the JS magic is just ... correct 😃

accounts/forms.py
giuband: The one I mean is the second one you propose, and you're right that is a bit confusing (actually I just realized it is wrong).

templates/accounts/upload.html
```diff
@@ -155,11 +166,84 @@
...
+    <!--<![endif]-->
+</object>
+<p>
+<img src="{{media_url}}/images/fugue-icons/icons/exclamation.png"> <span style="color:black">Don't leave before upload finishes!</span><br/>
+Hint: you can upload multiple files at once</p>
+{% endif %}
+
+</div>
+
+<div id="describe_button_block">
```
alastair: formatting?

Issue #751, opened 2016-09-08
Revise pack download functionality


ffont:
When packs are downloaded a zip file is created on the fly which includes sounds plus a license file (with a list of sounds and their licenses). If I recall correctly, on osx that zip file needs to include some checksum so that it can be opened correctly. For this reason we have to create the license files beforehand every time a pack is edited, which creates a number of other problems.
We should investigate if we can create these files on the fly and avoid having them precomputed. Eventually this will render Pack.process() method useless and we'll be able to get rid of this (which will simplify many bits of fs code).

andrebola:
After we talked I changed the code to generate a temp file with the license information. I decided to use python's tempfile module to handle the names of the temp files so there is no collision. Then, my idea is to use a cron to remove the files older than one week. Like this: 
```
find /path/to/files* -mtime +7 -exec rm {} \; 
```

Also, as @ffont suggested I made the method generic so it can work with any set of sounds, but I'm not sure about the text in the license file  :
https://github.com/MTG/freesound/blob/temp-license-file/templates/sounds/pack_attribution.txt

andrebola:
Following the suggestion of @alastair I re-implemented this issue. Now it doesn't need to use the filesystem since nginx queries a url to get the license file content.

ffont:
@andrebola can this be turned into a pull request?

ffont:
Closing this issue as all was implemented and deployed :)

PR #757, opened 2016-09-22
Pack download using temporary license file

andrebola:
According to #751, when the user downloads a pack of sounds we create a temporary file with the license information so we can remove the process method in Pack.

### DIFF
--- a/apiv2/views.py
+++ b/apiv2/views.py
@@ -37,6 +37,7 @@ from geotags.models import GeoTag
 from bookmarks.models import Bookmark, BookmarkCategory
 from accounts.views import handle_uploaded_file, send_activation
 from accounts.forms import RegistrationForm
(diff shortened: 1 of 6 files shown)


bdejong:
Idea for a small improvement: as we have pack descriptions, should we not
add them to the license file?

ffont:
I was testing this pull request but I don't seem to be able to make it work. 
I deployed in freesound test and it returns 502 responses when trying to download a pack (i.e., http://test.freesound.org/people/xserra/packs/5103/).
How did you test it?

andrebola:
Now is working, it was using the wrong path for the license file (it should use the '/secret/...') and there was a file permission issue.

ffont:
Now I tested and it looks fine.
Once we deploy we should monitor cpu impact of this feature and test with very large packs.

PR #758, opened 2016-09-30
Add flag of parental advisory on sounds

andrebola:
issue #692
The user can set if he wants to see explicit content or not. When the user is making a search, if the sound is explicit we will show it blurry. If the user enters to the sound it will show a message which
will block the page.


ffont:
Was this closing accidental?

andrebola:
My idea was to close the pull request while I'm cleaning my fork (some of this commits are not needed) so Travis doesn't compile and test every change I do. 
But I realize now that is not possible to reopen the PR in this case so I'll have to open a new one.

alastair:
You may be able to edit the PR and change the base, which may let you reopen it.  Did someone force-push to master?

andrebola:
I did force-push to master of my fork, that is the problem. Now the PR can't be reopened, changing the base didn't work.

PR #762, opened 2016-10-18
Add flag of parental advisory on sounds

andrebola:
issue #692
The user can set if he wants to see explicit content or not. When the user is making a search, if the sound is explicit we will show it blurry. If the user enters to the sound it will show a message which
will block the page.

Also fixing bug when a user renames a pack with the same name of a deleted Pack


ffont:
I've been testing the pull requests, here are some comments for improvement:

- Blurred sound players can still be clicked. I think you should add 'pointer-events:none' to .sample_player_small.blur
- The idea is not that if you have not checked the "I'm an adult" checkbox in settings you can't access sounds marked as explicit, the idea is that if you have not checked "I'm an adult" you see the sounds blurred along with a button to show the sound (like "this sound contains explicit content, are you sure you want to display it?"). If users mark the "I'm an adult" checkbox in settings then sounds are never blurred. Therefore, you should add the button to the sound player (and the sound page).
- Considering the previous point I think we can male "I'm an adult" false by default.
- I think the sound page could be blurred to similarly to the sound player. I like the warning icon you show and the extra explanation text. I'd put that on top of the blurred content and also show the button.

andrebola:
@ffont now the user can remove the message and the default value for is_adult is False. 
Do you think we should change the behavior when the user is not logged in? now we don't display the message for this case.

ffont:
I added some more styling to the pull request. Changes are in https://github.com/MTG/freesound/tree/andrebola-master.

Branch is now deployed in Freesound test and will be merged to master once tested by other developers.

Closing this pull request now as development moved to the aforementioned branch.

Issue #761, opened 2016-10-18
Report to users when they have duplicate accounts


alastair:
For email-login we want unique email addresses for all users.

Instead of automatically merging accounts, we have decided to ask users what to do, because they may have specific reasons to have separate accounts.

We need a new table, `SameAccounts`, which links accounts which share an email address.
Columns:
- core user id (account logged into most recently)
- other account id
- other account orig email

Change the email address of the account with the earliest `last_login` time from `email@provider.com` to `email%provider.com@freesound.org`, then we can finish the email login functionality, with a unique index on the email column.

When a user logs in, check if any of their accounts are in this table. If so, redirect them directly to a page explaining that they have multiple accounts and list the accounts. Put a check on Home as well, showing a banner which links to this page.

Tell them that they have three options:
1. Change the email address of their accounts
2. Merge them
3. Delete them
- If they log in to the account with the changed email address, tell them that they can change it -> send to the change form.
- If they log in to the account with the original email address, tell them to log in as the other one (give the username). If they don't know the password, tell them to contact support and provide a link with a message already filled in

If they want to merge or delete the accounts, tell them to contact support and provide a link with a message filled in.

When a user changes their email address, we need to remove them from the `SameAccounts` table if they were there

ffont:
If many users ask us to delete or merge accounts, this might be some extra work for us. I agree with all said but I'd try at least not to present change email/delete/merge as options at the same level. In general I'd tell users they have to change the email and that's it.
Then I'd say something like "If you want to get rid of one of the accounts or merge them, please contact us at ...".
Well maybe this comment looks obvious but just wanted to make that point.

bdejong:
I would say: let us give ONLY people the option to change or delete, not merge.

alastair:
90% of the accounts in this situation only have downloads on one of the accounts. Merging them is easy.

There are ~30 accounts with many uploaded sounds + tickets + comments + messages, etc. If we have to merge only one of these (which we already have to do, from our previous emails to users), then we have to write the code, therefore we will have code for the general merge case as well.

To me, the change email case only makes sense for people who have two distinct accounts and actively participate using both of them. I think that delete and merge will be the most useful actions.

qubodup:
GMail specific information:

martha@gmail.com will receive emails sent to martha+freesound@gmail.com

This is described on https://support.google.com/mail/answer/12096?hl=en

[...]

PR #763, opened 2016-10-21
Tell users when they have more than one account with the same email (#761)

alastair:
Creates a list of user accounts with shared email addresses and if a user is in this list when they log in, redirect them to an info page


templates/accounts/multi_email_cleanup.html
```diff
@@ -0,0 +1,10 @@
+{% extends "accounts/_section.html" %}
+
+
+{% block title %}edit account{% endblock %}
+
+{% block section_content %}
+
+<h1>Emailc leanup</h1>
```
ffont: <h1>Emailc leanup</h1> -> <h1>Email cleanup</h1>
;)

ffont:
@alastair what's the status of this? if we finish that we'll be able to close ~2-3 tickets ;)

alastair:
I've not done anything since I made these initial changes. We still need to:
 - give users a page to update this information (change email, merge, delete accounts)
 - use the mapping table when sending users an email from any location, to make sure we use their real email, and not the fake @freesound.org one that we gave them

ffont:
I have been updating this PR to finish it.
Continuing with what @alastair  did but simplifying a bit the remaining things to do, the thing now works like:

- Once the PR is applied and deployed a DB migration will be done which will search for users with duplicate emails and store them in a table (this is part of alastair's code). When we find 2 or more users that share the same email, the email of the most recent user is kept and the emails of the others are changed.
- Every time that a user logs in, we check if his email was changed because of being shared (he should be in the table). If that is the case, we show a message that says "you should update your email address" and point to the email reset form.
- If a user logs in whose email was changed because of being shared but her email is now different than the one we automatically set, this means that the user saw the warning and reseted her email. In this case we delete the corresponding objects in the duplicate emails table and redirect the user.

accounts/migrations/0010_auto_20170510_0945.py
```diff
@@ -0,0 +1,26 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('accounts', '0009_sameuser'),
+    ]
+
+    # Add unique constraint to 'email' field in auth_user table
+    operations = [
+        migrations.RunPython(replace_empty_email_addresses, migrations.RunPython.noop),
+        migrations.RunSQL('ALTER TABLE "auth_user" ADD CONSTRAINT "auth_user_email_uniq" UNIQUE ("email");'),
```
ffont: We should make this constraint take the lower or upper of email, as when we do validate emails we check iexact.
This means we should also make sure that we have no duplicates like a@b.com and A@B.com.

ffont:
When closing this, we should close https://github.com/MTG/freesound/issues/761 as well.

utils/mail.py
```diff
@@ -21,9 +21,39 @@
 from django.conf import settings
 from django.core.mail.message import EmailMessage
 from django.template.loader import render_to_string
-from django.core.mail import get_connection, EmailMultiAlternatives
+from django.core.mail import get_connection
 
 
+def generate_tmp_email(email):
```
alastair: Can we make the name of this method clearer to know what it does?

accounts/tests.py
```diff
@@ -669,3 +670,79 @@ def test_user_delete_remove_sounds(self):
...
+    def test_redirects_when_shared_emails(self):
+
+        # Try to log-in with user and go to messages page (any login_required page would work)
+        # User a is not in same users table, so redirect should be plain and simple to messages
+        resp = self.client.post(reverse('login'),
+                                {'username': self.user_a, 'password': '12345', 'next': reverse('messages')})
+        self.assertRedirects(resp, reverse('messages'))
+
+        # Now try with user_b. User b had a shared email with user_c, but user_b's email was not changed
+        # Redirect should again be plain and simple
```
alastair: Is this right? My thought is that we should do a redirect here if the user is in the `SameUsers` table, regardless of if they are as the main or second user

accounts/tests.py
```diff
@@ -669,3 +670,79 @@ def test_user_delete_remove_sounds(self):
...
+        self.assertRedirects(resp, reverse('messages'))
+
+        # Now user_c. User c had her email changed, therefore will be redirected to the email cleanup page
+        # instead of messages
+        resp = self.client.post(reverse('login'),
+                                {'username': self.user_c, 'password': '12345', 'next': reverse('messages')})
+        self.assertRedirects(resp, reverse('accounts-multi-email-cleanup') + '?next=%s' % reverse('messages'))
+
+        # Now user_c changes his email and tries again, redirect should go to email cleanup page and from there
+        # directly to messages (2 redirect steps)
```
alastair: This can be a separate test

accounts/tests.py
```diff
@@ -669,3 +670,79 @@ def test_user_delete_remove_sounds(self):
...
+            secondary_orig_email=self.user_b.email,  # Must be same email (original)
+        )
+        # User a never had problems with email
+        # User b and c had the same email, but user_c's was automaitcally changed to avoid duplicates
+
+    def test_redirects_when_shared_emails(self):
+
+        # Try to log-in with user and go to messages page (any login_required page would work)
+        # User a is not in same users table, so redirect should be plain and simple to messages
+        resp = self.client.post(reverse('login'),
```
alastair: Perhaps a comment here to say that we POST to login because the email check logic is done in the login view - this is why we can't use `self.client.login`

accounts/views.py
```diff
@@ -77,6 +77,71 @@
...
+            # If the logged has an email shared with other accounts we figure out if we changed his email or we changed
+            # the emails from the other accounts which shared it
+            original_email = same_user_qs.first().main_orig_email
+            if request.user.email == original_email:
+                # If this account's email was not changed, simply continue with the login redirect (user won't notice)
+                return response
+            else:
+                # If we changed this account's email, we advise the user to chose a new email address
+                return HttpResponseRedirect(reverse("accounts-multi-email-cleanup")
+                                            + '?next=%s' % request.POST.get('next'))
```
alastair: Is it possible that there is no `next` (e.g. we just click the login button?). If so, should we have `?next=None`, or skip this parameter?

accounts/tests.py
```diff
@@ -669,3 +670,79 @@ def test_user_delete_remove_sounds(self):
...
+
+        # Now user_c changes his email and tries again, redirect should go to email cleanup page and from there
+        # directly to messages (2 redirect steps)
+        self.user_c.email = 'new@email.com'  # Must be different than generate_tmp_email('c@d.com')
+        self.user_c.save()
+        resp = self.client.post(reverse('login'), follow=True,
+                                data={'username': self.user_c, 'password': '12345', 'next': reverse('messages')})
+        self.assertEquals(resp.redirect_chain[0][0],
+                          reverse('accounts-multi-email-cleanup') + '?next=%s' % reverse('messages'))
+        self.assertEquals(resp.redirect_chain[1][0], reverse('messages'))
```
alastair: I think that this test would be easier to write and reason about if the logic which checked emails and deleted `SameUser` objects was in a separate function, then we could just test the function.

accounts/views.py
```diff
@@ -77,6 +77,71 @@
...
+    same_user_qs = request.user.profile.has_many_emails(return_qs=True)
+    if not same_user_qs.exists():
+        # User should have never come here
+        return HttpResponseRedirect(reverse('accounts-home'))
+
+    original_email = same_user_qs.first().main_orig_email
+    main_user = same_user_qs.first().main_user  # i.e. user that kept the original email
+    other_users_with_same_email = set()
+
+    if request.user == main_user:
```
alastair: I think this is wrong - we should show this page if there is a entry for this user anywhere in `SameUser`, not just if they have the rewritten email.

accounts/views.py
```diff
@@ -77,6 +77,71 @@
...
+        # Main user should not be redirected here
+        return HttpResponseRedirect(reverse('accounts-home'))
+
+    # At this point, the current logged in user must be one of the users whose email
+    # was changed. We check if the user has changed his email again, and if he did we
+    # simply show a "your email problems have been fixed" message, otherwise we show
```
alastair: How do you show this message? I don't see anything which adds one.

templates/accounts/multi_email_cleanup.html
```diff
@@ -0,0 +1,28 @@
+{% extends "accounts/_section.html" %}
+
+
+{% block title %}edit account{% endblock %}
+
+{% block section_content %}
+
+    <h1>Your email address should be updated</h1>
+
+    <p>Dear <b>{{ request.user.username }}</b>, we recently introduced a new pocily in Freesound which does not allow
```
alastair: policy

templates/accounts/multi_email_cleanup.html
```diff
@@ -0,0 +1,28 @@
...
+        different user accounts to share a single email address. Only a few users are affected by this change, and you
+        happen to be one of them. But don't worry, the fix is easy ;)</p>
+
+    <p>All what you have to do is to either set a new email using the <a href="{% url 'accounts-email-reset' %}">reset
+        email form</a>, or submit a <a href="{% url 'contact' %}">support request</a> telling us your username
+        and the new email you want to use for your account.</p>
```
alastair: give a hint about gmail + addresses - link to https://gmail.googleblog.com/2008/03/2-hidden-ways-to-get-more-from-your.html perhaps?

templates/accounts/multi_email_cleanup.html
```diff
@@ -0,0 +1,28 @@
...
+
+    <p>FYI, your original email address was <b>{{ original_email }}</b>, and was used by the following accounts:</p>
```
alastair: For your information, the email address which your accounts are using is ..., and is used by the following accounts

templates/accounts/multi_email_cleanup.html
```diff
@@ -0,0 +1,28 @@
...
+    <ul>
+    {% for user in other_users %}
+        <li><a href="{% url 'account' user.username %}">{{ user.username }}</a></li>
+    {% endfor %}
+    </ul>
+    <p>Currently, <b>{{ original_email }}</b> is only used by <a href="{% url 'account' user_kept_original_email.username %}">
```
alastair: This block doesn't make sense to me - as far we are concerned there is only one email address which this user has.

accounts/forms.py
```diff
@@ -277,6 +277,15 @@ def __init__(self, *args, **kwargs):
         self.user = kwargs.pop('user', None)
         super(EmailResetForm, self).__init__(*args, **kwargs)
 
+    def clean_email(self):
+        email = self.cleaned_data["email"]
+        try:
+            User.objects.get(email__iexact=email)
```
ffont: That's a good idea, I extended Django's `UserManager` by adding this method.
https://github.com/MTG/freesound/pull/763/commits/50a73c5e3d6d0a049eec45f4a8f4a0122f59d1bf

accounts/forms.py
```diff
@@ -277,6 +277,15 @@ def __init__(self, *args, **kwargs):
...
+            raise forms.ValidationError(_("A user using that email address already exists."))
```
ffont: As discussed in the office, for now we'll just log when this happens and we'll decide which further action to take. https://github.com/MTG/freesound/pull/763/commits/f2329fb67f040fcbb3e67f31765ffff399766ede

accounts/migrations/0010_auto_20170510_0945.py
```diff
@@ -0,0 +1,26 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def replace_empty_email_addresses(apps, schema_editor):
+    User = apps.get_model("auth", "User")
+    users = User.objects.filter(email="")
+    for user in users:
+        user.email = '%s@freesound.org' % user.username
```
ffont: Done!
https://github.com/MTG/freesound/pull/763/commits/a20dcbfc4f25dd6ee1c3169e84018a9dc982831a

accounts/migrations/0010_auto_20170510_0945.py
ffont: I made the constraint on upper (https://github.com/MTG/freesound/pull/763/commits/cc6692e89cae7b8ab05f2654ab73b0f797909d76)
There is already an index on email upper.

accounts/models.py
```diff
@@ -85,6 +85,15 @@ class Profile(SocialModel):
     def __unicode__(self):
         return self.user.username
 
+    def has_many_emails(self, return_qs=False):
+        """Check if the user account associated with this profile
```
ffont: Now it always returns queryset. I think in your initial implementation it returned boolean, I added the qs option at some point and I guess later modified some of your original code so the boolean option was in fact not used anymore. I know the name of the method is not perfect if it returns a qs, but I think it makes the code readable so I've kept it.
https://github.com/MTG/freesound/pull/763/commits/2f4f6718dc9246d17e9ee4576fa7caefa0883879

utils/mail.py
ffont: Could be better but I think it's not too bad and there is a docstring which gives more details.
Unless you have an amazing name proposal I'll keep this one ;)

accounts/tests.py
ffont: OK, https://github.com/MTG/freesound/pull/763/commits/41e4b93a8aee152835e1d1799e4fd0aed942e7b6

accounts/tests.py
ffont: Well not really because users whose email was not changed are not "affected".
If account A and account B share the same email, we left the email untouched for the account with the most recent "last logged in" date. The email was only changed for the others.

accounts/tests.py
ffont: I like it together as it is showing the different possibilities in a single test. If someone is trying to understand the logic I think this helps.

accounts/tests.py
ffont: Yeah maybe, but still that sort of "integration" test would be needed in my opinion. The logic for the whole thing is all commented in the `multi_email_cleanup` function because of its potential complexity so I think it is ok there.

accounts/views.py
ffont: Only adding `next` if it is present: https://github.com/MTG/freesound/pull/763/commits/1ee3dae8cfbb9d44d87d19eb0a80c1f69f6ad0b1

accounts/views.py
ffont: Well this is part of the logic I thought we agreed on.
For me if the user is not directly affected there is not need to "bother" them.
For example think of the case:
- Person A creates account A
- After some time person A forgets about account A and creates account B (with same email)
- Person A now only uses account B
- We change the email of account A but keep intact account B (as user last logged in with account B)
- account A is never used again and we live happily ever after

accounts/views.py
ffont: I think the message is not really needed, I updated the comment: https://github.com/MTG/freesound/pull/763/commits/1133118c19bf09a07df6a6bded22d87d83d238b4

templates/accounts/multi_email_cleanup.html
ffont: https://github.com/MTG/freesound/pull/763/commits/f01e000719b55466bb078a9164d8954fe02c391e

templates/accounts/multi_email_cleanup.html
ffont: We could automatically detect if they use gmail and give the hint, but I would say let them do it themselves.

ffont:
I implemented a number of changes as agreed on Friday with @alastair.
The changes include a simplification in the logic of the whole thing.
This is possible because we realised that a single email is never shared with more than 2 accounts.

In the current approach, if a user logs in which is affected by the email issue, he is redirected to the email cleanup page (regardless of whether he logged in with one affected account or the other). By default we change the email in the secondary account (the one with oldest last log in), but user does not need to know about that. In the email cleanup page user is requested to change the email of one of the two accounts. When users change one of the emails, the original email is set to the other account.

ffont:
The contents of this PR have been merged into #863. I'm closing this PR without merging as I'll do the merge in #863.

Issue #759, opened 2016-10-03
Provide short, official, ASCII-only user and sound URLs for easier attribution


qubodup:
Freesound allows all kinds of characters in usernames. This can become a problem when trying to provide attribution in a restrictive or buggy system to create a credits screen for a video or in a video game engine, or a video host website or restrictive (no special characters, no unicode, max character count) content generators (website, video...).

It always becomes a huge problem when a lot of sounds are used.

Please reserve the following URL spaces:

freesound.org/s/[sound_id]
freesound.org/u/[user_id]

Visiting these pages will **redirect** to the full URL of the sound or user page. This is how every website with official short URLs does it. Example: https://youtu.be/A9CZLSaJhl4

YouTube messed with its user URL system so often, it is a mess. Each of the following URLs provides the same content without redirecting to a "true" URL.
https://www.youtube.com/qubodup
https://www.youtube.com/c/qubodup
https://www.youtube.com/user/qubodup
https://www.youtube.com/channel/UCGy7ppDyzQ4JX9WsSfkN6ow

What I suggest for freesound is clean: there is one short url and it redirects to the full one, just like the sounds or youtube video pages.

sound ID can be the current sound ID or a hash of it to make it even more compact.
user ID can be the internal user ID or a hash of it to make it even more compact. They can _not_ be a hash of the username, since username changes are common https://github.com/MTG/freesound/issues/706

I recommend against hashes, because they might cause additional headaches to make them language-safe: http://programmers.stackexchange.com/questions/143405/how-to-generate-language-safe-uuids

Making it easier to link to users and their sounds will benefit all.

ffont:
I think what you propose makes sense.
Let's if the others have something to say.

ffont:
Related issue (if not duplicated): https://github.com/MTG/freesound/issues/705

ffont:
Done and deployed in https://github.com/MTG/freesound/pull/766

Issue #705, opened 2016-02-04
Provide short, independent URLs for each sound, use them for credits


qubodup:
Sound URLs are long:

http://freesound.org/people/qubodup/sounds/332061/

But YouTube descriptions (for credits) and Tweets are limited in length and long URLs are extremely unappealing when it comes to appearance in credits of video games and film credit rolls.
- Longer URLs have the benefit of informing you of the author.
- Longer URLs (currently) have the drawback of breaking once a user changes their name (this is possibly by manual admin action, happens sometimes when user requests change). I have seen this happen multiple times.

What does it look like on other pages?
- https://soundcloud.com/qubodup/fable
- https://youtube.com/watch?v=Eam0ukCRKKk aka https://youtu.be/Eam0ukCRKKk
- http://qubodup.deviantart.com/art/Free-Pinball-575624409 aka http://fav.me/d9ipmmx
- https://vimeo.com/55279993

I suggest that
1. Freesound provides short URLs for each sound
   This URL should be either
   - http://freesound.org/s/332061 (far from ideal - 10% to be exact (2 out of 22 characters without "http://"))
   - http://freesound.org/332061 (best solution without using custom domain)
   - http://CUSTOMDOMAIN/332061 (best solution but might cause a lot of organizational overhead initially and a little bit of overhead yearly. Domain name suggestion shared on developer mailing list for safety, search for "custom short domain suggestion" to find the right thread).
2. Freesound replaces the URLs used in
   - http://freesound.org/home/attribution/?format=plaintext
   - http://freesound.org/home/attribution/?format=html
   - http://freesound.org/home/attribution/?format=regular
3. Freesound adds a "share" UI element similar to that on deviantart and youtube: a text input field that contains the short url, which auto-selects the text content on first click:
   ![share widget idea](https://cloud.githubusercontent.com/assets/18312/12805740/fb8ba7fc-cafe-11e5-86dc-68ee56fa6878.png)

alastair:
It should be pretty easy to support `/s/songid` or `/songid`. I don't think we need to change to alphanumeric ids.

It's worth pointing out that urls are automatically shortened on twitter, and so don't count towards the tweet length. Perhaps Youtube does the same thing?

We were talking in person, and discussed that it's nice to include the username in urls as another method of citing the sound creator, so I would recommend keeping it if possible. How about 
  * http://freesound.org/user/12343

qubodup:
> it's nice to include the username in urls as another method of citing the sound creator

[...]

PR #766, opened 2016-11-18
Add short links for access sounds

andrebola:
Added short link only for sounds, the link is displayed on sound's page.

issue #759
issue #705


qubodup:
the link is displayed on sound's page
I seem to be unable to find neither short user nor sound urls (I can confirm short sound urls appear to be working though) :)

ffont:
It might happen that you didn't see short sound urls for some sounds due to cache. You should see them now. The short user links we did not implement because user ids are never part of our "public" interface so we thought we could survive with current user urls.

Issue #754, opened 2016-09-14
Async user delete option


ffont:
We should implement the option of deleting users in an async manner, so that admins can schedule deletion of a user.
This could be implemented similarly to the whitelist_user option, using a gearman async task. It should support three methods for deleting users (full db delete, delete including sounds, delete keep sounds).

ffont:
We could also add a fourth option which is to delete only the sounds of a user.
Might become handy in some situations.

andrebola:
I think it would be better to have this fourth option in a separe method since is not related with deleting the user. 
This method should be very clear on what is doing since is not possible to undo the changes.

ffont:
I see that we're using more and more async actions and in the way we're doing it we need to run a different gearman queue for each one. Maybe we should simply have one queue for the async actions like whitelist and delete and then indicate the particular thing to do using the "action" parameter. This also applies to the notify random sound.

Therefore instead of merging this branch I'd refactor the whole async tasks thing (well, processing and analysis we can keep them as they are, with a separate queue). Shouldn't be too much. Also this will mean that we don't need to change deploy config every time that we add new async tasks.

andrebola:
I think I can change the worker to use a single queue, without having to change the code that calls the worker. Otherwise we would have to change the code several parts.

ffont:
We have implemented the delete functions and the infrastructure to do the delete asynchronously. We still need to add the links in the admin page. However, it would be good to add some functionality that shows what is going to be deleted in a delete page confirmation (just like in normal delete from django admin). Maybe the same thing can be reused? we should investigate.

andrebola:
This changes are read on #788

ffont:
Closing this issue as this was already implemented and deployed

PR #769, opened 2016-11-29
Notify random sound

andrebola:


sounds/views.py
```diff
@@ -73,6 +74,9 @@ def get_random_sound():
     if not random_sound:
         random_sound = Sound.objects.random()
         cache.set(cache_key, random_sound, 60*60*24)
+        gm_client = gearman.GearmanClient(settings.GEARMAN_JOB_SERVERS)
```
alastair: It won't happen often, but this will also send an email if we clear the cache

general/management/commands/gm_worker_async_tasks.py
```diff
@@ -90,3 +92,17 @@ def task_whitelist_user(self, gearman_worker, gearman_job):
                 count_done = count_done + 1
             self.write_stdout("Finished processing one ticket, %d remaining" % (len(tickets)-count_done))
         return 'true' if len(tickets) == count_done else 'false'
+
+    def task_email_random_sound(self, gearman_worker, gearman_job):
+        self.write_stdout("Notifying user of random sound of the day")
+        random_sound_id = gearman_job.data
+        random_sound = Sound.objects.get(id=random_sound_id)
+
+        send_mail_template(u'Random sound of the day.',
```
alastair: don't need a . in the subject. Maybe something a little more exciting?  "One of your sounds has been chosen as random sound of the day!"

general/management/commands/gm_worker_async_tasks.py
```diff
@@ -90,3 +92,17 @@ def task_whitelist_user(self, gearman_worker, gearman_job):
                 count_done = count_done + 1
             self.write_stdout("Finished processing one ticket, %d remaining" % (len(tickets)-count_done))
         return 'true' if len(tickets) == count_done else 'false'
+
+    def task_email_random_sound(self, gearman_worker, gearman_job):
+        self.write_stdout("Notifying user of random sound of the day")
```
alastair: Are we happy to send an email to any user regardless of how old the sound is? I guess the idea here is to engage people, including those who haven't used FS for a while.
Some situations which might happen:
 - user's email doesn't work any more
 - sound was uploaded 5 years ago and the user forgot that they even added it
 - the random sound is a goodsounds track, or something from one of the sample libraries [congratulations, one of your 4000 sounds has been chosen as random. F# is a great note!]

I realise that this means considering how the random sound selection works. Are we happy with how it currently works?

ffont: I would only care for the 3rd point, but I'd argue this is not a problem of this issue but more of the general concept of random sound of the day. modularsamples will get one email per week, as he uploaded 1/7 of the whole freesound contents. I think we should maybe add some constrains to the way we choose the random sound, but for now I think this PR is ok.

ffont:
Similarly to my comment in https://github.com/MTG/freesound/issues/754#issuecomment-264404432, I think we should refactor the way we handle the async tasks. I'd have one "async job runner" and then pass the details of the task that needs to be processed instead of having separate queues for all possible async tasks. I'd rather prefer to have such a refactoring instead of merging this particular PR. It should be relatively simple to do this refactoring.

Issue #660, opened 2015-04-23
Add "buttons" for admins and mods.


bdejong:
We (admins) and the moderators could use a little automatic button on their profile or in the forum threads. That way their opinion weighs a little more...

ffont:
I'm not sure I understand what that means. Can you elaborate more @bdejong ?
Maybe I've just been reviewing to many old issues...

bdejong:
This was about having a "badge" for forum mods, admins, sound mods so that
when we post something on the forum it says under our names that we are
"special" users.

Still a good idea!

bdejong:
Closing issue in favour of #784

PR #774, opened 2016-12-27
Donations

andrebola:
After the donation is complete paypal communicates this information to
the server.- To make this work on test server we need to white list paypal
for authentication.

If settings.debug is true we communicate with paypal's sandbox service,
otherwise we communicate with paypal production service.


ffont:
I tried this locally but I don't seem to be able to see the donate widget.
I set the donations goal in `local_settings.py`. What should I do to see the widget?
Also, is `PAYPAL_EMAIL` supposed to be our Paypal email? Is this correctly set?

Also the cache for the front page donations widget, was it so long before this change? https://github.com/MTG/freesound/blob/f334044869bee8bf5b807a8efe58012b728544aa/general/management/commands/create_front_page_caches.py#L47
I don't understand now why do we have such a long number here (unless this is invalidated every time there is a donation).

andrebola:
yes, PAYPAL_EMAIL is the email used for paypal. 
First you need to generate the cache using the command : https://github.com/MTG/freesound/blob/master/general/management/commands/create_front_page_caches.py

ffont:
Still I can't see the widget (after running command).
I guess cache might behave differently in local.
I could however check the donate page and it seems to work.
However, how are we supposed to test a donation? Does paypal suggest some way to test?

Thoughts/issues/ideas:
- https://github.com/MTG/freesound/blob/master/freesound/urls.py#L144 this will have to point to /home/donate/ (the new donation url)
- I don't seem to see any list of donors except those shown in the donate page (last X). I think we should have a dedicated page for that (like freesound.org/donors)
- We can probably pre-populate donations table with data from pledgie. It allows to export donors from the last campaigns (since ~2011).

andrebola:
I'll make the changes you suggested.

For testing, if debug is enabled we use paypal's sandbox, I have some accounts to make donations on the sandbox (I will give you the password).

bdejong:
I would put this in a new subfolder rather than mixing it in with Account... Why not make a separate application within the Freesound folder? It seems kind of messy to merge it with Account.

accounts/tests.py
```diff
@@ -555,3 +555,21 @@ def test_user_delete_remove_sounds(self):
...
+                'payer_email': 'fs@freesound.org',
+                'custom': 'Anonymous',
+                'mc_currency': 'USD',
+                'mc_gross': '1.00'}
+
+        with mock.patch('accounts.views.requests') as mock_requests:
+            mock_response = mock.Mock(text='VERIFIED')
+            mock_requests.post.return_value = mock_response
+            resp = self.client.post(reverse('donation-complete'), params)
+        self.assertEqual(resp.status_code, 200)
```
bdejong: something weird happened with the indentations here

accounts/views.py
```diff
@@ -1069,3 +1065,33 @@ def pending(request):
     }
     tvars.update(paginate(request, pendings, settings.SOUNDS_PENDING_MODERATION_PER_PAGE))
     return render(request, 'accounts/pending.html', tvars)
+
+
+@csrf_exempt
+def donation_complete(request):
+    params = {'cmd': '_notify-validate'}
+    for key, value in request.POST.items():
+        params[key] =  value
```
bdejong: This seems a bit scary: you are passing absolutely all POST parameters to the paypal API. Wouldn't it make more sense to just pick the ones that are needed?

accounts/views.py
```diff
@@ -1069,3 +1065,33 @@ def pending(request):
...
+    return HttpResponse("OK")
+
+
+def donate(request):
+    tvars = {'paypal_email': settings.PAYPAL_EMAIL}
+    return render(request, 'accounts/donate.html', tvars)
+
+
+def donors(request):
+    donations = Donation.objects.all()[:10]
```
bdejong: I thought we said we'd like there to be a way for donators to not show their name...
Also, why choose 10 of them?

accounts/views.py
```diff
@@ -1069,3 +1065,33 @@ def pending(request):
...
+@csrf_exempt
+def donation_complete(request):
+    params = {'cmd': '_notify-validate'}
+    for key, value in request.POST.items():
+        params[key] =  value
+    if settings.DEBUG:
+        paypal_validation_url = "https://www.sandbox.paypal.com/cgi-bin/webscr"
+    else:
+        paypal_validation_url = "https://www.paypal.com/cgi-bin/webscr"
+    r = requests.post(paypal_validation_url, data=params)
```
bdejong: single char parameters are kind of unpythonic

accounts/views.py
```diff
@@ -1069,3 +1065,33 @@ def pending(request):
...
+        Donation.objects.get_or_create(transaction_id=params['txn_id'], defaults={
+        'email': params['payer_email'],
+        'display_name': params['custom'],
+        'amount': params['mc_gross'],
+        'currency': params['mc_currency']})
+    return HttpResponse("OK")
+
+
+def donate(request):
+    tvars = {'paypal_email': settings.PAYPAL_EMAIL}
```
bdejong: "tvars"?

general/management/commands/create_front_page_caches.py
```diff
@@ -33,10 +35,15 @@ def handle(self, **options):
...
+        donations_date = settings.DONATIONS_DATE
+
         rss_cache = render_to_string('rss_cache.html', locals())
         cache.set("rss_cache", rss_cache, 2592000) # 30 days cache
 
-        pledgie_cache = render_to_string('pledgie_cache.html', locals())
-        cache.set("pledgie_cache", pledgie_cache, 2592000) # 30 days cache
\ No newline at end of file
+        donations = accounts.models.Donation.objects\
+                .filter(created__gt=donations_date).all().aggregate(Sum('amount'))
```
bdejong: This doesn't make a lot of sense to me. All donations which were done after the date of the donation? So we'll never be able to look back? Wouldn't it make more sense to create a Model for DonationCampaigns somehow and attach people to a particular campaign? That way we could also create new campaigns in the Admin rather than having to refresh the front page cache.
Also, careful: paypal lets people donate in various currencies, which could mean that you are summing over multiple currencies (EUR/$/...) which will make this break.

templates/accounts/donate.html
```diff
@@ -0,0 +1,36 @@
...
+        <form name="_xclick" action="https://www.sandbox.paypal.com/yt/cgi-bin/webscr" method="post">
+      {% else %}
+        <form name="_xclick" action="https://www.paypal.com/yt/cgi-bin/webscr" method="post">
+      {% endif %}
+          <input type="hidden" name="cmd" value="_donations">
+          <input type="hidden" name="currency_code" value="EUR">
+          <input type="hidden" name="business" value="{{ paypal_email }}">
+          <input type="hidden" name="item_name" value="Freesound">
+          <label>Name</label><input type="text" name="custom" value="Anonymous">
+          <input type="hidden" name="notify_url" value="{{ request.scheme }}://{{ request.get_host }}{% url 'donation-complete' %}">
```
bdejong: I think we have a registered abs_url template tag for getting the absolute URL from a view.

templates/accounts/donors.html
```diff
@@ -0,0 +1,26 @@
...
+
+{% block content %}
+<div id="content_full"> <!-- START of Content area -->
+
+  <h1>Donations</h1>
+   <h2>Last Donations</h2>
+    <div class="donations-list">
+    {% for donation in donations %}
+      <div class="donation-item">
+        <div class="donation-item-left"><img src="/media/images/70x70_avatar.png"></div>
```
bdejong: could we display the avatar of the user?

accounts/models.py
```diff
@@ -260,3 +260,12 @@ def create_user_profile(sender, instance, created, **kwargs):
         profile.save()
 
 post_save.connect(create_user_profile, sender=User)
+
+
+class Donation(models.Model):
+    email = models.CharField(max_length=255)
```
bdejong: Could we not link directly to the profile that is doing the donation? I.e. pass the Freesound user ID to Paypal when the donation is being done... We could still ask the user if she prefers to be anonymous or not.

accounts/views.py
ffont: ``tvars`` is the name we decided to use for the dictionary of variables passed to templates. When we did the refactoring of accounts and sounds module (and probably some others) and stopped using ``render_to_template`` in favour of ``render``, we started using ``tvars`` as the dictionary that is almost always built before calling ``render`` and passed as a parameter. Before we were most of the times using ``locals()`` or simply building the dictionary in line with the ``render_to_template``call.

accounts/views.py
ffont: It might be that donation objects can be anonymous or not anonymous, so the name might appear as anonymous. Is this the case Andrés?
We should probably paginate this. I'm not a big fan of Django's class-based views, but if we simply want to show a paginated list of donors this can probably be done in less than five lines of code using a [ListView](https://docs.djangoproject.com/en/1.10/ref/class-based-views/generic-display/#listview) or something like this.

general/management/commands/create_front_page_caches.py
ffont: I agree that it will be more flexible if we have a DonationCampaign object. We didn't think of it in the first place to keep it simple, but it would not complicate things too much and would be useful. We could then have campaigns about very specific things also if we want.

templates/accounts/donate.html
ffont: We we do: https://github.com/MTG/freesound/blob/4e4b91fa45adabb5294487eeed384596425fd8bb/general/templatetags/absurl.py

accounts/views.py
andrebola: This view listens to a notification made from paypal when someone makes a donation, here we re-send all the data to paypal for validating it. 
There are around 30 parameters and I don't this is an issue.

accounts/views.py
andrebola: yes, the user chooses the name that he want to display. ('Anonymous' by default)

general/management/commands/create_front_page_caches.py
andrebola: My first implementation was using a model with campaigns, which I changed for this simplified version. I will change it now.
About currencies, in the donations form I specify the currency I want to receive the donation (I think you also can set this in the configuration of the account). So if the account of the donor is in other currency paypal makes a conversion.

accounts/models.py
andrebola: Now we allow to donate without been logged in and we don't link the donation to the profile, this email is the one returned by paypal after the donation. Right now I only pass display_name to paypal. The only problem is that there's only one parameter to pass extra data so if we want to pass also the id I would have to encode both attributes together.
(first I wanted to know if this was needed)

templates/accounts/donors.html
andrebola: Yes, if we pass the id of the user to paypal. (see last comment)

accounts/tests.py
andrebola: I think it's fine, the assert is outside of the with block.

andrebola:
I added the script to Drive for importing donations from pledgie. The script is called "import_donors.py".

accounts/tests.py
bdejong: Mmm, I disagree: If the client.post fails silently, resp might be None and resp.status_code won't exist

accounts/views.py
bdejong: ok

templates/accounts/donation_list.html
```diff
@@ -0,0 +1,41 @@
...
+   <h2>Last Donations</h2>
+    <div class="donations-list">
+    {% for donation in object_list %}
+      <div class="donation-item">
+        <div class="donation-item-left"><img src="/media/images/70x70_avatar.png"></div>
+        <div class="donation-item-right"><p>{{donation.display_name}} <br /> {{donation.amount}} {{donation.currency}}</p></div>
+      </div>
+    {% endfor %}
+    </div>
+    {% if is_paginated %}
```
bdejong: don't we have a pagination block somewhere you can reuse? @ffont also

accounts/views.py
```diff
@@ -1069,3 +1066,35 @@ def pending(request):
...
+    params = {'cmd': '_notify-validate'}
+    for key, value in request.POST.items():
+        params[key] =  value
+    if settings.DEBUG:
+        paypal_validation_url = "https://www.sandbox.paypal.com/cgi-bin/webscr"
+    else:
+        paypal_validation_url = "https://www.paypal.com/cgi-bin/webscr"
+    req = requests.post(paypal_validation_url, data=params)
+    if req.text == 'VERIFIED':
+        campaign = DonationCampaign.objects.order_by('date_start').last()
```
bdejong: wouldn't it be better to pass the campaign ID to paypal and get it back from them rather than assuming it's the last campaign?

accounts/models.py
```diff
@@ -260,3 +260,17 @@ def create_user_profile(sender, instance, created, **kwargs):
...
 
 post_save.connect(create_user_profile, sender=User)
+
+class DonationCampaign(models.Model):
+    goal = models.DecimalField(max_digits=6, decimal_places=2)
+    date_start = models.DateField()
+
+
+class Donation(models.Model):
+    email = models.CharField(max_length=255)
```
bdejong: This still doesn't allow for people to donate while logged in AND use their Freesound username...

I.e. imho donations should really work like the contact form:

```
if logged_in:
   donate as user or donate anonymously
else:
   give your name here or donate anonymously
```

@ffont your opinion?

bdejong:
I would still *heavily* vote for putting this in its own project rather than mixing it in with Account.

accounts/views.py
andrebola: I was assuming we will have only one campaign at a time, also note there is no 'end_date' on campaigns.

templates/accounts/donation_list.html
ffont: @andrebola is doing pagination using a `ListView` which does not use our typical pagination strategy with our [templatetag](https://github.com/MTG/freesound/blob/4e4b91fa45adabb5294487eeed384596425fd8bb/general/templatetags/paginator.py) and [pagination](https://github.com/MTG/freesound/blob/4e4b91fa45adabb5294487eeed384596425fd8bb/utils/pagination.py) function. 
I think it's good if we do things here in the officially proposed "Django way". A number of other simple views that we have could in fact be refactored similarly though no need for that now.

Nevertheless we should make the "new" paginator look the same as the old one (show the numbers etc), I'm sure there are options and ways to do it.

accounts/views.py
ffont: I think it is safe to assume that there will be always one campaign. However I'm not sure I understand @bdejong comment. What does paypal know about our campaigns? As far as I understans campaigns are something we use to simply "group" the donations.

accounts/models.py
ffont: I agree with @bdejong 's point. Donations could work same way as for contact form.

ffont:
Yes, I'd move donations outside accounts too...

accounts/views.py
bdejong: @ffont you can easily pass additional parameters to Paypal. CompaignID could be one of them. Then you can grab the campaign from paypal when you get the response...

I think assuming we only have one Campaign is kind of a bad programming idea, because... what if we ever want to run two campaigns at the same time? We might ask users to help us pay for <new design> and <some new feature> at the same time!

bdejong:
BTW, if you move it outside of applications, make sure you collapse all the migrations into one migration for the final PR. I know this is a PITA, but... it would make the PR a lot more readable.

accounts/views.py
```diff
@@ -19,6 +19,7 @@
 #
 
 import datetime, logging, os, tempfile, shutil, hashlib, base64, json
+import requests
```
bdejong: can be removed

donations/views.py
```diff
@@ -0,0 +1,94 @@
...
+    if settings.DEBUG:
+        paypal_validation_url = "https://www.sandbox.paypal.com/cgi-bin/webscr"
+    else:
+        paypal_validation_url = "https://www.paypal.com/cgi-bin/webscr"
+    req = requests.post(paypal_validation_url, data=params)
+    if req.text == 'VERIFIED':
+        extra_data = json.loads(base64.b64decode(params['custom']))
+        campaign = DonationCampaign.objects.get(id=extra_data['campaign_id'])
+        user = None
+        if extra_data['user_id']:
```
bdejong: if "user_id" in extra_data?

donations/views.py
```diff
@@ -0,0 +1,94 @@
...
+@csrf_exempt
+def donation_complete(request):
+    '''This view listens to a notification made from paypal when someone makes
+    a donation, it validates the data and then stores the donation.
+    '''
+    params = {'cmd': '_notify-validate'}
+    for key, value in request.POST.items():
+        params[key] =  value
+    if settings.DEBUG:
+        paypal_validation_url = "https://www.sandbox.paypal.com/cgi-bin/webscr"
```
bdejong: I would move all of these URL's to the settings and just use SETTINGS.PAYPAL_VALIDATION_URL in here.
i.e. use the settings to decide what this URL is rather than hardcoding the url in your functions.

donations/views.py
andrebola: If the value of user_id is None then is an anonymous donation, do you think it would be better to not send the user_id key in that case?

bdejong: that's a good idea...

freesound/local_settings.example.py
```diff
@@ -46,7 +46,8 @@
 
 GEARMAN_JOB_SERVERS = ["localhost:4730"]
 
-PLEDGIE_CAMPAIGN=14560
+
+PAYPAL_EMAIL = "fs@freesound.org"
```
bdejong: by the way this is paypal@freesound.org I think...!

andrebola:
This PR is working on test.freesound.org. 

@bdejong we need a better text for the mail thanking the donations: https://github.com/MTG/freesound/blob/donations/templates/donations/email_donation.txt

bdejong:
How about this ( @alastair could you perhaps check the wording? )

```django
{% block salutation %}{% if user %}{{ user.username }}{% else %}{{display_name}}{% endif %}{% endblock %}

{% block body %}
It looks like you just donated to Freesound...

You, are awesome!

Your money will be used for further development of Freesound and for general support of community-building efforts for Freesound!

If you didn't choose to donate anonymously your name should now be displayed in the list of donors:
<{% absurl 'donors' %}>
{% endblock %}
```

ffont:
Hi,

Some notes about what's currently in test:

- "Thank you for supporting Freesound, you can complete the following form with the name that will appear with the donation" -> "Thanks for supporting Freesound! Please choose the name that will appear with the donation:"
- When donating and being logged in, the options could be: 
  - {username} (your Freesound username)
  - Anonymous
  - Other: [input field]
 In this way you can also use an alternative name (as when donating and being logged out).
- The full list of donations should be paginated with page numbers as in other paginated lists. 
- User avatars in the full list of donations can be smaller, maybe you can reuse design from: https://www.freesound.org/people/frederic.font/following_users/

andrebola:
@ffont your suggestions are now working on test

templates/donations/donation_list.html
```diff
@@ -1,49 +1,22 @@
...
-
-      <li>
-      <span style="color: #888;">
-          Page {{ page_obj.number }} of {{ page_obj.paginator.num_pages }}.
-      </span>
-      </li>
-      </ul>
-    {% endif %}
+    <div style="clear:both;padding-top: 15px;">
+    {% show_paginator page_obj.paginator page_obj page_obj.number request "donations" %}
```
ffont: Does this mean that our `show_paginator` templatetag works out of the box with `ListView`'s paginator? neat.

ffont:
More comments:

- List of donations page lacks a link back to donate
- The design for this list could be improved (I was only suggesting before). I'd make the box fill the full width (like in [activity stream](https://www.freesound.org/home/stream/)), and I would also group the objects by month.
- When logged in the text for the options I would set it literally as I was suggesting in the previous message. For my user it would be:
 - frederic.font
 - Anonymous
 - Other: [input box]
- I would also unify the form for logged in and non logged in, but for non-logged in you only have "Anonymous" and "other" options...
- Also remove the bullet points from the form, the radio buttons already act as bullet points

templates/donations/donation_list.html
andrebola: yes!

ffont:
Nice, I'd say we're almost there :)
I'd increase the number of donors per page (maybe set a variable in settings?)
Also the default value for donation when you're logged in I'd use the username instead of anonymous.

I have another comment that maybe should go to a new issue. We were commenting that we could thank users that have donated by showing a badge next to their names. Implementation of the badge is certainly out of the scope of this PR (we should think about badges more generally, e.g., use badges for moderators, etc.), and probably requires a refactor of how do we display users in Freesound (i.e. implementing a proper templatetag), but maybe thinking in the future we could already do something like set a boolean flag to user profiles that have donated (or save the total amount or number of donations). What do you think?

andrebola:
I think it would be good to show this badge next to the username or at least somewhere in the user's page. This could be realated with the issue #660 (see last comment)

The donations are linked to the user so depending on what we want to do it might be necessary to have an additional field or not.

ffont:
Ok lets leave this for issue #660 then or some other bigger issue that includes #660.
I'd say we should just test a bit more and the PR could be merged. I guess I need a sandbox paypal account right? I think you mentioned you created one. Could you send me the credentials?

donations/views.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@csrf_exempt
+def donation_complete(request):
+    '''This view listens to a notification made from paypal when someone makes
+    a donation, it validates the data and then stores the donation.
+    '''
+    params = {'cmd': '_notify-validate'}
+    for key, value in request.POST.items():
+        params[key] =  value
```
bdejong: Careful: a POST item is actually not a regular dictionary:
https://docs.djangoproject.com/en/1.10/ref/request-response/#django.http.QueryDict

So, you should really be doing:

```python
params = request.POST.copy()
params.update({'cmd': '_notify-validate'})
```

donations/views.py
```diff
@@ -0,0 +1,90 @@
...
+def donate(request):
+    ''' Donate page: display form for donations where if user is logged in
+    we give the option to doneate anonymously otherwise we just give the option
+    to enter the name that will be displayed.
+    If request is post we generate the data to send to paypal.
+    '''
+    if request.method == 'POST':
+        campaign = DonationCampaign.objects.order_by('date_start').last()
+        returned_data = {"name": None, "campaign_id": campaign.id}
+        annon = request.POST.get('annon', None)
```
bdejong: couldn't all of this stuff be done by creating a Form object?

it might be more work, but probably (much) less error-prone...

donations/views.py
```diff
@@ -0,0 +1,90 @@
...
+
+        data = {"url": settings.PAYPAL_VALIDATION_URL,
+                "params": {
+                    "cmd": "_donations",
+                    "currency_code": "EUR",
+                    "business": settings.PAYPAL_EMAIL,
+                    "item_name": "Freesound",
+                    "custom": returned_data_str,
+                    "notify_url": return_url}}
+        return JsonResponse(data)
```
bdejong: pedantic mode on: http://stackoverflow.com/a/6388237/1194410 ;)

donations/views.py
```diff
@@ -0,0 +1,90 @@
...
+    else:
+        tvars = {}
+        return render(request, 'donations/donate.html', tvars)
+
```
bdejong: pedantic mode on: why create a variable if you will only use it in one place?

bdejong:
Added a few more (some pedantic) comments, but this is getting pretty close, I agree!!

ffont:
I tested it and it seems to work, but should I have received an email from Freesound?
(I donated with my user).
Freesound test only sends emails to addresses listed in `ALLOWED_EMAILS`, but my address is there ;)
If you want to test for your user you should edit `local_settings.py` for test in deploy repository.

andrebola:
@ffont The mail address used is the one of his paypal account, since for some cases we don't have other address. If you want, for the case that the user is donating logged in, we could use the address of the fs account.

ffont:
Sure, I think when the user is logged in we should use his email address (independently of whether he wants his name to appear in the list or appear as anonymous or with a different name).

ffont:
I've been testing a bit more and seems to work nicely 💃. Just some more comments:

- Usernames in donors list should be truncated at X chars so that they don't overflow (try with a long name). We have templatetag for that.
- Users that donate while logged in and using their username should be listed with a link to their profile pages (like the name links to the profile). I guess this is when `donation.user.username == donation.display_name`.
- If making a donation while logged in but setting the name to "anonymous" or "other" then I did not receive a fs "thank you" email. Does this mean that `donation.user` is only filled when user is logged in AND choses to donate with her name? I think this should not be the case. In my opinion we should link the donation with the user if logged in, even if she choses to appear as anonymous or under another name in the donors list. In fact the text says "...choose the name that will appear...", does not say that the donation will be anonymous even to Freesound. If you want to make the donation completely anonymous then a user should log out. So lets make sure that we store the user if user is logged in and send the thank you email. In fact, in the thank you email it says "If you didn't choose to donate anonymously your name should now be displayed in the list of donors", therefore implying that you should receive this email even if donating anonymously right?

I think after these changes I'll finally deploy on monday 👍

alastair:
> Users that donate while logged in and using their username should be listed with a link to their profile pages (like the name links to the profile). I guess this is when donation.user.username == donation.display_name..
Of course, we don't want users to be able to pretend to donate as someone else, either

ffont:
Errr, user avatar should ONLY be shown when donating logged in and as Freesound user. If logged in but anonymous or alternative name is chosen, you should not see the avatar. Maybe you can add an `is_anonymous` property to the donation model and use this where needed.
Other than that I think this is ready.

bdejong:
Hey guys, checked it out on Test. Very nice.

One thing: displaying the amount donated is not a good idea I think. Or it needs to be an option as well. I.e. "show how much I donated".

Issue #632, opened 2014-09-19
Refactor 'most downloaded in the last week' list of sounds


ffont:
Apparently to create that list, the 'created' field of the sound model is taken into account. However, the date of the created field might not match the date when the sound is approved in moderation, preventing many sounds to appear in this list. We should probably use moderation date instead of creation date.

ffont:
Deployed

PR #776, opened 2016-12-27
Change 'Most downloaded sounds' to use moderation_date

andrebola:
issue #632


ffont:
Is moderation_date set on whitelisted sounds?
Make sure that this is set.
I'd also set this as an option. I kind of don't remember exactly why we didn't make this change before. There are probably other side effects like "uploaded in the las week" would really mean "moderated in the last week" and therefore dates might seem to be confusing.

Issue #652, opened 2015-01-13
To many geotags in main geotags map


ffont:
The current way in which we display sounds with geotags in the main geotags map can not handle the number of geotagged sounds that we have in our db (the generated json seems to be too long). We should change the way in which data is acquired so that it can be correctly displayed. Best way is probably to request sounds in pages instead of all at once.

Right now, and to allow the map to load we imposed a limit in the number of sounds that are displayed, but this is only a temporary solution.

ffont:
Actually what happened is that there was a corrupted geotag which broke the javascript al client side (generating an invalid json). Nevertheless, we should improve the way in which the json is requested by the client. Changing from "bug" to "improvement" now.

ffont:
If we want to improve speed we should probably do the clustering on the server.
Have a look at http://postgis.net (postgresql extension)

andrebola:
Right now for the page: https://www.freesound.org/browse/geotags/ took about 1 sec. to load the response and the json was 1 MB.

After doing some research I found that we could use postgis to cluster the markers on the server, the clusters could be generated using [K-means or by a distance ](http://gis.stackexchange.com/a/11778). But we have to do a new request to the server every time the user zoom in the map and calculate the clusters again depending on the position (this could be cached).

The other option is to use solr to return the json information of all the markers, this could be faster of what we have now and easier to implement than the postgis option. 

Solr also supports [clustering of the points](https://wiki.apache.org/solr/SpatialClustering), but I think I would prefer to use postgis in the case of server-side clustering.

bdejong:
There is also: https://cwiki.apache.org/confluence/display/solr/Spatial+Search

Which would enable you do to very quick rectangle searches, but clustering would still be needed at zoomed out viewing.

andrebola:
I think with solr we could reduce only about one second (of querying postgres and generating the json), if that is enough I could do it. Otherwise we would have to implement the clustering.

ffont:
> After doing some research I found that we could use postgis to cluster the markers on the server, the clusters could be generated using K-means or by a distance . But we have to do a new request to the server every time the user zoom in the map and calculate the clusters again depending on the position (this could be cached).

That's right. A simple caching wouldn't help much because rectangle borders are rather "continuous" so it wouldn't be straightforward to define useful cache keys. Maybe we can optimise the json thing by using solr and that's it...

PR #777, opened 2016-12-27
Changed global geotag map to use solr data

andrebola:
issue #652

The idea with this is to test global geomap to check if works faster with solr.

### DIFF
--- a/geotags/views.py
+++ b/geotags/views.py
@@ -56,7 +56,7 @@ def generate_json(sound_queryset):
 
     return HttpResponse(json.dumps(sounds_data), content_type="application/json")
 
-@cache_page(60 * 15)
+#@cache_page(60 * 15)
 def geotags_json(request, tag=None):
     if tag:
         #sounds = Sound.objects.select_related('geotag').filter(tags__tag__name=tag).exclude(geotag=None)
@@ -80,7 +80,9 @@ def geotags_json(request, tag=None):
         query.set_query_options(field_list=["id", "geotag"],
                 rows=results.num_found, filter_query=filter_query)
         results = SolrResponseInterpreter(solr.select(unicode(query)))
-        sounds_data = [[s['id']]+ s['geotag'][0].split(' ') for s in results.docs]
+        sounds_data = map(lambda x: [x['id'],
+            x['geotag'][0].split(' ')[1],
+            x['geotag'][0].split(' ')[0]], results.docs)
         return HttpResponse(json.dumps(sounds_data), content_type="application/json")
     return generate_json(sounds)
 
--- a/geotags/views.py
+++ b/geotags/views.py
@@ -23,6 +23,7 @@ from django.http import Http404, HttpResponse
 from django.shortcuts import render_to_response
 from django.template import RequestContext
 from sounds.models import Sound
+from utils.search.solr import Solr, SolrQuery, SolrResponseInterpreter
 from django.views.decorators.cache import cache_page
 from django.contrib.auth.models import User
 import json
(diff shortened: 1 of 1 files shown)


ffont:
I just tested that in freesound test and the obtained results do not seem to be the same using solr than before. See the attached images below. First one corresponds to geotags map before using solr, second one geotags map when using solr. Something is clearly wrong. This should be investigated.

before

after

andrebola:
@ffont is it possible solr in test doesn't have all the sounds?

ffont:
It does not have all the sounds but this should not be a problem
The comparison is with pg-based geotags and solr-based geotags in test (I see the same in local).
Maybe the problem is just that solr index is not updated (that would be a first thing to try locally).
However I think that the problem is that solr is returning way less results that pg. I guess solr can't return all results at once, so it is possible that we won't really be able to implement this geotags thing using solr. Same problem we had with the combined search in the API.

andrebola:
In solr the information of lat and long was in other order that is why they were in the wrong position in the map. 
Let's try on test if solr is returning all the items, in my local works fine with around 4k items.

bdejong:
Te biggest problem with this approach (getting the lat/lon from Solr) is
that IIRC we don't invalidate and resubmit sounds to Solr at all. I.e.the
Solr index is always out of date...

I also don't really understand why Solr would be quicker at just dumping
lat/lon to a JSON than Postgres. Did you guys do some analysis on what the
slow part of this is? Is it really the Postgres query? If so, can't we just
optimise that bit?

On Dec 28, 2016 18:31, "Andres Ferraro" <notifications@github.com> wrote:

In solr the information of lat and long was in other order that is why they
were in the wrong position in the map.
Let's try on test if solr is returning all the items, in my local works
fine with around 4k items.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/MTG/freesound/pull/777#issuecomment-269510692>, or mute
the thread

andrebola:
The problem of the index not been updated could be fixed. But I think the queries can't be optimized since are done with raw sql. 

The idea behind this PR it was to test if the performance was better, but we already now that the problem is that the json file is too big (2mb). The only option to reduce this file is to do the clustering on the server but that's harder.

ffont:
If performance does not change significantly then I would not change the current method.
Maybe we could however add some indication that the page is "loading" so users don't get confused. For example add a rolling loading indicator on top of the map or something like this. Or add a loading bar "loading geotags...".

bdejong:
The problem is the floating point numbers. They explode in JSON as you need
20+ bytes to encode a 4-byte float.

What you could do is encode the data in BSON and use a BSON parser on the
client side. This would optimize stuff dramatically just by changing the
intermediate format. You would gain at least a factor 5 for
encode/download/decode!

For example:
http://stackoverflow.com/questions/7541552/is-a-binary-json-javascript-library-available-for-browsers

Or you could use protocol buffers:
https://github.com/dcodeIO/ProtoBuf.js/

I'm sure there are more binary data transfer types for JavaScript.

On Jan 2, 2017 2:30 PM, "Frederic Font" <notifications@github.com> wrote:

If performance does not change significantly then I would not change the
current method.
Maybe we could however add some indication that the page is "loading" so
users don't get confused. For example add a rolling loading indicator on
top of the map or something like this. Or add a loading bar "loading
geotags...".

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/MTG/freesound/pull/777#issuecomment-269972707>, or mute
the thread

ffont:
That makes total sense...

alastair:
We do this in dunya for some bits of data. We may not even need a schema-based storage format - if you just write out latlnglatlng to a file using 16 or 32-bit integers (e.g. by taking a decimal value and multiplying by 1000), you can read them directly by using Uint16Array in javascript.

andrebola:
I have created a new branch with this changes, see PR #783

bdejong:
@alastair uuuugh, ugly hack! @andrebola I should have read this before commenting on your PR.
please, pretty please user another, nicer intermediary format.

ffont:
I have no preference in any solution (bson, proto buffers or raw binary), but in any case @bdejong you're not stating why shouldn't we use raw binary. To me it looks like I don't see a real difference for such a small problem (surely not in performance?), and using raw binary we could also (probably) reuse some code from Dunya. So I see this as being justified. Unless @bdejong can convince me that using one of the other formats will be worth the change, I won't leave it as it is 🎆

bdejong:
Raw binary has no metadata attached to it. So, there is no way to interpret raw binary data without someone telling you exactly what the format is. I.e. you have to document separately what you are sending.

With any other data serialization format (bson, protobuf, avro, thrift, ...) you get the benefit that the interpretation is human readable.

And the code will look nicer on BOTH ends:

https://github.com/py-bson/bson

```python
>>> import bson
>>> a = bson.dumps({"A":[1,2,3,4,5,"6", u"7", {"C":u"DS"}]})
```

https://github.com/mongodb/js-bson

```javascript
var data = bson.deserialize(data)
```

versus messing around with data/structure packing and unpacking.

ffont:
We have a bit of a mess with this pull request and pull request https://github.com/MTG/freesound/pull/783. I think we decided there is no real advantage on using solr, but we should definitely deploy the bytearray thing. I think this PR should be closed although it contains a conversation relevant to https://github.com/MTG/freesound/pull/783.

PR #783, opened 2017-01-11
Change json response of geotags for bytearray

andrebola:
issue #652

### DIFF
--- a/geotags/views.py
+++ b/geotags/views.py
@@ -35,7 +35,7 @@ def generate_bytearray(sound_queryset):
     # sounds as bytearray
     packed_sounds = cStringIO.StringIO()
     for s in sound_queryset:
-        if not math.isnan(s.geotag.lat) and not math.isnan(s.geotag.lat):
+        if not math.isnan(s.geotag.lat) and not math.isnan(s.geotag.lon):
             packed_sounds.write(struct.pack("i", s.id))
             packed_sounds.write(struct.pack("i", int(s.geotag.lat*1000000)))
             packed_sounds.write(struct.pack("i", int(s.geotag.lon*1000000)))
--- a/geotags/views.py
+++ b/geotags/views.py
@@ -18,6 +18,8 @@
 #     See AUTHORS file.
 #
 
+import cStringIO
+import struct
 from django.conf import settings
 from django.http import Http404, HttpResponse
 from django.shortcuts import render_to_response
@@ -47,16 +49,18 @@ geoquery = """SELECT sounds_sound.id,
             ORDER BY sounds_sound.created DESC
                      %(end)s"""
 
-def generate_json(sound_queryset):
-    # When using Django Sound queries:
-    #sounds_data = [[s.id, s.geotag.lat, s.geotag.lon] for s in sound_queryset]
+def generate_bytearray(sound_queryset):
+    # sounds as bytearray
+    packed_sounds = cStringIO.StringIO()
+    for sound in sound_queryset:
+        packed_sounds.write(struct.pack("i", sound.id))
+        packed_sounds.write(struct.pack("i", int(sound.lat*1000)))
+        packed_sounds.write(struct.pack("i", int(sound.lon*1000)))
 
-    sounds_data = [[s.id, s.lat, s.lon] for s in sound_queryset if not math.isnan(s.lat) and not math.isnan(s.lon)]
-
-    return HttpResponse(json.dumps(sounds_data), content_type="application/json")
+    return HttpResponse(packed_sounds.getvalue(), content_type='application/octet-stream')
 
 @cache_page(60 * 15)
-def geotags_json(request, tag=None):
+def geotags_barray(request, tag=None):
     if tag:
         #sounds = Sound.objects.select_related('geotag').filter(tags__tag__name=tag).exclude(geotag=None)
         join = """INNER JOIN "tags_taggeditem"
@@ -70,13 +74,14 @@ def geotags_json(request, tag=None):
         #sounds = Sound.public.select_related('geotag').all().exclude(geotag=None)
         q = geoquery % {"join": "", "where": "", "end": ""}
         sounds = Sound.objects.raw(q)
-    return generate_json(sounds)
 
-def geotags_box_json(request):    
+    return generate_bytearray(sounds)
+
+def geotags_box_barray(request):
     box = request.GET.get("box","-180,-90,180,90")
     try:
-        min_lat, min_lon, max_lat, max_lon = box.split(",")  
-        qs = Sound.objects.select_related("geotag").exclude(geotag=None).filter(moderation_state="OK", processing_state="OK")        
+        min_lat, min_lon, max_lat, max_lon = box.split(",")
+        qs = Sound.objects.select_related("geotag").exclude(geotag=None).filter(moderation_state="OK", processing_state="OK")
         if min_lat <= max_lat and min_lon <= max_lon:
             sounds = qs.filter(geotag__lat__range=(min_lat,max_lat)).filter(geotag__lon__range=(min_lon,max_lon))
         elif min_lat > max_lat and min_lon <= max_lon:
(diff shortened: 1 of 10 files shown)


bdejong:
Um, sorry, but I think this is absolutely the worst possible solution to this problem. With all the various intermediate formats out there (bson, protobuf, capnproto, ...) using raw binary should really be avoided.

geotags/views.py
```diff
@@ -70,55 +74,56 @@ def geotags_json(request, tag=None):
...
             sounds = qs.exclude(geotag__lat__range=(max_lat,min_lat)).filter(geotag__lon__range=(min_lon,max_lon))
         elif min_lat <= max_lat and min_lon > max_lon:
             sounds =qs.filter(geotag__lat__range=(min_lat,max_lat)).exclude(geotag__lon__range=(max_lon,min_lon))
         elif min_lat > max_lat and min_lon > max_lon:
-            sounds = qs.exclude(geotag__lat__range=(max_lat,min_lat)).exclude(geotag__lon__range=(max_lon,min_lon))        
+            sounds = qs.exclude(geotag__lat__range=(max_lat,min_lat)).exclude(geotag__lon__range=(max_lon,min_lon))
 
         sounds_data = [[s.id, s.geotag.lat, s.geotag.lon] for s in sounds]
-        return HttpResponse(json.dumps(sounds_data), content_type="application/json")
+        return generate_bytearray(sounds_data)
```
alastair: `generate_bytearray` takes a queryset here, not a list/tuple - maybe some tests? ;)

ffont:
@bdejong Re the solution I'm answering in Issue #777 as the discussion there is more focused on that.
Other than that looks good to me, however I see bigger changes than I expected (specially in javascirpt part) so this must be tested carefully).

templates/accounts/account.html
```diff
@@ -322,37 +322,46 @@
...
-                        stopAll();
-                        ajaxLoad( '/geotags/infowindow/' + id, function(data, responseCode)
+  
+                var raw_data = new Int32Array(oReq.response);
+
+                var id = null;
+                var lat = null;
+                var lon = null;
+
+                $.each(raw_data, function(index, item) {
```
alastair: Can we use a for loop with an increment of 3 here, instead of .each/modulo?

geotags/views.py
```diff
@@ -47,16 +49,18 @@
...
 
-def generate_json(sound_queryset):
-    # When using Django Sound queries:
-    #sounds_data = [[s.id, s.geotag.lat, s.geotag.lon] for s in sound_queryset]
+def generate_bytearray(sound_queryset):
+    # sounds as bytearray
+    packed_sounds = cStringIO.StringIO()
+    for sound in sound_queryset:
+        packed_sounds.write(struct.pack("i", sound.id))
+        packed_sounds.write(struct.pack("i", int(sound.lat*1000)))
```
alastair: should be `sound.geotag.lat`, right?

geotags/views.py
```diff
@@ -47,16 +49,18 @@
...
+        packed_sounds.write(struct.pack("i", int(sound.lon*1000)))
 
-    sounds_data = [[s.id, s.lat, s.lon] for s in sound_queryset if not math.isnan(s.lat) and not math.isnan(s.lon)]
-
-    return HttpResponse(json.dumps(sounds_data), content_type="application/json")
+    return HttpResponse(packed_sounds.getvalue(), content_type='application/octet-stream')
```
alastair: Can we also gzip this? I don't know the best way to make that happen. Should we do it in django, or is there something that we can tell nginx to make it compress responses to these endpoints?

```
alastair@mahler:~/code/freesound$ ls -l geotags_box*
-rw-r--r-- 1 alastair alastair 227904 Jan 12 10:46 geotags_box_barray
-rw-r--r-- 1 alastair alastair 815860 Jan 12 10:46 geotags_box_json

-rw-r--r-- 1 alastair alastair 105619 Jan 12 10:46 geotags_box_barray.gz
-rw-r--r-- 1 alastair alastair 294641 Jan 12 10:46 geotags_box_json.gz
```
Binary gets us 4x smaller than json, and gzipping another 2x smaller, which is a pretty good result, I think

geotags/views.py
andrebola: I think 1000 is too low since we allow full zoom in, we should use at least 10000 or 1000000

'sound' is the output of the raw query, in the old code was: https://github.com/MTG/freesound/blob/master/geotags/views.py#L54
So I think it's OK

ffont: The raw query gets id, lat and lon in SELECT, thus it should work in most views, but `geotags_box_json ` https://github.com/MTG/freesound/blob/master/geotags/views.py#L75 does not make a raw query so it will fail. I guess this view should be refactored to use the raw query so it works like the others. Otherwise it should be annotated somehow so s.lat and s.lon are available.

geotags/views.py
```diff
@@ -47,16 +49,18 @@
...
-    #sounds_data = [[s.id, s.geotag.lat, s.geotag.lon] for s in sound_queryset]
+def generate_bytearray(sound_queryset):
+    # sounds as bytearray
+    packed_sounds = cStringIO.StringIO()
+    for sound in sound_queryset:
+        packed_sounds.write(struct.pack("i", sound.id))
+        packed_sounds.write(struct.pack("i", int(sound.lat*1000)))
+        packed_sounds.write(struct.pack("i", int(sound.lon*1000)))
 
-    sounds_data = [[s.id, s.lat, s.lon] for s in sound_queryset if not math.isnan(s.lat) and not math.isnan(s.lon)]
```
ffont: Please make this check too in `generate_bytearray` version. It was added to avoid "corrupted" geotags to break the whole map. I know the problem is "corrupted" geotags but anyway, we don't want user-facing things to break.

geotags/views.py
ffont: This is related to the comment above. This view is the only one that does not use raw queries. Either this view is refactored to use a raw query (needs careful testing though), or the queryset is annotated with lat and lon properties and passed to `generate_bytearray` directly. Another option is to check if using raw queries is still needed with the version of Django that we use (see comment https://github.com/MTG/freesound/blob/098bfc05e975163aeb3931c648c83e225a31e0fe/geotags/views.py#L34).

geotags/views.py
alastair: https://en.wikipedia.org/wiki/Decimal_degrees lists approximations of scales. I wouldn't go higher than 5dp, maybe 4 is OK too.

ffont: As along as filesize does not change, I would use the best resolution possible.
It is quite typical in Freesound that you have N recordings from the same place but don't want them to completely overlap in the map so you place them in slightly different places (probably less than level 5 size). So lets do the better resolution we can.

bdejong: From a binary point of view: you are using 4-byte ints to pack the data. Those go from -2147483648 to 2147483647. You want to encode lat and lon (float), which as a maximum range go from -180 to 180. That means you can use a lot more precision. You can can use: 10000000. (i.e. 2147483647/180 = 11930464.7, round down to nearest "round number" gets you 10000000)

geotags/views.py
bdejong: can't we do this filtering in either the query (raw) or queryset (django query)?

templates/accounts/account.html
bdejong: and shouldn't this functionality be put into a separate class/function? It will be repeated all overt the place and the "1000" magic number is going to appear everywhere...

geotags/views.py
andrebola: I checked the queries and they were INNER JOIN so I changed back to django queries.

ffont:
I tested the PR but I saw some bugs:

- https://github.com/MTG/freesound/blob/947480de3307c59aa1d196f18d817bb275bc0b69/geotags/views.py#L38 you're checking `s.geotag.lat` two times instead of lat and lon.
- In the `/browse/geotags/` page, all sound info boxes show the same content to be (like all sounds are `degge.aiff by hanstimm`). This does not seem to happen in user home/profile or pack page maps. I think it takes the first sound it receives or something like this.

Issue #730, opened 2016-04-13
Tag recommendation endpoint raises 500 if service not running


alastair:
```
Traceback:
File "/home/alastair/code/freesound/env/local/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  109.                         response = callback(request, *callback_args, **callback_kwargs)
File "/home/alastair/code/freesound/utils/tagrecommendation_utilities.py" in get_recommended_tags_view
  78.                 tags, community = get_recommended_tags(input_tags)
File "/home/alastair/code/freesound/utils/tagrecommendation_utilities.py" in get_recommended_tags
  69.     return recommended_tags['tags'][:max_number_of_tags], recommended_tags['community']

Exception Type: TypeError at /tagrecommendation/recommendtags/
Exception Value: 'bool' object has no attribute '__getitem__'
Request information:
GET: No GET data

POST:
input_tags = u'foo bar '
```

This is a server error, but we should catch the error and return something custom rather than the standard django error. Perhaps 503 (Service unavailable).

ffont:
If recommendation server does not return, then no tags are recommended. Not a big deal, but as @alastair says we should catch the error in the view and return something which the client understands. No need to show anything to users, just handle error internally and do nothing.

PR #786, opened 2017-01-18
Return error 503 when tag recommendation is not working

andrebola:
issue #730

### DIFF
--- a/utils/tagrecommendation_utilities.py
+++ b/utils/tagrecommendation_utilities.py
@@ -18,6 +18,7 @@
 #     See AUTHORS file.
 #
 
+import urllib2
 import logging, traceback
 from django.conf import settings
 from tagrecommendation.client import TagRecommendation
@@ -50,21 +51,15 @@ def get_recommended_tags(input_tags, max_number_of_tags=30):
             recommended_tags = False
 
     if not recommended_tags:
+        recommended_tags = TagRecommendation.recommend_tags(input_tags)
+
+        if not recommended_tags['tags']:
+            recommended_tags['community'] = "-"
+
         try:
-            recommended_tags = TagRecommendation.recommend_tags(input_tags)
-
-            if not recommended_tags['tags']:
-                recommended_tags['community'] = "-"
-
-            try:
-                cache.set(cache_key, recommended_tags, TAGRECOMMENDATION_CACHE_TIME)
-            except:
-                pass
-
-        except Exception, e:
-            logger.error('Could not get a response from the tagrecommendation service (%s)\n\t%s' % \
-                         (e, traceback.format_exc()))
-            recommended_tags = False
+            cache.set(cache_key, recommended_tags, TAGRECOMMENDATION_CACHE_TIME)
+        except:
+            pass
 
     return recommended_tags['tags'][:max_number_of_tags], recommended_tags['community']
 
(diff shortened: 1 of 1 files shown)


utils/tagrecommendation_utilities.py
```diff
@@ -75,8 +75,11 @@ def get_recommended_tags_view(request):
         if input_tags:
             input_tags = list(clean_and_split_tags(input_tags))
             if len(input_tags) > 0:
-                tags, community = get_recommended_tags(input_tags)
-                return HttpResponse(json.dumps([tags, community]), content_type='application/javascript')
+                try:
+                    tags, community = get_recommended_tags(input_tags)
+                    return HttpResponse(json.dumps([tags, community]), content_type='application/javascript')
+                except Exception:
```
alastair: Can we have a specific exception here?

utils/tagrecommendation_utilities.py
```diff
@@ -158,3 +161,7 @@ def get_all_categories_view(request):
         return HttpResponse(json.dumps(categories), content_type='application/javascript')
 
     return HttpResponse(json.dumps([]), content_type='application/javascript')
+
+
+class HttpResponseUnavailabileError(HttpResponse):
+        status_code = 503
```
alastair: 4 spaces, or 8?

utils/tagrecommendation_utilities.py
```diff
@@ -158,3 +161,7 @@ def get_all_categories_view(request):
         return HttpResponse(json.dumps(categories), content_type='application/javascript')
```
alastair: There are some other views here which return an `HttpResponse`, should these also return 503 when needed?

utils/tagrecommendation_utilities.py
```diff
@@ -50,21 +51,15 @@ def get_recommended_tags(input_tags, max_number_of_tags=30):
...
+        if not recommended_tags['tags']:
+            recommended_tags['community'] = "-"
 
-        except Exception, e:
-            logger.error('Could not get a response from the tagrecommendation service (%s)\n\t%s' % \
-                         (e, traceback.format_exc()))
-            recommended_tags = False
+        try:
+            cache.set(cache_key, recommended_tags, TAGRECOMMENDATION_CACHE_TIME)
+        except:
```
alastair: Even though this was just copied from above, it would be nice to take the chance to fix exceptions here as well, since you're working on this code

utils/tagrecommendation_utilities.py
```diff
@@ -78,10 +73,12 @@ def get_recommended_tags_view(request):
...
                     tags, community = get_recommended_tags(input_tags)
                     return HttpResponse(json.dumps([tags, community]), content_type='application/javascript')
-                except Exception:
+                except urllib2.URLError, e:
+                    logger.error('Could not get a response from the tagrecommendation service (%s)\n\t%s' % \
+                         (e, traceback.format_exc()))
                     return HttpResponseUnavailabileError()
 
-    return HttpResponse(json.dumps([[],"-"]), content_type='application/javascript')
+    return HttpResponseUnavailabileError()
```
alastair: I'm not sure this response is correct in this case - the returned data is "empty", which isn't an error case, it's just the case which is encountered when there is no data to respond to.
Looking at the javascript which processes this endpoint, it looks like there is a case which checks empty results: https://github.com/MTG/freesound/blob/132c96be9895ba67cd4a094882b36dec15876b95/templates/accounts/describe_sounds.html#L183
If this endpoint now returns 503, the js would have to be changed to perform this action on the POST failure. Instead, I think we should only return http503 in the single situation where the recommendation service is unavailable.

utils/tagrecommendation_utilities.py
```diff
@@ -75,10 +70,15 @@ def get_recommended_tags_view(request):
         if input_tags:
             input_tags = list(clean_and_split_tags(input_tags))
             if len(input_tags) > 0:
-                tags, community = get_recommended_tags(input_tags)
-                return HttpResponse(json.dumps([tags, community]), content_type='application/javascript')
+                try:
+                    tags, community = get_recommended_tags(input_tags)
+                    return HttpResponse(json.dumps([tags, community]), content_type='application/javascript')
+                except urllib2.URLError, e:
```
alastair: we should use the modern `except Exception as e` format here

utils/tagrecommendation_utilities.py
andrebola: I think there is no need to handle exceptions here, like we do at this line https://github.com/MTG/freesound/blob/master/sounds/views.py#L77

PR #788, opened 2017-01-23
User delete

andrebola:
Through the admin page you can check what changes are going to happen when deleting an user and then trigger the async task


templates/accounts/delete_confirmation.html
```diff
@@ -0,0 +1,39 @@
...
+
+{% block breadcrumbs %}
+<div class="breadcrumbs">
+<a href="{% url 'admin:index' %}">{% trans 'Home' %}</a>
+</div>
+{% endblock %}
+
+{% block content %}
+<p>The User "{{anonymised}}" will be anonymised</p>
+  {% if logic_deleted %} {{ logic_deleted.count}} Packs will be logically deleted {% endif %}
```
ffont: What does logically deleted mean? I would add not more information...

accounts/admin.py
```diff
@@ -76,21 +116,45 @@ def full_delete(self, request, obj):
```
ffont: the `full_delete` action should also call the async version.

accounts/admin.py
```diff
@@ -29,18 +31,55 @@
...
-    for user in queryset:
-        user.profile.delete_user(remove_sounds=True)
+    if request.POST.get('confirmation', False):
+        gm_client = gearman.GearmanClient(settings.GEARMAN_JOB_SERVERS)
+        for user in queryset:
+            gm_client.submit_job("delete_user",
+                    json.dumps({'user_id':user.id, 'action': "delete_user_delete_sounds"}),
+                wait_until_complete=False, background=True)
+        messages.add_message(request, messages.INFO,
+             '%d Soft deleted users including her sounds' % (queryset.count()))
```
ffont: The message should talk about the future? "users will be soft deleted asynchronously".
The same applies to the other messages of this kind.

ffont:
I tested this PR but had some problems with the deletion code. I think the delete functions sometimes fail (because of broken db constrains or stuff like that) and then the async delete does not succeed  but we are never notified. So the code in this PR I think is alright, but it calls other code that can break ;) The fix for this therefore is probably to handle exceptions in a way that at least admins are notified when a user was tried to be deleted but it didn't succeed. It is hard to know which exceptions can be raised and make delete fail, but I guess its mostly db exceptions. So if there is a way to handle that I'd do it. The other option would be to make sure that in case of error a ticket is opened in sentry (then we can leave the deletion code as it is for now). What do you think @alastair ?

alastair:
I wouldn't want to rely on sentry notifications for error reporting. Are we aware of the types of errors that can occur during this delete step? 
Ideally we would report errors back to the admin users when they happen. What's the best way of doing this? email, in-site notifications? Can we just use the message system?

ffont:
The types of errors are related to database constrains that we've been slowly removing but still cause problems. This means that there are underlying bugs that should be fixed, I agree, but maybe a bit unpredictable right now. I think that as long as we get notified when these happen it is not too bad for now. Why wouldn't you rely on Sentry? That's what we do for all other kinds of errors...

andrebola:
Since the deletion is done asynchronously I don't think we could use the in-site notifications, but we could send an email or a message to the admin. 

If everything works fine then any user can be deleted without any problem. So, if there is an error is because of a failure in the system, this should also be notified to sentry. I don't know why they are not been notified since this should work in the same way of the other tasks in gearman.

ffont:
Well I didn't say these are not being notified, just that we should check. I'll do it and then merge the PR if it works.

ffont:
regardless of the sentry thing, there were a couple of comments in the PR, will you have a look @andrebola ?

templates/accounts/delete_confirmation.html
```diff
@@ -16,7 +16,7 @@
 
 {% block content %}
 {% for anon in anonymised %}
-  <p>The User "{{anon.name}}" will be anonymised</p>
+<p>The User "{{anon.name}}" will be {% if anon.deleted %}deleted{% else %}anonymised{% endif %}</p>
```
ffont: This I don't understand, when is this shown and in which cases do we see deleted or anonymised?

ffont:
I merged the pull request but there's something I don't understand yet

Issue #614, opened 2014-04-14
Simple AJAX checking of username


bdejong:
At signup it is getting harder and harder to figure out if your username already exists or not. I've had people tell me they had to go through 4 cycles of trying. Some simple javascript-checking could be rather useful...

alastair:
This is a lot faster since we added some more indexes to the username field, but an ajax check is still a good idea

PR #792, opened 2017-01-26
Add username check by AJAX on registration

andrebola:
issue #614

### DIFF
--- a/accounts/views.py
+++ b/accounts/views.py
@@ -77,6 +77,17 @@ def crash_me(request):
     raise Exception
 
 
+def check_username(request):
+    username = request.GET.get('username', None)
+    res = {"result": False}
+    if username:
+        try:
+            user = User.objects.get(username__iexact=username)
+        except User.DoesNotExist:
+            res['result'] = True
+    return HttpResponse(json.dumps(res), content_type='application/javascript')
+
+
 @login_required
 def bulk_license_change(request):
     if request.method == 'POST':
--- a/accounts/views.py
+++ b/accounts/views.py
@@ -28,7 +28,7 @@ from django.contrib.auth import logout
 from django.core.urlresolvers import reverse
 from django.db.models import Count
 from django.http import HttpResponseRedirect, HttpResponse, HttpResponseBadRequest, Http404, \
-    HttpResponsePermanentRedirect, HttpResponseServerError
+    HttpResponsePermanentRedirect, HttpResponseServerError, JsonResponse
 from django.shortcuts import get_object_or_404, render
 from django.views.decorators.csrf import csrf_exempt
 from django.contrib import messages
@@ -85,8 +85,7 @@ def check_username(request):
             user = User.objects.get(username__iexact=username)
         except User.DoesNotExist:
             username_valid = True
-    return HttpResponse(json.dumps({'result': username_valid}),
-            content_type='application/javascript')
+    return JsonResponse({'result': username_valid})
 
 
 @login_required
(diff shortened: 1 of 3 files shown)


templates/accounts/registration.html
```diff
@@ -7,6 +7,23 @@
     $(function () {
     	$("#id_username").focus()
     });
+    $(document).ready(function(){
```
bdejong: ```$(function () {```

is already "on document ready", you might want to put it in there...

templates/accounts/registration.html
```diff
@@ -7,6 +7,23 @@
...
+      if (!($('#username_check')).length) {
+        $('#id_username').parent().append('<div id="username_check" style="color: #bd2d33;font-style: italic;"></div>');
```
bdejong: It might be cleaner to work with http://api.jquery.com/toggle/ rather than setting the HTML content...

accounts/views.py
```diff
@@ -77,6 +77,17 @@ def crash_me(request):
     raise Exception
 
 
+def check_username(request):
+    username = request.GET.get('username', None)
+    res = {"result": False}
```
bdejong: Very, very minor comments on style: why create the dict here and not just say: `username_valid = False` and then at return write: `return JsonResponse(dict(result=username_valid))` ? This might be a lot more understandable when reading the code...

andrebola: Yes, that is much better. I keep the brackets instead of 'dict' because is slightly faster: http://stackoverflow.com/questions/1929274/in-python-use-dict-with-keywords-or-anonymous-dictionaries

templates/accounts/registration.html
andrebola: I changed the code to use toggle and is much better now. 
I still have to add the div with JS since I want to place it next to the username input and the hole form it's rendered using {{form.as_p}}.

accounts/views.py
```diff
@@ -77,6 +77,18 @@ def crash_me(request):
...
 
+def check_username(request):
+    username = request.GET.get('username', None)
+    username_valid = False
+    if username:
+        try:
+            user = User.objects.get(username__iexact=username)
+        except User.DoesNotExist:
+            username_valid = True
+    return HttpResponse(json.dumps({'result': username_valid}),
```
bdejong: Why not JsonResponse?

Btw, "readability" versus "speed" is always the worst kind of reason to choose a certain method if you don't talk about relative speed. ;) That said I'm not holy about { } versus dict( )

templates/accounts/registration.html
```diff
@@ -3,10 +3,22 @@
...
 
 {% block section_content %}
+<style> #username_check { display:none; color: #bd2d33; font-style: italic; }</style>
 <script type="text/javascript">
     $(function () {
-    	$("#id_username").focus()
-    });
+      $("#id_username").focus()
+      
+      $('#id_username').parent().append('<div id="username_check">The username is not available</div>');
```
bdejong: Can you please change this to "Sorry, this username is already being used by someone else."?

alastair:
I just used this on the registration page. It would be nice to have an indication that it's checking, especially if you type in a name which is used then change it to something else which is used. Could we have

 1. type in username, indicator says 'checking,...' (spinner?)
 2. message says username is bad
 3. type in new name, indicator says 'checking...' again
 4. message says OK

If the username is good the first time, go straight to 4

andrebola:
I made this changes on https://github.com/MTG/freesound/commit/e34ac04cb3541af6c1825a2ef1de7d14e15708aa

Issue #794, opened 2017-01-31
Copy previews and uploaded sounds to secondary location


ffont:
For our internal testing purposes, we need to be able to have sound previews and sound original files duplicated in a different disk volume. In order to do that, previews should be copied to the alternative volume at the moment these are generated, and original files should also be copied to that volume once they are described.

PR #795, opened 2017-01-31
Mirror previews sounds

ffont:
See issue https://github.com/MTG/freesound/issues/794 for description.

### DIFF
--- a/accounts/views.py
+++ b/accounts/views.py
@@ -65,6 +65,7 @@ from bookmarks.models import Bookmark
 from messages.models import Message
 from oauth2_provider.models import AccessToken
 from follow import follow_utils
+from utils.mirror_files import copy_files_to_mirror_locations
 
 
 audio_logger = logging.getLogger('audio')
@@ -516,8 +517,9 @@ def describe_sounds(request):
                 logger.info("Moved original file from %s to %s" % (sound.original_path, new_original_path))
                 sound.original_path = new_original_path
                 sound.save()
+
             # Copy to mirror location
-            copy_sound_file_to_mirror_location(new_original_path)
+            copy_files_to_mirror_locations(sound, ['SOUNDS'])
 
             # Set pack (optional)
             pack = forms[i]['pack'].cleaned_data.get('pack', False)
@@ -803,18 +805,6 @@ def handle_uploaded_file(user_id, f):
         logger.warning("failed writing file error: %s", str(e))
         return False
     return True
-
-
-def copy_sound_file_to_mirror_location(source_path):
-    base_destination_path = settings.MIRROR_DISK_LOCATIONS['SOUNDS']
-    if base_destination_path is not None:
-        destination_path = source_path.replace(settings.SOUNDS_PATH, base_destination_path)
-        try:
-            os.makedirs(os.path.dirname(destination_path))
-        except OSError:  # I.e. path already exists
-            pass
-        shutil.copy2(source_path, destination_path)
-        logger.info('Copied sound file to mirror location %s' % destination_path)
 
 
 @csrf_exempt
--- a/accounts/views.py
+++ b/accounts/views.py
@@ -516,6 +516,8 @@ def describe_sounds(request):
                 logger.info("Moved original file from %s to %s" % (sound.original_path, new_original_path))
                 sound.original_path = new_original_path
                 sound.save()
+            # Copy to mirror location
+            copy_sound_file_to_mirror_location(new_original_path)
 
             # Set pack (optional)
             pack = forms[i]['pack'].cleaned_data.get('pack', False)
@@ -801,6 +803,18 @@ def handle_uploaded_file(user_id, f):
         logger.warning("failed writing file error: %s", str(e))
         return False
     return True
+
+
+def copy_sound_file_to_mirror_location(source_path):
+    base_destination_path = settings.MIRROR_DISK_LOCATIONS['SOUNDS']
+    if base_destination_path is not None:
+        destination_path = source_path.replace(settings.SOUNDS_PATH, base_destination_path)
+        try:
+            os.makedirs(os.path.dirname(destination_path))
+        except OSError:  # I.e. path already exists
+            pass
+        shutil.copy2(source_path, destination_path)
+        logger.info('Copied sound file to mirror location %s' % destination_path)
 
 
 @csrf_exempt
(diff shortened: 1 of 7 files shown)


freesound/settings.py
```diff
@@ -259,6 +259,13 @@
 # local settings if needed ;)
 DATA_URL = "/data/"
 
+# Locations where sounds and previews will be mirrored (if specified)
+# If locations do not exist, they will be created
+MIRROR_DISK_LOCATIONS = {
+    'SOUNDS': None,  # mirrors contents of SOUNDS_PATH
```
alastair: Do we want just one or multiple locations?

freesound/settings.py
```diff
@@ -259,6 +259,13 @@
...
+    'PREVIEWS': None,  # mirrors contents of PREVIEWS_PATH
+}
+
```
alastair: It'd be a good idea to have a setting here to turn it on and off, and check the setting in the copy util

accounts/views.py
```diff
@@ -803,6 +805,18 @@ def handle_uploaded_file(user_id, f):
     return True
 
 
+def copy_sound_file_to_mirror_location(source_path):
```
alastair: Can we put both copy functions in the same place in a util-y file? (so we don't have one in a view and one in a management command)

freesound/settings.py
alastair: (I see now the implementation does nothing if the data is None. Are you happy with that to turn on and off? If so, document this in the comments

freesound/settings.py
ffont: would be nice but lets implement this only if we need it

freesound/settings.py
ffont: yeah that was the idea to turn on and off, I'll document :)

accounts/views.py
ffont: there is very similar code executed after sound uploads (after describe, after csv upload and after api upload). I think this deserves a deeper refactoring so I'll leave as it is for now. I'll open an issue for that refactoring.

alastair: If you want to leave it like this, OK, but to me it seems like only a little work to make a start on this by creating a `sound_mirror` or `sound_upload` util file

ffont:
> This copies sounds and previews but not analysis or user avatars. Do we want to make a complete copy of all data?

In fact you're right according to our last conversations we also need it at least for displays (waveforms and spectrograms). Doing it for analysis would be also easy as it all happens async.

For avatars we would need to do it in the view for now, so its not best solution (as in for sounds) but could work.

accounts/views.py
ffont: Well,  now that I will implement the copying for displays I'll do it.

andrebola:
Maybe I don't understand this completely but wouldn't be easier to use rsync to make a copy of the files and then just configure in django the other location?

ffont:
the problem with rsync is that there are a A LOT of files and then it takes almost one hour only to do the comparison, so the syncing is not immediate. We need this mirroring to use it in "real-time", so we can completely replace one volume with another without problems or even load balance traffic between them (if this is possible).

ffont:
I implemented the requested refactorings and extra functionalities .
What's missing is mirroring of avatars, but I think we do not need that for now.
Mirroring all sound content should be enough.

sounds/management/commands/csv_bulk_upload.py
```diff
@@ -119,6 +120,9 @@ def handle(self, *args, **options):
                 sound.original_path = new_original_path
                 sound.save()
 
+            # Copy to mirror location
+            copy_files_to_mirror_locations(sound, ['SOUNDS'])
```
bdejong: The usage of magic string variables is kind of ugly. You could misspell them and python will not tell you anything. You could write "SOUDNS".

Perhaps this could be an idea:

*settings.py*

```
MIRROR_SOUNDS = "SOUNDS"
MIRROR_PREVIEWS = "PREVIEWS"
MIRROR_DISPLAYS = "DISPLAYS"
MIRROR_ANALYSIS = "ANALYSIS"

MIRROR_DISK_LOCATIONS = {
    MIRROR_SOUNDS: None,
    MIRROR_PREVIEWS: None,
    MIRROR_DISPLAYS: None,
    MIRROR_ANALYSIS: None
 }
```

*csv_bulk_upload.py*

```
copy_files_to_mirror_locations(sound, [settings.MIRROR_SOUNDS])
```

Event nicer would be using enums: https://docs.python.org/3/library/enum.html
They were back-ported to python 2.X: https://pypi.python.org/pypi/enum34

utils/mirror_files.py
```diff
@@ -0,0 +1,62 @@
...
+                    source_destination_tuples.append((
+                        source_path,
+                        source_path.replace(settings.ANALYSIS_PATH, destionation_base_path)
+                    ))
+
+    # Do the actual copying of the files
+    for source_path, destination_path in source_destination_tuples:
+        try:
+            os.makedirs(os.path.dirname(destination_path))
+        except OSError:  # I.e. path already exists
```
alastair: A little thing - this `except` has to have more checks on it: http://stackoverflow.com/a/600612

utils/mirror_files.py
```diff
@@ -0,0 +1,66 @@
...
+    copy_files_to_mirror_locations(sound, ['path'], settings.SOUNDS_PATH, settings.MIRROR_SOUNDS)
+
+
+def copy_previews_to_mirror_locations(sound):
+    copy_files_to_mirror_locations(
+        sound, ['preview.HQ.mp3.path', 'preview.HQ.ogg.path', 'preview.LQ.mp3.path', 'preview.LQ.ogg.path'],
+        settings.PREVIEWS_PATH, settings.MIRROR_PREVIEWS)
+
+
+def copy_displays_to_mirror_locations(sound):
```
alastair: `display`

utils/mirror_files.py
```diff
@@ -0,0 +1,66 @@
...
+    copy_files_to_mirror_locations(
+        sound, ['display.spectral.L.path', 'display.spectral.M.path', 'display.wave.L.path', 'display.wave.M.path'],
+        settings.DISPLAYS_PATH, settings.MIRROR_DISPLAYS)
+
+
+def copy_analysis_to_mirror_locations(sound):
```
alastair: `analysis`

utils/mirror_files.py
alastair: oh, never mind

Issue #796, opened 2017-01-31
Generate download link endpoint


ffont:
For a number of use cases such as required by the integration with Audio Commons Ecosystem, Freesound needs to be able to provide a link that can be used to download a file without user/app authentication (i.e. without OAuth2 or token authentication). To solve this problem the idea is to add a "download link" API endpoint which generates a url that can be used for such purpose. This generated  url has a finite lifetime. The "download link" endpoint needs to be accessed using OAuth2 just as the normal download endpoint requires.

PR #797, opened 2017-01-31
Download tokens

ffont:
See description in issue https://github.com/MTG/freesound/issues/796


apiv2/models.py
```diff
@@ -105,3 +102,20 @@ def client_secret(self):
     @property
     def version(self):
         return "V2"
+
+
+class DownloadToken(models.Model):
```
bdejong: Why not using something like JWT tokens for this? That way you don't need to store anything in the DB!

As the length of the token in this case is irrelevant you can dump all the information you need in the token JSON.

Sign it with a private key and you can check the signature when you receive the token again.

apiv2/models.py
```diff
@@ -105,3 +102,20 @@ def client_secret(self):
...
+    client_id = models.ForeignKey(ApiV2Client, null=True)
+    token = models.CharField(max_length=255, db_index=True)
+    expires = models.DateTimeField()
+    created = models.DateTimeField(auto_now_add=True)
+    sound = models.ForeignKey('sounds.Sound')
+
+    def is_expired(self):
+        if not self.expires:
+            return True
+        return timezone.now() >= self.expires
```
bdejong: This is an inherent property of JWT tokens: they have a date validity.

ffont:
Hmm sure, makes total sense, I guess I did not know about JWT and I didn't think about this option.
I'll have a look. Any recommendation for a python library? I see there are a number...

bdejong:
https://github.com/jpadilla/pyjwt/

alastair:
We use a home-made variation of this for delete sound links (and maybe some other things too). If we think that it's a good idea, should we use the same tool in all places in FS?

apiv2/views.py
```diff
@@ -983,6 +1003,27 @@ def post(self, request,  *args, **kwargs):
...
+@permission_classes([])
+def download_from_token(request, token):
+    try:
+        token_contents = jwt.decode(token, settings.SECRET_KEY, leeway=10)
+    except jwt.ExpiredSignatureError:
+        raise UnauthorizedException(msg="This token has expried.")
+    except jwt.InvalidTokenError:
+        raise UnauthorizedException(msg="Invalid token.")
+    try:
+        sound = Sound.objects.get(id=token_contents.get('sound_id', None))
```
alastair: single or double quotes?

alastair:
Looks OK to me. We talked in person about if we wanted to log a download to the downloads table (not now)

We should consider also putting in the AC spec that a download token should be submitted as a header too, so that services who don't want to read the parameter from the URL have a way of doing auth

apiv2/views.py
```diff
@@ -54,14 +54,14 @@
     from collections import OrderedDict
 except:
     from freesound.utils.ordered_dict import OrderedDict
-from urllib import unquote, quote
+from urllib import quote
```
bdejong: unneeded import

apiv2/views.py
bdejong: the last rules I read were: always use single everywhere, except if your string contains a single quote, then use double. :-)

apiv2/views.py
```diff
@@ -482,6 +482,26 @@ def get(self, request,  *args, **kwargs):
...
+    __doc__ = 'Get a url to download a sound without authentication.'
+
+    def get(self, request, *args, **kwargs):
+        sound_id = kwargs['pk']
+        logger.info(self.log_message('sound:%i get_download_link' % (int(sound_id))))
+        try:
+            sound = Sound.objects.get(id=sound_id, moderation_state="OK", processing_state="OK")
+        except Sound.DoesNotExist:
+            raise NotFoundException(resource=self)
+        download_token = jwt.encode({
```
bdejong: some newlines to separate logical blocks would be nice ;)

apiv2/models.py
```diff
@@ -24,10 +24,7 @@
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.contrib.sites.models import Site
-
-
-# Monkeypatch __unicode__ in provider.oauth2.models.Client class as it is not very informative (and it shows better in the admin)
-#Client.__unicode__ = lambda self: u'%s, %s, %s' % (self.name, self.user, self.client_id) #new_unicode
+from django.utils import timezone
```
bdejong: Unused import

apiv2/views.py
ffont: `quote`? `quote` is used. I removed `unquote` because pycharm highlighted it as not used, but in fact these lines are unrelated with the changes in the PR, is one of these code "tidy ups" that I should have probably committed separately but that, hmm, not too bad is it? ;)

apiv2/views.py
ffont: sure, never got to be consistent with this. I always try to use single quote, then some strings come which have single quotes inside and I try to continue using single quote and backslash the quotes inside. Then  I got tired and I switch to double quote again. Then I just forget about what did I decide...

PR #805, opened 2017-02-13
License upstream

andrebola:


ffont:
@alastair are you planning to have a look to this since the proposal was yours? (if I remember well)

alastair:
I'll get to it on thursday

freesound/urls.py
```diff
@@ -66,6 +66,7 @@
     url(r'^people/(?P<username>[^//]+)/packs/(?P<pack_id>\d+)/delete/$', sounds.views.pack_delete, name="pack-delete"),
     url(r'^people/(?P<username>[^//]+)/packs/(?P<pack_id>\d+)/download/.*$', sounds.views.pack_download, name="pack-download"),
     url(r'^people/(?P<username>[^//]+)/packs/(?P<pack_id>\d+)/downloaders/$', sounds.views.pack_downloaders, name="pack-downloaders"),
+    url(r'^secret_retrieve/people/(?P<username>[^//]+)/packs/(?P<pack_id>\d+)/licenses/$', sounds.views.pack_licenses, name="pack-licenses"),
```
alastair: I don't think this needs to have a `secret_retrieve` prefix. The `/secret` prefix for sounds is related to our use of `X-Accel-Redirect`, which maps to a file on disk. This URL is served from django, so I'm happy for it to be 

    /people/x/packs/y/license

utils/downloads.py
```diff
@@ -31,7 +32,7 @@
 from sounds.models import License
 
 
-def download_sounds(sounds_list, sounds_url=None):
+def download_sounds(sounds_list, sounds_url=None, licenses_url=None):
```
alastair: Is this method used anywhere other than the `packs_download` method? We shouldn't have optional arguments if we don't need this method somewhere called with these arguments as `None`

sounds/views.py
```diff
@@ -297,7 +298,26 @@ def pack_download(request, username, pack_id):
...
+    return download_sounds(pack_sounds, sounds_url, licenses_url)
+
+
+def pack_licenses(request, username, pack_id):
+    pack = get_object_or_404(Pack, id=pack_id)
+    sounds_list = pack.sound_set.filter(processing_state="OK",
+            moderation_state="OK").select_related('user', 'license')
+    users = User.objects.filter(sounds__in=sounds_list).distinct()
+    sounds_url = reverse('pack', args=[username, pack_id])
+    licenses = License.objects.all()
```
alastair: This list of licenses should be only the licenses which are in sounds of this pack, not all available licenses

sounds/views.py
```diff
@@ -297,7 +298,26 @@ def pack_download(request, username, pack_id):
...
+    sounds_url = reverse('pack', args=[username, pack_id])
+    return download_sounds(pack_sounds, sounds_url, licenses_url)
+
+
+def pack_licenses(request, username, pack_id):
+    pack = get_object_or_404(Pack, id=pack_id)
+    sounds_list = pack.sound_set.filter(processing_state="OK",
+            moderation_state="OK").select_related('user', 'license')
+    users = User.objects.filter(sounds__in=sounds_list).distinct()
+    sounds_url = reverse('pack', args=[username, pack_id])
```
alastair: This has to be an absolute url, with domain

utils/downloads.py
```diff
@@ -56,7 +58,7 @@ def download_sounds(sounds_list, sounds_url=None):
 
     filelist = "%02x %i %s %s\r\n" % (license_crc,
                                     os.stat(tmpf.name).st_size,
```
alastair: There is still code here regarding a temp file - I thought the reason for doing this change was so that we didn't need a temp file? We should be able to get filesize based on `len(attribution.encode('UTF-8'))`, right?

utils/downloads.py
```diff
@@ -31,18 +32,19 @@
...
     From a list of sounds generates the HttpResponse with the information of
     the wav files of the sonds and a text file with the license. This response
     is handled by mod_zipfile of nginx to generate a zip file with the content.
     """
     users = User.objects.filter(sounds__in=sounds_list).distinct()
     # Generate text file with license info
-    licenses = License.objects.all()
+    license_ids = sounds_list.values('license_id').distinct()
+    licenses = License.objects.filter(id__in=license_ids).all()
     attribution = render_to_string("sounds/pack_attribution.txt",
```
alastair: We get this data and render the template twice in this code, once here and once in the `pack_licenses` view. Should we have a util method (as part of the `Pack` class?) so that we only have to make changes in one place?

utils/downloads.py
```diff
@@ -38,26 +33,10 @@ def download_sounds(sounds_list, licenses_url, pack):
...
-            licenses=licenses,
-            sound_list=sounds_list))
-
-    tmpf = tempfile.NamedTemporaryFile(dir=settings.PACKS_PATH, delete=False)
-    tmpf.write(attribution.encode("UTF-8"))
-    tmpf.close()
-    secret_name = tmpf.name.replace(settings.PACKS_PATH, settings.PACKS_SENDFILE_URL)
-
-    os.chmod(tmpf.name, stat.S_IWUSR | stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH)
+    attribution = pack.get_attribution()
```
alastair: no pack argument.

It'd be good to see some tests for these methods to catch these errors!

PR #806, opened 2017-02-15
Add 4 graphs to monitor app

andrebola:
Created a page: "/monitor/stats" with the following information:

Line Charts with:
 - Downloads last week
 - Sounds from last week
 - Users registered last week
Also added Graph showing most used tags last week

### DIFF
--- a/media/js/stats.js
+++ b/media/js/stats.js
@@ -1,0 +1,149 @@
+$( document ).ready(function() {
+
+  $.get(dataUrl, function(d){
+      loadTagGraph(d);
+      
+      // Display charts with Downloads, Uploads and Registers
+      displayCharts('.users', d.new_users, 'Users');
+      displayCharts('.uploads', d.new_sounds, 'Sounds');
+      displayCharts('.downloads', d.new_downloads, 'Downloads');
+  });
+});
+
+function truncate(str, maxLength, suffix) {
+  if(str.length > maxLength) {
+    str = str.substring(0, maxLength + 1); 
+    str = str.substring(0, Math.min(str.length, str.lastIndexOf(" ")));
+    str = str + suffix;
+  }
+  return str;
+}
+
+function loadTagGraph(data){
+  // Display most used tags with bubbles
+  var margin = {top: 20, right: 200, bottom: 0, left: 20},
+    width = 500,
+    height = 650;
+
+  var c = d3.scaleOrdinal(d3.schemeCategory20c);
+
+  var xScale = d3.scaleTime()
+    .range([0, width]);
+  xScale.domain(d3.extent(data.tags_stats, function(d) { return new Date(d['day']); }));
+
+  var xAxis = d3.axisTop(xScale);
+
+  var formatDate = d3.timeFormat("%d-%m");
+  xAxis.tickFormat(formatDate);
+
+  var svg = d3.select('.tags').append("svg")
+    .attr("width", width + margin.left + margin.right)
+    .attr("height", height + margin.top + margin.bottom)
+    .style("margin-left", margin.left + "px")
+    .append("g")
+    .attr("transform", "translate(" + margin.left + "," + margin.top + ")");
+
+  svg.append("g")
+    .attr("class", "x axis")
+    .attr("transform", "translate(0," + 0 + ")")
+    .call(xAxis);
+
+  var rScale = d3.scaleLinear()
+      .domain([0, d3.max(data.tags_stats, function(d) { return d['tag_id__count'] })])
+      .range([2, 9]);
+
+
+  for (var j = 0; j < data.tags_stats.length; j++) {
+    var g = svg.append("g").attr("class","journal");
+
+    var circles = g.append("circle")
+      .attr("cx", xScale(new Date(data.tags_stats[j]['day'])))
+      .attr("cy", j*20+20)
+      .attr("r", rScale(data.tags_stats[j]['tag_id__count']))
+      .style("fill", function(d) { return c(j); });
+
+
+    var text = g.append("text")
+      .attr("y", j*20+25)
+      .attr("x", xScale(new Date(data.tags_stats[j]['day'])))
+      .attr("class","value")
+      .text(data.tags_stats[j]['tag_id__count'])
+      .style("fill", function(d) { return c(j); })
+      .style("display","none");
+
+    g.append("text")
+      .attr("y", j*20+25)
+      .attr("x",width+20)
+      .attr("class","label")
+      .text(truncate(data.tags_stats[j]['tag__name'],30,"..."))
+      .style("fill", function(d) { return c(j); })
+      .on("mouseover", mouseover)
+      .on("mouseout", mouseout);
+  };
+
+  function mouseover(p) {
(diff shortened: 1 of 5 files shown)


monitor/views.py
```diff
@@ -89,11 +98,41 @@ def monitor_home(request):
...
+def stats_ajax(request):
+    last_week = datetime.datetime.now()-datetime.timedelta(weeks=1)
+
+    new_sounds = sounds.models.Sound.objects.filter(created__gt=last_week)\
+            .extra(select={'day': 'date(created)'}).values('day').order_by()\
+            .annotate(Count('id'))
+
+    new_users = User.objects.filter(date_joined__gt=last_week)\
+            .extra(select={'day': 'date(date_joined)'}).values('day')\
+            .order_by().annotate(Count('id'))
```
bdejong: shouldn't we differentiate between activated and non-activated?

monitor/views.py
```diff
@@ -89,11 +98,41 @@ def monitor_home(request):
...
     return render(request, 'monitor/monitor.html', tvars)
 
 
+@cache_page(60 * 60 * 24)
+def stats_ajax(request):
+    last_week = datetime.datetime.now()-datetime.timedelta(weeks=1)
+
+    new_sounds = sounds.models.Sound.objects.filter(created__gt=last_week)\
+            .extra(select={'day': 'date(created)'}).values('day').order_by()\
+            .annotate(Count('id'))
```
bdejong: Perhaps we should take the moderation and processing state into account?

monitor/views.py
```diff
@@ -89,11 +98,41 @@ def monitor_home(request):
...
+
+    new_users = User.objects.filter(date_joined__gt=last_week)\
+            .extra(select={'day': 'date(date_joined)'}).values('day')\
+            .order_by().annotate(Count('id'))
+
+    new_downloads = sounds.models.Download.objects.filter(created__gt=last_week)\
+            .extra({'day':"date(created)"}).values('day').order_by().annotate(Count('id'))
```
bdejong: This counts both pack and sound downloads, are you sure you want that?

bdejong: For quotes, please use either " or ' but mixing them is a bit ugly... A good rule seems to be: always use ' except if your string has a single quote in it. I.e. 'This is a great string' but "I've got a single dash in here".

monitor/views.py
```diff
@@ -89,11 +98,41 @@ def monitor_home(request):
...
+
+    top_tags = TaggedItem.objects.filter(created__gt=last_week)\
+            .values('tag_id').distinct().annotate(num=Count('tag_id'))\
+            .order_by('num')[:30]
```
bdejong: `order_by('-num')` I would guess?

templates/base.html
```diff
@@ -42,6 +42,7 @@
         <script type="text/javascript" src="{{ media_url }}html_player/player.js"></script>
         {% google_analytics %}
         {% endblock %}
+        {% block extra-head %}{% endblock %}
```
bdejong: Don't do this... use {{block.super}}: https://docs.djangoproject.com/en/1.7/topics/templates/

templates/monitor/stats.html
```diff
@@ -0,0 +1,57 @@
+{% extends "base.html" %}
+
+{% block extra-head %}
+<style type="text/css">
+  .tags{font-family: Arial, sans-serif;font-size:10px;}
```
bdejong: please choose readability over this format.

also, make sure your formatting is consistent.

monitor/urls.py
```diff
@@ -18,17 +18,16 @@
 #     See AUTHORS file.
```
bdejong: General comment: please run pylint over your code... It will help you follow PEP8 and apply a bunch more useful rules...

monitor/views.py
ffont: This could be different lines maybe in the same plot, makes sense.

monitor/views.py
ffont: Again this could be two lines, one with is_active=True, the other is_active=False. In this way we will quickly spot if there are problems with activation at some point.

monitor/views.py
ffont: Again, lets separate packs and sounds into two lines ;)

monitor/views.py
ffont: Another idea could be to reuse code to generate a tag cloud like we have in user pages. It should be easy.

monitor/urls.py
andrebola: done on 411d93e

monitor/views.py
andrebola: I think the tag cloud doesn't add more information to the one graph I'm using, if you prefer I could replace it

media/js/stats.js
```diff
@@ -0,0 +1,206 @@
+$( document ).ready(function() {
+
+  $.get(dataUrl, function(d){
+      loadTagGraph(d);
+      console.log(d)
+      var active_users = [],
+      non_active_users = [];
+      for (var j = 0; j < d.new_users.length; j++) {
```
bdejong: you could replace this with d.new_users.forEach( function (element) { } ) - maybe it's nicer looking...

monitor/views.py
```diff
@@ -89,11 +98,58 @@ def monitor_home(request):
...
 
 
+@cache_page(60 * 60 * 24)
+def stats_ajax(request):
+    last_week = datetime.datetime.now()-datetime.timedelta(weeks=169)
+
+    new_sounds_mod = sounds.models.Sound.objects\
+            .filter(created__gt=last_week, moderation_date__isnull=False)\
+            .extra(select={'day': 'date(moderation_date)'}).values('day')\
+            .order_by().annotate(Count('id'))
```
bdejong: I might be wrong, but doesn't this need a groupBy somehow?

andrebola: The way of doing a group by in django is by using annotate after .values:

http://stackoverflow.com/a/629691/611467

bdejong: 👍  thx for the explanation!

ffont:
I'm merging this pull request so we can deploy and "test" it in production. In test or localhost we don't have as much data so it is harder to test. If there more comments, things to do we should probably reopen this pull request after the merge.

ffont:
Stats are now online at http://freesound.org/monitor/stats/
Comments:

- I'd be nice if graphs take take the full width and have ~half the height (so you can see ~2 together in an average screen)
- Maybe we could extend the period to 2 weeks? given that the queries are cached...
- The tag graph I'm not sure I understand, I think there are some problems as the dates doesn't seem to be right :

screen shot 2017-02-22 at 10 42 24

If possible maybe a total could would also be good (next to the tag name)...

andrebola:
@ffont if you put the mouse over the tag name then you see the number on each day. But I think there is something wrong when displaying the tags, I will take a look.

alastair:
Should this page be restricted to admins only?

alastair:
It would be nice to see tooltips with values on the graphs if possible.

Can we also add day of week to the x axis?

Issue #808, opened 2017-02-28
Add more statistics to monitor/stats page


ffont:
We recently introduced a nice statistics page for Freesound only accessible by admins. @andrebola added some stats that he considered interesting, but we could add other interesting stats. The goal of this issue is to brainstorm a bit about which stats could be added and then add them ;)

The following are new stats to add which are "single numbers", so these could be shown in a table at the top of the stats page, no need for evolution plots (although the computation of these should also be cached):

- N registered users `User.objects.filter(is_active=True).count()`
- N uploaders `User.objects.filter(is_active=True, profile__num_sounds__gt=0).count()`
- N sounds (moderated+processed ) `Sound.objects.filter(processing_state="OK", moderation_state="OK").count()`
- N packs `Pack.objects.all().count()`
- % of sounds inside a pack (total sounds with pack non null/total sounds)
- N downloads `Download.objects.all().count()` + avg n downloads per sound
- N comments `Comment.objects.all().count()` + avg n comments per sound
- N ratings `Rating.objects.all().count()` + avg n ratings per sound
- N tags `Tag.objects.all().count()`
- N tag applications `TaggedItem.objects.all().count()` + avg tag applications per sound
- N posts in forums `Post.objects.all().count()`
- N threads in forums `Thread.objects.all().count()`

Also we could add new "graph" stats for "number of donations in the last week" and maybe including amount donated?

Also I think that maybe we can set a cronjob that retrieves the monitor/stats page once a day early in the morning, so the numbers are already computed the first time an admin loads it ;)

andrebola:
The graph that currently shows number of Downloads could be improved since the number of sounds is much bigger than the number of packs so it's imposible to see a variation in the packs, it would be better to either show the packs in a different graph or instead of showing the number of packs we could graph the number of sounds downloaded in the form of packs (sum the number of sounds on each downloaded pack).

The tags graphs is not very clear, currently shows the 30 most used tags on new sounds of the last week, I think it would be better to show a line for each tag showing the position for each day, something like this:
![f15e22b4-3b7f-11e5-9496-12b6d811f0ea](https://cloud.githubusercontent.com/assets/991305/23403762/5a4ff1a2-fd90-11e6-8fad-d547de6659ed.jpg)

It would be interesting to display also the most used tags of the downloaded sounds in the last week.

[...]

PR #809, opened 2017-03-01
More stats issue #808

andrebola:


monitor/views.py
```diff
@@ -106,7 +108,7 @@ def monitor_home(request):
 
 @cache_page(60 * 60 * 24)
 def sounds_stats_ajax(request):
-    last_week = datetime.datetime.now()-datetime.timedelta(weeks=1)
+    last_week = datetime.datetime.now()-datetime.timedelta(weeks=2)
```
ffont: I think all these variables called `last_week` which now cover 2 weeks should be renamed. Maybe something more generic like `time_period` or `time_span` or something like this.

templates/monitor/stats.html
```diff
@@ -36,28 +36,56 @@
...
+      <li><b>Comments:</b> <span id="total-comments"></span></li>
+      <li><b>Average comments per sound:</b> <span id="avg-comments"></span></li>
+      <li><b>Ratings:</b> <span id="total-ratings"></span></li>
+      <li><b>Average ratings per sound:</b> <span id="avg-ratings"></span></li>
+      <li><b>Tags:</b> <span id="total-tags"></span></li>
+      <li><b>Tags instances:</b> <span id="total-used-tags"></span></li>
+      <li><b>Posts:</b> <span id="total-posts"></span></li>
+      <li><b>Threads:</b> <span id="total-threads"></span></li>
+    </ul>
+
```
ffont: I would consider putting this into a `<table>` instead of `<ul>`. In this way we will probably be able to copy/paste to spreadsheets and things like that. Minor thing but might be worth.

templates/monitor/stats.html
```diff
@@ -36,28 +36,56 @@
...
+
+    </div>
+ 
     <h2>Users</h2>
-    <h3>Registered users of last week</h3>
+    <h3>Registered users on lasts week</h3>
```
ffont: "last two weeks"?
Check also in the other `<h3>`'s

andrebola:
I don't think I can set the width to 100% since we use this value to calculate ranges of the graphs, but the width is fixed to the maximum. 
About the averages I calculated it as: Number of Comments / Number of sounds. And the same for ratings and downloads.

Other than that the first list of the issue #808 is done. 
Now I will start with:

- Change Packs graph for number of sounds in pack
- try other graphs to display tags, use cloud of tags to display the most used tags
- display most used tags on downloaded sounds (we currently show the most common tags on uploaded sounds)
- See how to get most frequent query terms from graylog
- Add totals of donations

ffont:
Cool, I'll merge this PR then. I'll copy the "missing features" to the issue for easier reference.

PR #820, opened 2017-03-17
Donation popup

andrebola:
Changed popup from branch banner to follow rules on utils/downloads.py, 

The popup will be shown X days after de user made a donation, if the user made more that Y downloads in a period of Z days. Also the popup will be shown a maximum of W times a day.


freesound/settings.py
```diff
@@ -275,6 +281,10 @@
 MIRROR_ANALYSIS = None  # list of locations to mirror contents of SOUNDS_PATH, set to None to turn off
 MIRROR_AVATARS = None  # list of locations to mirror contents of AVATARS_PATH, set to None to turn off
 
+# Footer banner
+FOOTER_BANNER_CONTENTS = \
+    '<a href="javascript:void(0);" onclick="openSurveyPage();hideFooterBanner();setSurveyVisited();">' \
+    '<img src="/media/images/content/freesound-survey-2017.png"/></a>'  # Set to None for no banner
```
bdejong: HTML in the settings seems rather ugly...

freesound/settings.py
```diff
@@ -171,6 +171,12 @@
 MAX_UNMODERATED_SOUNDS_IN_HOME_PAGE = 5
 ALLOWED_AUDIOFILE_EXTENSIONS = ['wav', 'aiff', 'aif', 'ogg', 'flac', 'mp3']
 
+# DONATION POPUP
+POPUP_DOWNLOADS_IN_PERIOD = 1
```
bdejong: some explanation about what these mean would be nice. perhaps in comments?

media/js/freesound.js
```diff
@@ -126,3 +126,64 @@ function show_hide_bookmark_form(id)
     }
     $("#bookmark_form_"  + id.toString()).toggle();
 }
+
+// FOOTER BANNER
+function hideFooterBanner(){
+    $('#fotterBanner').hide(200);
```
bdejong: what's "fotter"? "footer" perhaps?

media/js/freesound.js
```diff
@@ -126,3 +126,64 @@ function show_hide_bookmark_form(id)
...
+function hideModal(){
+    $("#fsmodal").hide();
+}
+
+function openModal(){
+    $("#fsmodal").show();
+}
+
+function generateModalHTML(title, contents){
+    return '<div id="fsmodal" class="modal"><div class="modal-content"><div class="modal-header"><span onclick="hideModal();" class="close">&times;</span>'
```
bdejong: Make sure that this stuff is compatible with screen-readers: "onclick" stuff sometimes isn't. And we have plenty of blind users. Doing this in Javascript seems also a bit messy. Why not add the code in the page default hidden?

media/js/freesound.js
```diff
@@ -126,3 +126,64 @@ function show_hide_bookmark_form(id)
...
+        + title + '</div><div class="modal-body">' + contents + '</div></div></div>';
+}
+
+function createAndOpenModal(show_modal_url, title, content){
+    hideModal();
+    var numberDownloads = $.cookie("numberDownloads");
```
bdejong: You could use a session variable for this rather than a cookie, but... no preference...

sounds/views.py
```diff
@@ -244,6 +245,16 @@ def sound(request, username, sound_id):
     is_explicit = sound.is_explicit and (not request.user.is_authenticated() \
                         or not request.user.profile.is_adult)
 
+    # Show modal asking users for donations after clicking in download
```
bdejong: This is kind of messy... keep the templating stuff inside the templates. You are doing string-insertion by python string-mashing. That's quite bad...

freesound/settings.py
andrebola: @ffont I'm not sure what to do with the code of the banner on this branch, should I remove it or they might be used at the same time?

media/js/freesound.js
andrebola: We use the cookie to store the downloads of a user in a period of one day, I don't know if it's a good idea to store that in a session variable

freesound/settings.py
ffont: yeah that's all my fault I know ;)
maybe we should have waited for my first pull request to be approved before building more stuff on top of it. I don't know, I guess I thought it was the best place to put that but I also felt a bit weird. Any suggestion where should we place that?

media/js/freesound.js
ffont: The idea was to have it as generic as possible and easy to reuse, like you can just call this function anywhere and it will generate the modal...

sounds/views.py
ffont: I don't understand the comment, do you mean the `is_explicit...` line?

media/js/freesound.js
andrebola: Now the modal is available anywhere and the content can be set on template or with js.

sounds/views.py
andrebola: Now the content of the modal is set on template so after_download_modal is not used anymore

freesound/settings.py
andrebola: I would place it on base.html since is the same to change that file or settings.py,

ffont: sure, makes sense!
I can do this change too for the survey code, which I hope to release soon (this week)

bdejong:
@ffont / @andrebola so... this is the only PR I should look at now and ignore the other one (the one with the popup for donations)?

ffont:
yeah but I'm still working on it!

ffont:
Ok I now added all the changes that I wanted. Now this pull request mainly includes two features:

- Display a banner in the footer whose contents are set on base.html
- Display a modal after user downloads

The modal part is configurable so that the modal loads dynamically the content to be shown from a view and the view decides whether to actually show any content and what content to show. Currently the after download modal can be configured to show an advertisement for the Freesound survey 2017 (it is set to that), or to suggest user's making a donation (and following some configurable rules to decide when to show the popup). 

For now we only want to show the survey popup, and maybe in a few weeks we start with the donation campaign. The parameters for how often should the donations modal be shown will be an object for discussion (Andres implemented it in a way that it can be configured but we might want to add other variables). However, these should not be the point of discussion in this PR as we need to merge it "as soon as possible" so we release the survey.

Hope this is all clear :)

accounts/migrations/0008_auto_20170331_1702.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
bdejong: What does this file do in the PR?

freesound/settings.py
```diff
@@ -171,6 +171,17 @@
 MAX_UNMODERATED_SOUNDS_IN_HOME_PAGE = 5
 ALLOWED_AUDIOFILE_EXTENSIONS = ['wav', 'aiff', 'aif', 'ogg', 'flac', 'mp3']
 
+# After download modal
+AFTER_DOWNLOAD_MODAL = 'survey'  # Set it to none for no modal after download
```
bdejong: This kind of calls for an ENUM or pre-defined constants...

Magic strings nooooh!

media/js/freesound.js
```diff
@@ -126,3 +125,51 @@ function show_hide_bookmark_form(id)
...
+
+
+// After download banner
+
+function afterDownloadModal(show_modal_url, sound_name){
+    hideModal(); // Hide the modal just in case it was shown
+
+    // Keep track of number of downloads this day
+    var numberDownloads = $.cookie("numberDownloads");
+    if (numberDownloads === undefined) {numberDownloads = 0;}
```
bdejong: Always nicer, in a nice javascript-y way:

var numberDownloads = $.cookie("numberDownloads") || 0;

sounds/views.py
```diff
@@ -253,12 +254,45 @@ def sound(request, username, sound_id):
...
+    elif settings.AFTER_DOWNLOAD_MODAL == 'survey':
+        title = "Participate in the Freesound survey!"
+        contents = """
+        Thanks for downloading <b>%s</b>!<br />
+        Please, consider participating in the <a href="javascript:void(0);" onclick="openSurveyPage();hideModal();
+        hideFooterBanner();setSurveyVisited();">Freesound Survey 2017</a> if you have not participated yet :)
+        """ % sound_name
+        if request.COOKIES.get('surveyVisited', 'no') != 'yes':
+            show = True
+    return JsonResponse({"show": show, "title": title, "contents": contents})
```
bdejong: Why, instead of returning a JSON object and then doing the display in Javascript don't you just respond with a regular HTML page, rendered by the rendering engine of Django? That way you'd remove all of this string-parsing from the Python code. TBH this is very messy.

utils/downloads.py
```diff
@@ -53,3 +54,30 @@ def download_sounds(licenses_url, pack):
...
+    user. This will be based on 3 settings indicating how many days
+    after a donation we show the popup again, after how many downloads we
+    display the popup and for how long. The popup will be shown a number of
+    times per day.
+    """
+
+    if times_displayed and times_displayed > settings.DONATION_MODAL_DISPLAY_TIMES_DAY:
+        return False
+
+    last_donation = user.donation_set.order_by('created').last()
```
bdejong: I always find it annoying if variables are created 6 lines before they are used. This line immediately made me think: hey, what if the user never donated. Suggestion: move this to before the if.

utils/downloads.py
```diff
@@ -53,3 +54,30 @@ def download_sounds(licenses_url, pack):
...
+    period = datetime.datetime.now()\
+            - datetime.timedelta(days=settings.DONATION_MODAL_DOWNLOAD_DAYS)
+    num_downloads_in_period = Download.objects.filter(user=user,
```
bdejong: So, the number of downloads are counted in a cookie, but on top of that we're also doing a downloads-count via the Database on each get() triggered by each download? It seems to me this either warrants doing one OR the other, but not both.

accounts/migrations/0008_auto_20170331_1702.py
ffont: Hmmm good question. It went as follows: 
1) I checked out donations popup branch and merged master into it because I knew there were db changes and otherwise I could not work on donations popup branch without reverting db changes locally
2) Django complained there were differences with the models and db so I run make migrations and generated this file. This should have been run while I was on the other branch and commit there, but I guess I mis-thought this was because some side-effect of the merge of master to donations popup branch and I run it here and committed/pushed.

So human mistake here. However to me this raises interesting questions: how to deal with branches with different db "versions"?

freesound/settings.py
ffont: Sure, I kind of forget abut these often... fixed in https://github.com/MTG/freesound/pull/820/commits/a9e4df7e7f9acd995393a36e52ede59768a4de05

media/js/freesound.js
ffont: Fixed in https://github.com/MTG/freesound/pull/820/commits/a0c044a6ca758b9c75906e450bd74e0ef2b12fec

sounds/views.py
ffont: Agree, changed to return the html content rendered in Django (https://github.com/MTG/freesound/pull/820/commits/b77f221eecb91a181a9d32b14c5d9493498fa1ba)

utils/downloads.py
ffont: Fixed

utils/downloads.py
ffont: I'd say there was a problem with the naming. What the cookie was supposed to count was the number of times the modal was shown, not really the number of downloads. I, https://github.com/MTG/freesound/pull/820/commits/cef84553675c026d7b28c25ca8413209246ece85 I refactored this code and renamed stuff so now I think its less confusing.

ffont:
I made some extra changes according to @bdejong's comments.
I'd like to merge this today so we release the survey. The specific parameters for deciding when to show the donations modal can be decided somewhere else.

bdejong:
Will review ASAP

accounts/migrations/0008_auto_20170331_1702.py
bdejong: I thought django always provided backward and forward migrations...? Shouldn't this be enough in almost all cases?

sounds/views.py
```diff
@@ -253,12 +254,33 @@ def sound(request, username, sound_id):
...
+def after_download_modal(request):
+    """
+    This view checks if a modal should be shown after the user has downloaded a sound, and returns either the contents
+    of the modal if needed.
+    """
+    response = HttpResponse()  # Default empty response with no content
+    sound_name = request.GET.get('sound_name', 'this sound')  # Gets some data sent by the client
+
+    if settings.AFTER_DOWNLOAD_MODAL == settings.AFTER_DOWNLOAD_MODAL_DONATION:
+        times_shown_in_last_day = int(request.GET.get('num_times_shown', 0))
```
bdejong: This will crash if num_times_shown is not convertable to int

utils/downloads.py
```diff
@@ -53,3 +54,29 @@ def download_sounds(licenses_url, pack):
...
+        # If modal has been shown more than settings.DONATION_MODAL_DISPLAY_TIMES_DAY times, don't show it again today
+        return False
+
+    donation_period = datetime.datetime.now() - datetime.timedelta(days=settings.DONATION_MODAL_DAYS_AFTER_DONATION)
+    last_donation = user.donation_set.order_by('created').last()
+    if not last_donation or last_donation.created < donation_period:
+        # If there has never been a donation or last donation is older than settings.DONATION_MODAL_DAYS_AFTER_DONATION,
+        # check if the number of downloads in the last settings.DONATION_MODAL_DOWNLOAD_DAYS days if bigger than
+        # settings.DONATION_MODAL_DOWNLOADS_IN_PERIOD. If that is the case, show the modal.
+        num_downloads_in_period = Download.objects.filter(
```
bdejong: This query returns a queryset while the `if` below requires a number. There is a count( ) missing here.

accounts/migrations/0008_auto_20170331_1702.py
ffont: Yes it does, but sometimes this might imply destroying data etc or might take a lot of time for changes in models with large number of rows. I'm not saying it is not possible, just that it's not very practical.

sounds/views.py
ffont: Fixed in https://github.com/MTG/freesound/pull/820/commits/b70d4f2407a427f288da5a041c72cbef7d5cddcb.

utils/downloads.py
ffont: Fixed in https://github.com/MTG/freesound/pull/820/commits/00a2c207a695c5049c489d8cbccb5d4c766aa099

ffont:
Hey,
I added a new bunch of changes. Answering to Bram's comments:

> I have a sneaky suspicion "afterDownloadModal" (javascript) and "should_suggest_donation" (python) could both use some more testing...

I updated the test for `should_suggest_donation` in https://github.com/MTG/freesound/pull/820/commits/6c6faa78b8a3825c790fb761475ff808f32a0dec. Now it is more comprehensive. I think this is probably enough for now.

> I think it's also bad form to mix multiple features in PR's and rather use PR's as feature branches to implement one feature. But... too late for that!

I'm not sure I understand what you mean. However I agree that PRs should not have more than one feature in them (unless these are very minimal and can be grouped, like "small layout chaeges"). In that case it's clear that there was a problem because one PR started building on top of another yet unreviewed/unmerged PR. We should have probably waited until first PR was "done" before starting with 2nd, but the two features were similar and I guess it seemed reasonable to do it this way. So yes, mistake, but we do try to make single-feature PRs.

> For the donation modal stuff I'm still unsure why we're using this mix of javascript and Python. You could have just stored something in the cache, for example and do everything in python.

Agree, I refactored (https://github.com/MTG/freesound/pull/820/commits/b496116e5206959524056804f5f01470be5bf621) now we do all in the server.

> After download -> always show modal
Show modal -> returns some chosen http error code if should not be shown / returns http 200 if it should be shown.

Well this is very close to how it actually works. Return empty response if nothing should be shown, return html content for the modal if modal should be shown. I in fact first implemented by returning http status 204 (no content) if no modal needed to be shown, but jquey interprets 204 as success (it is in fact success) so I still had to check for the status in the response to decide whether to show it or not (also I thought that 204's definition did not perfectly match our case). Now I reimplemented it to return http 404 (https://github.com/MTG/freesound/pull/820/commits/a34a1cd0d1949ea07ac29c77eabcc5797172d1a7), which sounds like the most appropriate but to me it's still not quite right.

PR #826, opened 2017-03-24
Changed donations form to indicate amount and monthly donation

andrebola:
I added an input field on donations form to enter the amount, also added a checkbox to indicate monthly recurrent donations.

Also at this page it would be good to give a brief explanation about why is a good idea to donate and what we are planning to do, I think @ffont will do this after the survey.


andrebola:
Since PR #827 is merged, now this PR should be discarded

ffont:
It was set to merged automatically, I guess because #827 includes the commits from this one.

andrebola:
Yes, that's OK. Now this PR and the branch change-donations-form should be deleted

PR #827, opened 2017-03-24
Change donations form v2

andrebola:
Here I propose a different donation form than PR #826, it changes on styles and elements locations but the behaviour is the same


ffont:
I added some more layout changes, as I did not like the "titles".
But good job, I'll merge :)

PR #828, opened 2017-03-27
Email notification settings for users

andrebola:
issue #743


general/management/commands/gm_worker_async_tasks.py
```diff
@@ -95,7 +96,11 @@ def task_email_random_sound(self, gearman_worker, gearman_job):
         random_sound_id = gearman_job.data
         random_sound = Sound.objects.get(id=random_sound_id)
 
-        send_mail_template(\
+        disabled = UserEmailSetting.objects.filter(user=random_sound.user,
+                                email_type__name="random_sound").count()
```
bdejong: I know this is not easy, but I think we need a helper for this... Using raw strings for definite things always seems risky... You could have a utility function that throws an exception for example:

UserEmailSetting.objects.filter(user=random_sound.user, email_type=get_email_type("random_sound")).count()

so that at least the string is "checked at runtime"...

bdejong: Even better, the functionality will always be the same:
if not disabled, send email.
So this should really be one function call:

if email_not_disabled(user, "something"):  # make this throw errors if "something" is an unknown email type!
    send_mail_template(...)

accounts/fixtures/email_type.json
```diff
@@ -0,0 +1 @@
+[{"model": "accounts.emailtype", "pk": 1, "fields": {"description": "Email notification sent when user's sound is selected as random sound of the day", "name": "random_sound", "display_name": "Selected sound as random sound of the day"}}, {"model": "accounts.emailtype", "pk": 2, "fields": {"description": "Email notification sent when user's receives a private message", "name": "private_message", "display_name": "New private message received"}}, {"model": "accounts.emailtype", "pk": 4, "fields": {"description": "Email notification sent when user's receives a comment on a sound", "name": "new_comment", "display_name": "New comment on sound"}}]
```
ffont: I guess this fixture is generated by django's `manage.py dumpdata app.Model > fixture.json` command.  Given that this fixture is small, using `--indent 4` as parameter should do the trick and make it more readable.

accounts/forms.py
```diff
@@ -327,3 +319,18 @@ def __init__(self, *args, **kwargs):
...
 
+
+class EmailSettingsForm(forms.Form):
+    enabled_stream_emails = forms.BooleanField(
+        help_text="Receive weekly stream update email notifications (only when new sounds are uploaded by "
+                  "users you follow or that have tags you follow)",
+        label = "",
+        required=False
+    )
+    wants_newsletter = forms.BooleanField(help_text="Subscribed to newsletter", label="", required=False)
```
ffont: Agree!

accounts/forms.py
```diff
@@ -327,3 +319,18 @@ def __init__(self, *args, **kwargs):
                 }
         super(DeleteUserForm, self).__init__(*args, **kwargs)
 
+
+class EmailSettingsForm(forms.Form):
+    enabled_stream_emails = forms.BooleanField(
```
ffont: I guess this is related to the main comment. Stream email preferences should be merged with these, agree (so stream emails should become a new email type).

andrebola: The default value for stream email is False so that means we will start with a lot of entries in the UserEmailSettings table. given that we have other fields regarding stream email I thought we could leave this separate, but if you want I can change it.

ffont: ok I see, then we should support both behaviours (default send and default not send). This is basically done in the logic code before sending an email, but it should be reflected in preferences definition somehow. Maybe a good option is that the email types that we have include the action. So we could rename the EmailType to EmailPreferenceType and then make the names like "UnmuteStreamEmails", "MuteSoundCommentsEmails". I think this makes sense.

andrebola: What if we add a field to EmailPreferenceType with the default value? e.g it would be false for StreamEmails and true for the others.

ffont: To what is now `EmailType` right?. That sounds ok, could be called `send_by_default` or something like this.

accounts/models.py
```diff
@@ -118,6 +116,65 @@ def locations(self):
             )
         )
 
+    def email_not_disabled(self, email_type_name):
+        # Raise exception if the email_type doesn't exists
+        email_type = EmailPreferenceType.objects.get(name=email_type_name)
+
+        # when send_by_default == invert_default means email is disabled
+        invert_default = UserEmailSetting.objects.filter(user=self.user,
+                email_type=email_type).count()
```
bdejong: Would be nicer to use exists() rather than count: https://docs.djangoproject.com/en/1.9/ref/models/querysets/#exists

accounts/models.py
```diff
@@ -118,6 +116,65 @@ def locations(self):
...
+        return email_type.send_by_default != invert_default
```
bdejong: Ooooooh, DANGER DANGER:

```python
>>> True == 1
True
>>> True == 2
False
```

Comparing bools to ints is a BAAAAAAD idea :-) By using exists( ) you will get a bool, so it should be fine.

accounts/models.py
```diff
@@ -118,6 +116,65 @@ def locations(self):
...
+        email_preferences = self.user.email_settings.values('email_type__id')
+        # if email_type not in email_preferences then default value must be True
+        not_disabled = all_emails.exclude(id__in=email_preferences)\
+                .filter(send_by_default=True)
+
+        # if email_type in email_preferences then default value must be False
+        enabled = all_emails.filter(id__in=email_preferences,
+                send_by_default=False)
+
+        return list(enabled) + list(not_disabled)
```
bdejong: Minor nitpick: for these kind of things using set( ) is better

accounts/models.py
```diff
@@ -118,6 +116,65 @@ def locations(self):
...
+
+    def update_enabled_email_types(self, email_type_ids):
+        # Update user's email_settings from the list of enabled email_types
+
+        # First get current value of stream_email to know if
+        # profile.last_stream_email_sent must be initialized
+        had_enabled_stream_emails = self.user.email_settings\
+                .filter(email_type__name='stream_email').count()
```
bdejong: pls use `email_type = EmailPreferenceType.objects.get(name='stream_email')` here as well...

accounts/models.py
```diff
@@ -278,7 +335,21 @@ def create_user_profile(sender, instance, created, **kwargs):
...
         profile.save()
 
 post_save.connect(create_user_profile, sender=User)
+
+
+class EmailPreferenceType(models.Model):
+    description = models.TextField(max_length=1024, null=True, blank=True)
+    name = models.CharField(max_length=255)
+    display_name = models.CharField(max_length=255)
+    send_by_default = models.BooleanField(default=True)
```
bdejong: some help text would be interesting for this field

PR #844, opened 2017-04-12
Migration from django 1.9 to django 1.11

andrebola:
We are migrating to django 1.11 which is LTS release, with support until 2020

### DIFF
--- a/bookmarks/views.py
+++ b/bookmarks/views.py
@@ -18,7 +18,7 @@
 #     See AUTHORS file.
 #
 
-from django.shortcuts import render_to_response, get_object_or_404
+from django.shortcuts import render, get_object_or_404
 from django.template import RequestContext
 from django.contrib.auth.models import User
 from sounds.models import Sound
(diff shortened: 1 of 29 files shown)


bookmarks/views.py
```diff
@@ -36,41 +36,49 @@
...
-    
-    return render_to_response('bookmarks/bookmarks.html', combine_dicts(locals(),paginate(request, bookmarked_sounds, 30)), context_instance=RequestContext(request))
+
+    tvars = {
+        'bookmark_categories': bookmark_categories,
+        'category': category,
+        'n_uncat': n_uncat,
+        'user': user,
+    }
+    tvars = combine_dicts(tvars, paginate(request, bookmarked_sounds, 30))
```
alastair: I don't like `combine_dicts`, can we just use `tvars.update(paginate(....))` instead?

bdejong: combine_dicts should really only be used in functional chains. As { }.update() doesn't return anything it can't be used in that way, but in this case you don't need it...

andrebola:
After updating the version on PR #845 we will remove the use of local( ) method

PR #845, opened 2017-04-12
Migration from django 1.9 to django 1.11

andrebola:
We are migrating to django 1.11 which is LTS release, with support until 2020


andrebola:
I think this is ready to review/merge, since the migration to django 1.11 is complete. If we want to change all the use of locals() for a different syntax we can create a new issue for that.

Issue #802, opened 2017-02-10
Add API client usage statistics/monitor page


ffont:
Currently users of the freesound API have no idea of the usage they're making of the API unless they manually log it. It would be good to provide a page with statistics of API usage for the last days/weeks. We could simply provide a graph of number of requests per day (including a separation between normal requests and post requests (which have a different rate limit)). We cal also show the number of throttled requests of we have it.
Additionally it would be great to show plot of resources accessed.

All of these should be rather simple to implement by querying the graylog API. Graylog computes all the data that we need.

PR #847, opened 2017-04-18
Add new stats page for api usage #802

andrebola:
the link to this page is missing, maybe it could be on developers help page?


monitor/views.py
```diff
@@ -104,6 +105,21 @@ def queries_stats_ajax(request):
...
 
+@login_required
+def api_usage_stats_ajax(request):
+    query = urllib.urlencode({
+        'query': 'api_client_username:%s' % (request.user.username),
+        'range': '604800',
+        'filter': 'streams:530f2ec5e4b0f124869546d0',
+        'interval': 'day'
+    })
+    req = urllib2.Request("http://mtg-logserver.s.upf.edu/graylog/api/search/universal/relative/histogram?%s" % (query))
```
alastair: can we use `requests` here, for the same reason in the paypal query stuff - better retry/error catching capabilities, and an API to pass username/passwords

monitor/views.py
```diff
@@ -104,6 +105,21 @@ def queries_stats_ajax(request):
     return JsonResponse(json.load(urllib2.urlopen(req)))
 
 
+@login_required
+def api_usage_stats_ajax(request):
+    query = urllib.urlencode({
+        'query': 'api_client_username:%s' % (request.user.username),
+        'range': '604800',
+        'filter': 'streams:530f2ec5e4b0f124869546d0',
```
alastair: magic number - if something goes wrong with our graylog instance and we have to reinstall it, this will probably change

monitor/views.py
```diff
@@ -104,6 +105,21 @@ def queries_stats_ajax(request):
...
+        'interval': 'day'
+    })
+    req = urllib2.Request("http://mtg-logserver.s.upf.edu/graylog/api/search/universal/relative/histogram?%s" % (query))
+    base64string = base64.b64encode('%s:%s' % (settings.GRAYLOG_USERNAME,
+        settings.GRAYLOG_PASSWORD))
+    req.add_header("Authorization", "Basic %s" % base64string)
+    return JsonResponse(json.load(urllib2.urlopen(req)))
```
alastair: also consider error cases here, where the query fails, and return an appropriate error message

ffont:
I think the description of the issue was a bit misunderstood. The idea is not to have general stats for API usage, but to be able to show stats of a particular API client to the developers. The idea is that in the API credentials apply page (http://www.freesound.org/apiv2/apply) we add a link in each row to "monitor" which goes to something like http://www.freesound.org/apiv2/apply/credentials/CLIENT_ID/monitor/ and shows a graph of API usage of that particular client (only visible by the user that owns the client and by superusers). We could also add a line of the rate limit per application.

I think it will be easy to adapt the code however, as you implemented almost all what's needed to make it work per client.

andrebola:
@ffont this shows the api usage for the currently logged user

monitor/views.py
```diff
@@ -97,11 +96,31 @@ def monitor_home(request):
...
+        'range': 14 * 60 * 60 * 24,
+        'filter': 'streams:530f2ec5e4b0f124869546d0',
+        'interval': 'day'
+    }
+    req = requests.get('http://mtg-logserver.s.upf.edu/graylog/api/search/universal/relative/histogram',
+            auth=(settings.GRAYLOG_USERNAME, settings.GRAYLOG_PASSWORD), params=params)
+    try:
+        return JsonResponse(req.json())
+    except requests.HTTPError:
+        return JsonResponse({'error': 'Error at collecting usage stats'})
```
bdejong: This should send the correct http code... At least a 500 or something.

templates/api/monitor_api_credential.html
```diff
@@ -0,0 +1,75 @@
...
+              tickEvery: d3.timeDay.every(1),
+              legendData: [{color: 'crimson', name: 'requests'}, {color: 'black', name: 'rate limit'}]
+            }, {});
+        } else {
+          $('.api-usage').html('<p>There are no requests in the given period.</p>')
+        }
+      });
+
+    </script>
+    <script src="https://d3js.org/d3.v4.min.js" type="text/javascript"></script>
```
bdejong: Is this the right way to use these? I.e. does D3 say "go ahead use these urls"?

monitor/views.py
```diff
@@ -97,11 +96,31 @@ def monitor_home(request):
...
 def queries_stats_ajax(request):
-    req = urllib2.Request("http://mtg-logserver.s.upf.edu/graylog/api/search/universal/relative/terms?query=%2A&range=1209600&filter=streams%3A531051bee4b0f1248696785a&field=query")
-    base64string = base64.b64encode('%s:%s' % (settings.GRAYLOG_USERNAME,
-        settings.GRAYLOG_PASSWORD))
-    req.add_header("Authorization", "Basic %s" % base64string)
-    return JsonResponse(json.load(urllib2.urlopen(req)))
+    params = {
+        'query': ':',
+        'range': 14 * 60 * 60 * 24,
+        'filter': 'streams:531051bee4b0f1248696785a',
```
bdejong: I mirror Alastair's comment, is there not a better way to query this from greylog rather than a magic string?

templates/api/monitor_api_credential.html
andrebola: It says you can use it here: https://d3js.org/

> To link directly to the latest release, copy this snippet: 
> ...



I prefer to have an external dependency if it's reliable since will be faster than FS (because of all the problems we have regarding IO), and it's probable that users have it cached.

We could also use: https://cdnjs.com/libraries/d3

monitor/views.py
andrebola: hmm I thought the comment was about the 'range' attribute, now I understand...

templates/api/monitor_api_credential.html
bdejong: :+1:

monitor/views.py
bdejong: I think it would be best to take out that string "API....." and put it in the preferences....

ffont:
@andrebola are you planning on making more changes or is this ready for re-review/merging?
We should follow some kind of convention for this situations in which changes are requested. I guess the easiest way is to simply write a comment when something is ready for reviewing again.

andrebola:
I was waiting for comments regarding d3 as external import, other than that I think is ready.
From now on I will let you know if I'm not working on the PR anymore.

bdejong:
if d3 says you can load from their page all is well!

bdejong:
btw, https://github.com/MTG/freesound/pull/848 is no longer building on travis because of the ipython requirements change. Also see #849

Issue #610, opened 2014-04-09
Do not show average rating for sounds rated less than N times


ffont:
Sound ratings should only be shown for these sounds that have been rated more than N times, where N could be something like 3.

PR #851, opened 2017-04-21
Do not show average rating for sounds rated less than N times

andrebola:
#610

Now when the sound has less than N ratings, we don't display the average. 
Note: If you go to the page of a sound the number of ratings  is still shown (but not the average).


sounds/models.py
```diff
@@ -417,7 +417,9 @@ def duration_ms(self):
         return self.duration * 1000
 
     def rating_percent(self):
-        return int(self.avg_rating*10)
+        if self.num_ratings > settings.MIN_NUMBER_RATINGS:
```
bdejong: Very small style comment. IMHO early return should also be done on the easier path.

```
If self.num_ratings <= ...:
     return 0

return ....
```

.travis.yml
```diff
@@ -16,6 +16,7 @@ before_install:
 - mkdir -p ../freesound-data/uploads
 - mkdir -p ../freesound-data/avatars
 install:
+- pip install -U pip
```
bdejong: This one is sneaking into multiple PRs.

andrebola: Yes, that will be a problem when merging. But if I don't add this it will fail on build. 
Any suggestions?

ffont:
I guess I should merge PRs faster ;)
What about a PR which only the change in the travis file which I can merge without even testing and then you merge back to the different PR branches?

andrebola:
I updated Travis file on PR #852

Issue #773, opened 2016-12-27
oEmbed functionality


waterjump:
Currently id-to-embed code functionality requires manual construction.  oEmbed is a standard for getting embed codes without the need for private API access or special parsing.

bdejong:
On every sound page there is the embed code which you can easily copy-paste. I don't know how implementing a relatively old standard from 2008 (which was never really widely adopted by oEmbed consumers) would make things better.

Could you give us some better insights, @waterjump ?

waterjump:
oEmbed is used to get the embed code programmatically using the direct link to the sound, rather than a human user navigating to it, copying the existing embed code.  I only suggested this because I'm tasked with embedding audio samples from Freesound, in addition to SoundCloud and YouTube, and I have to manually parse out the sound number.  Here's what it looks like in Ruby:

```ruby
if audio_sample.include?('freesound')
  freesound_id = /\d{6}/.match(audio_sample).to_s
  return unless freesound_id.present?
  "<iframe frameborder='0' scrolling='no' src='http://www.freesound.org/embed/sound/iframe/#{freesound_id}/simple/small/' width='375' height='30'></iframe>"
end``` 

oEmbed will also prevent code like this from breaking if sound ids change format, for example from 6 digits to 7.

bdejong:
Sean, sounds already have from 2 to 7 digits: https://freesound.org/people/Bram/sounds/11/ is the first sound in the database

ffont:
Implemented in https://github.com/MTG/freesound/issues/853

PR #853, opened 2017-04-24
Add oembed endpoint that returns iframe

andrebola:
issue #773

Follows specification on http://oembed.com/

The only required parameter is 'url' which corresponds to the sound e.g: https://freesound.org/people/Bram/sounds/11/

I also would like to specify somewhere how to use this but I'm not sure where should it go.


sounds/views.py
```diff
@@ -798,6 +798,28 @@ def embed_iframe(request, sound_id, player_size):
     return render_to_response('sounds/sound_iframe.html', locals(), context_instance=RequestContext(request))
 
 
+def oembed(request):
+    url = request.GET.get('url', None)
+    view, args, kwargs = resolve(urlparse(url)[2])
```
bdejong: What happens if URL=None?

sounds/views.py
```diff
@@ -798,6 +798,28 @@ def embed_iframe(request, sound_id, player_size):
...
+    if not 'sound_id' in kwargs:
+        raise Http404
+    sound_id = kwargs['sound_id']
+    sound = get_object_or_404(Sound, id=sound_id, moderation_state='OK', processing_state='OK')
+    player_size = request.GET.get('size', 'medium')
+    if player_size == 'large':
+        sizes = [920, 245]
```
bdejong: we should probably remove these from here and from the template and make them constants somewhere...

sounds/views.py
andrebola: I changed this, now returns 404 if 'url' is not present

bdejong:
👍

PR #856, opened 2017-04-25
Change all urls to graylog for http

andrebola:


monitor/views.py
```diff
@@ -125,7 +125,7 @@ def queries_stats_ajax(request):
 def api_usage_stats_ajax(request, client_id):
     try:
         auth = (settings.GRAYLOG_USERNAME, settings.GRAYLOG_PASSWORD)
-        req = requests.get('https://logserver.mtg.upf.edu/graylog/api/streams',
+        req = requests.get('http://mtg-logserver.s.upf.edu/graylog/api/streams',
```
alastair: I don't like how the host name is duplicated in this file many times, along with auth and error catching. Perhaps a single get helper method which takes a url and returns json or None/raise Exception would help us make changes in the future.

The logic to get a stream ID is duplicated twice in this file. Can we have a common function to get this data?

How long does it take to make these two queries? In my original comment I was saying that putting a stream id in the settings file would be enough. Now this means for every query we make 2 lookups, one of which almost never changes. I would consider putting the streamid back in the settings, or at the very least caching it.

andrebola: I agree, it's better to have this on settings

PR #857, opened 2017-04-26
Update RadioSelect render of choices value to dj11

andrebola:
On django 1.11 they changed how to access the widget from templates: https://groups.google.com/forum/#!topic/django-users/z1V_qwSsiPs This brokes how we set the initial value of the donation_type field (which is a radio choice).

I also changed how we access to cleaned_data on the form clean method


ffont:
scary...

Issue #691, opened 2015-11-09
Log in with email address


bdejong:
As a user I would like to log in with my email address so I don't have to remember my username.

As a security measure this should not enable people to check to see if someone has an account on freeosund. I.e. if you just put in an email address and a (wrong) password the error should not say: "your password was incorrect" as this verifies the email address.

alastair:
additionally, the password reset field should allow email addresses as well as usernames

alastair:
4000 users have usernames that look like email addresses (have a @ in them) but the email address is different to the username. I think we should check first the username, and then the email address.
We can adapt our existing auth backend (https://github.com/MTG/freesound/blob/master/accounts/modelbackend.py)

More problematic is that there are 200 email addresses shared between more than 1 user (it looks like no email address is used by more than 2 though, perhaps people forgot their username and created a new account?). We need to either remove them, change the address, or email the users and ask them to make a decision.
Then, we need to make the email address a unique column

bdejong:
That sucks. Can you see if some of those users are inactive? I did not know this, but it looks like django does not enforce email uniqueness:

https://github.com/django/django/blob/master/django/contrib/auth/models.py#L322

bdejong:
btw, you can -for most users- decide on the last_login date if they really need to be contacted: if they last logged in 5 years ago, there is really no need to contact them and we just delete the account.

PR #863, opened 2017-05-10
Email login

ffont:
Finally implementes login with email https://github.com/MTG/freesound/issues/691
Must be merged after https://github.com/MTG/freesound/pull/763


accounts/modelbackend.py
```diff
@@ -23,11 +23,32 @@
...
+    def authenticate(self, request, username=None, password=None, **kwargs):
         """ authenticate against case insensitive username """
-        try:
-            user = User.objects.get(username__iexact=username)
-            if user.check_password(password):
-                return user
-        except User.DoesNotExist:
-            return None
+
+        if '@' in username:
```
bdejong: I don't really understand all of this if/else stuff, especially as the login via username is in here twice (DRY!).

Can't you just do the try/except and default to username first?

```python
user = None
try:
   user = User.objects.get(username__iexact=username) # default to username
except User.DoesNotExist:
   try:
       user = User.objects.get(email__iexact=username) # try with email
   except User.DoesNotExist:
       pass

if user and user.check_password(password):
   return user

return None
```

ffont: The logic can be simplified and not have the same code twice, you're right. I did not do it because I tried to abstract the repeated part into a function but at the end it seemed to me the code was much less readable. However, if instead of abstracting into a function I just do one user lookup at the end I'll work the same. I committed that https://github.com/MTG/freesound/pull/863/commits/1fc30ee7084e04bd40650bc6b3a58b959139b7d0).

However, imagine the current situation: 

- User A has an account with username `user_a` and email `a@fs.org`
- User B has an account with username `a@fs.org` and email `whatever@fs.org`

If we check username first and only check email on the except case, User A won't be able to login using the email. Emails are valid usernames, but usernames are not valid emails if they don't have an @. The `if @ ...` clause is only to skip one db lookup when not using emails to log in.

accounts/modelbackend.py
```diff
@@ -23,11 +23,28 @@
 
 
 class CustomModelBackend(ModelBackend):
-    def authenticate(self, username=None, password=None):
+
+    def authenticate(self, request, username=None, password=None, **kwargs):
         """ authenticate against case insensitive username """
```
alastair: Update documentation. Include rationale for checking email addresses first.

accounts/modelbackend.py
```diff
@@ -23,11 +23,28 @@
...
+
+        if '@' in username:
+            # In this case, user is most probably using email+password pair to login
+            # Try to get user object from password and check password
+            try:
+                user = User.objects.get(email__iexact=username)
```
alastair: Do we have a case-insensitive index on the email field?

accounts/modelbackend.py
```diff
@@ -23,11 +23,28 @@
...
 
 class CustomModelBackend(ModelBackend):
-    def authenticate(self, username=None, password=None):
+
+    def authenticate(self, request, username=None, password=None, **kwargs):
         """ authenticate against case insensitive username """
+
+        if '@' in username:
+            # In this case, user is most probably using email+password pair to login
+            # Try to get user object from password and check password
```
alastair: "from password" ?

accounts/modelbackend.py
```diff
@@ -23,11 +23,28 @@
...
+            try:
+                user = User.objects.get(email__iexact=username)
+                if user.check_password(password):
+                    return user
+            except User.DoesNotExist:
+                # If user was not find by email, it could be that user who's logging in has
```
alastair: found

accounts/modelbackend.py
```diff
@@ -23,11 +23,28 @@
...
+        # Do normal username+password login check
         try:
             user = User.objects.get(username__iexact=username)
             if user.check_password(password):
                 return user
         except User.DoesNotExist:
-            return None
+            pass
+
+        return None
```
alastair: :+1:

accounts/forms.py
```diff
@@ -214,8 +214,13 @@ def __init__(self, *args, **kwargs):
         super(FsAuthenticationForm, self).__init__(*args, **kwargs)
         self.error_messages.update({
             'inactive': mark_safe(_("You are trying to log in with an inactive account, please <a href=\"%s\">activate "
-                                    "your account</a> first." % reverse("accounts-resend-activation")))
+                                    "your account</a> first." % reverse("accounts-resend-activation"))),
+            'invalid_login': _(
+                "Please enter a correct username/email and password. Note that both "
+                "fields may be case-sensitive."
```
alastair: both may be case-sensitive? what does this mean?
Emails and usernames are not. Passwords are, but can we take that for granted?

apiv2/urls.py
```diff
@@ -97,7 +99,8 @@
 
     # Oauth2
     url(r'^oauth2/', include('apiv2.oauth2_urls', namespace='oauth2_provider')),
-    url(r'^login/$', LoginView.as_view(template_name='api/minimal_login.html'), name="api-login"),
+    url(r'^login/$', login, {'template_name': 'api/minimal_login.html',
+                             'authentication_form': FsAuthenticationForm}, name="api-login"),
```
alastair: for my interest - does this mean that logins on the API have always been case-sensitive? What difference does this form have to the default one?

accounts/modelbackend.py
ffont: That I wanted to comment with you. I checked that and I think we do, but I'm not 100% sure.

accounts/forms.py
ffont: Usernames are emails are case-insensitive so I guess password are the only case-sensitive. I just copied the default message here and added the /email part without questioning the other part. Maybe we should just put "Note that passwords are case sensitive" (I just checked, they are).

apiv2/urls.py
ffont: Yes, this means that logins through API have always been case-sensitive. The difference is case-sensitiveness and (from now on) email option.

accounts/modelbackend.py
alastair: from looking at the table, it looks like we do. I still recommend that we get the sql query for this lookup (`str(User.objects.get(email__iexact=username).query)`) and do an `EXPLAIN` in an sql session to check that it uses the index.

ffont:
I implemented the small fixes that @alastair proposed. Should be ready for merging.

accounts/modelbackend.py
```diff
@@ -25,17 +25,26 @@
 class CustomModelBackend(ModelBackend):
 
     def authenticate(self, request, username=None, password=None, **kwargs):
-        """ authenticate against case insensitive username """
+        """ authenticate against case insensitive username or email
+        
+        In case there is an @ sign in the provided username field, it is likely that
+        the user is trying to authenticate by providing an email+password pair (rather than
+        username+password). In that case, we first try to authenticate using the email and
+        if it does not succeed we try to eithenticate using username. We do email first
```
alastair: authenticate

PR #865, opened 2017-05-11
Twitter cards

andrebola:
To make this work Twitter must validate the card, for doing so we need to open a request on: 

https://cards-dev.twitter.com/validator

This validation forms asks for some information about the twitter account and also sends an email to the related address for verification. 

Also, we have to enter an example so they can validate it, I was using the test server (authentication must be disabled first), with the sound:

https://test.freesound.org/people/andrebola3/sounds/274375/

### DIFF
--- a/media/html_player/embed.css
+++ b/media/html_player/embed.css
@@ -236,3 +236,52 @@ body.small {
     text-decoration: underline;
     color: #AB4646;
 }
+
+/* full_size */
+
+body.full_size {
+    height: 100%;
+    background-color: #FBFAFC;
+    border: 1px solid #F5F5F5;
+    border-top: 3px solid #c07879;
+}
+
+.widget.medium.full_size .player_wrapper {
+    float: left;
+    width: 120px;
+    position: absolute;
+    top: 25%;
+    padding: 8px;
+}
+
+.widget.medium.full_size .sample_info {
+    width: 72%;
+    float: right;
+    position: absolute;
+    right: 0;
+    top: 25%;
+    height: 100%;
+    background-color: #fff;
+    border: 1px solid #F5F5F5;
+}
+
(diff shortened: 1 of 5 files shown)


ffont:
Hey, I'm not sure how is this supposed to look like, but this is what I see when I load: https://test.freesound.org/embed/sound/iframe/274375/simple/full_size/

screen shot 2017-05-12 at 09 56 07

I guess what twitter want is that the embed can adapt to different sizes. I think to make it simpler and more visually attractive we should use as base the design of the large embed (https://test.freesound.org/embed/sound/iframe/274375/simple/large/):

screen shot 2017-05-12 at 09 57 41

Then we should resize the player to fit 100% of width (if this is possible).
I played a little bit around by modifying css values in the browser and I could get to what I had in mind by changing these things:

- Set width to `width:100%;` in `.widget.large .sample_info` and  `border-top: solid 2px #AB4646;`
- Set with to `width:100%;`, height to `height: 22vw;`, and `background-color: black;` in `.player.large`.
- Remove `margin-left` in `.widget.large .player.large` and reduce margin top to `margin-top: 4px;`
- Remove `margin-bottom` from `.player`.
- Remove background image from `body.large` (it is "replaced" by the red top border added in first point of this list)
- Edit background image properties of the waveform image so that:
```
background-image: url(https://test.freesound.org/data/displays/274/274375_5223450_wave_L.png);
background-size: contain;
background-repeat: no-repeat;
```

Examples (when they appear with full width you can only tell the difference by the size of the controls, but if you open them in their original size you'll see better):

screen shot 2017-05-12 at 10 17 34
screen shot 2017-05-12 at 10 17 50
screen shot 2017-05-12 at 10 17 58
screen shot 2017-05-12 at 10 18 05

PR #869, opened 2017-05-16
Roll back code of django-cors-headers

andrebola:
In order to make this code work we have to use MiddlewareMixin from
django.utils.deprecation.

This is not a definitive solution, we should use django-cors-headers as
an external dependency.


andrebola:
I tested this on test.freesound.org and work OK, if you see this error: 
```
'Access-Control-Allow-Origin' header contains multiple values '*, *', but only one is allowed
```

Check if nginx is not setting the header:
```
add_header Access-Control-Allow-Origin *;

ffont:
Did you actually find this error in test? Or you're mentioning it for some other reason.
IIRC, we use nginx to add this header for downloads and static files, API responses' headers are added only in Django.

Also, did you try to use the library as a dependency (https://github.com/ottoyiu/django-cors-headers) instead of bundling it ourselves? To me this is what we should do so unless there is a strong reason not to do it, I't go for it.

ffont:
Why not write a test to check cors with the current code and then switch to library to make sure it works nice? Maybe test a couple of urls (api and non-api).

ffont:
`corsheaders` is only adding the `ACCESS-CONTROL-ALLOW-ORIGIN` to `options` requests (which is what the test tests), but this should be fine. I'll merge and deploy.

Issue #837, opened 2017-04-04
email stream settings explanation & confirmation


bdejong:
![screen shot 2017-04-04 at 17 31 06](https://cloud.githubusercontent.com/assets/805269/24665073/9321907a-195c-11e7-8450-8d5d29ded27e.png)

1. The last setting should be: get a weekly update on all users and tags you follow.
2. when you save there is no indication that something actually saved.

andrebola:
1 -> Done

ffont:
@andrebola does your comment mean that you're working on it? ;)

andrebola:
No, I just changed that text on the admin.  
(sorry about the short comment) But I can do the second part if you want

ffont:
but is it committed somewhere?

PR #887, opened 2017-06-09
Redirect to home after changed settings

andrebola:
issue #837


ffont:
I complemented this PR according to issue #837.
Instead of redirecting to 'home' I added redirection to 'user settings' page (which is the page from which users click on 'edit email notification settings').

Issue #568, opened 2014-02-19
Remember last geotag location


qubodup:
Geotagging an amazingly unique feature of Freesound and its use should be encouraged (by making it more comfortable to use).

When geo-tagging a sound, the default location appears to be 23.885837699862005/21.796875 at a zoom level of 2.

Whenever a geotag is saved (when describing sounds after upload or on the edit page, the position and zoom levels should be saved and restored the next time a geo-tagging form is shown

PR #892, opened 2017-06-13
Use last geolocation as initial position for next uploaded sound

andrebola:
issue #568

### DIFF
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -326,6 +326,13 @@ class Profile(SocialModel):
         if commit:
             self.save()
 
+    def get_last_latlong(self):
+        lasts_sound_geotagged = Sound.objects.filter(user=self.user).exclude(geotag=None).order_by('-created')
+        if lasts_sound_geotagged.count():
+            last_sound = lasts_sound_geotagged[0]
+            return last_sound.geotag.lat, last_sound.geotag.lon, last_sound.geotag.zoom
+        return None
+
     class Meta(SocialModel.Meta):
         ordering = ('-user__date_joined', )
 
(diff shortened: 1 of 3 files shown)


ffont:
(please read issue description https://github.com/MTG/freesound/issues/568 before this comment)

The current implementation has the problem that sounds are always assigned a geotag.
Say user a uploads a sound with a geotag. Then he records and uploads another sound but does not want it to have a geotag. This second sound will be assigned the geotag of the first sound by default instead of "none".

In my opinion, the way to fix this issue is that we do not set the geotag by default until user clicks on "show geotag map" link. At that point, we can set the default lat/lon/zoom as the one which was used last time (be careful to only do this once for a given sound, otherwise every time the user clicks on "show geotag map" it would override the current selection).

The same thing can be applied when editing an existing sound. In that case we could also default to the last geotag, but here we always show the geotag map so there is no "show geotag map" button and we can't use the same solution as before to add the geotag. I think a good solution involves some sort of enable/disable geotag maybe by making the widget half transparent or something like this when not enabled. However this requires more design work and stuff that might change rather soon with the redesign we're carrying out, maybe it is not necessary to add the last geotag feature to the edit sound page.

andrebola:
I think this is not exactly what is happening, if the user uploads a new sound it will not have a geotag unless the user clicks on "Show Geotag map" when describing the sound.

But if the user clicks on "Show Geotag map" and then clicks on "Hide Geotag map", then the sound will have the geotag assigned to the last sound as you said. Note that this is the same behavior that we used to have since I only changed the initial value of the map.  

If you want I can change this so when the user clicks on "Hide Geotag map" we unset the geotag values, I think in that case we should change the text for "Hide Geotag map and unset value" (or something like that).

alastair:
>  in that case we should change the text for "Hide Geotag map and unset value" (or something like that).
Maybe "Remove Geotag", which removes the tag and hides the map

andrebola:
There is already a checkbox with the text Remove geotag

Issue #799, opened 2017-02-03
Refactor handling of sounds after description/upload


ffont:
When sounds are uploaded and described a number of things needs to be done like moving the file to another location, send to process, etc. The code for carrying out such operations is mostly duplicated in 3 places now:

- `accounts.views.describe_sounds`
- `sounds.management.commands.csv_bulk_upload`
- `apiv2.apiv2_utils.create_sound_object`

This code should be moved into a function called from these different places.
Be aware that the code is not completely the same, there might be some small differences ;)

PR #893, opened 2017-06-15
Move common code for upload new sound to utils

andrebola:
issue #799

### DIFF
--- a/sounds/management/commands/csv_bulk_upload.py
+++ b/sounds/management/commands/csv_bulk_upload.py
@@ -131,7 +131,7 @@ class Command(BaseCommand):
                 'license': licensef,
                 'pack': packnamef,
                 'description': descriptionf,
-                'tags': tagsf,
+                'tags': tagsf.split(),
                 'geotag': geotagf
             }
 
--- a/sounds/management/commands/csv_bulk_upload.py
+++ b/sounds/management/commands/csv_bulk_upload.py
@@ -28,7 +28,7 @@ from geotags.models import GeoTag
 from utils.filesystem import md5file
 from utils.text import slugify
 import shutil, os, csv
-
+import utils.sound_upload
 
 class Command(BaseCommand):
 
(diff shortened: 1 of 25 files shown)


sounds/views.py
```diff
@@ -461,6 +461,7 @@ def pack_edit(request, username, pack_id):
     if request.method == "POST":
         form = PackEditForm(request.POST, instance=pack)
         if form.is_valid():
+            pack.sound_set.all().update(is_index_dirty=True)
```
ffont: Why is this change here? Isn't this called wight below?

sounds/management/commands/csv_bulk_upload.py
```diff
@@ -124,106 +124,41 @@ def handle(self, *args, **options):
...
-                sound.pack.process()
+            except utils.sound_upload.NoAudioException:
+                continue
+            except utils.sound_upload.AlreadyExistsException:
+                print 'The file %s is already part of freesound, not uploading it' % (dest_path,)
+                continue
+            except utils.sound_upload.CantMoveException as e:
+                print e.message
 
-            print 'Successfully uploaded sound ' + sound.original_filename
```
ffont: Why did you remove the print? Is is somewhere else?

ffont:
@andrebola  I left a couple of comments.
It looks good but I'd like to be 100% this works well.

For example, is the new `is_explicit` thing considered here? (I guess not as the two PR are parallel)
We have a test of the `create_sound` util function and for the view to upload sounds (https://github.com/MTG/freesound/blob/master/accounts/tests.py#L549), but we are not testing the the apiv2 endpoint and the the csv bulk upload . I guess if we don't add the `is_explicit` field then by default `is_explicit` is set to False and that's it (which is OK), but do the other stuff work? I wanted to test manually but this branch now conflicts with master and I'm not sure how to fix it. Can you fix the conflicts so I can manually test the PR? It would be also good to add those missing tests although it might be more complex...

sounds/management/commands/csv_bulk_upload.py
andrebola: yes, it's on line 155

andrebola:
Merged with Master added is_explicit.  I think we don't have much tests for the api in general.

Issue #894, opened 2017-06-19
IntegrityError: new row for relation "sounds_sound" violates check constraint "sounds_sound_num_comments_check"


ffont:
https://logserver.mtg.upf.edu/sentry/freesound-web/issues/666/

```
IntegrityError: new row for relation "sounds_sound" violates check constraint "sounds_sound_num_comments_check"
DETAIL:  Failing row contains (393776, 3481306, 2017-05-23 11:00:10.914712+02, NPR Humpty or Ned.mp3, /home/fsweb/freesound/freesound-data/sounds/393/393776_3481306.m..., 393776__breeze87__npr-humpty-or-ned, humpy dumpy is going through a lot of changes, one of them is hi..., null, 2, null, 27528, mp3, 60.2405, 345, 0, 22050, 305546, 1, f1acabdcfa9ca7e09dd495cd5404e5df, t, OK, 2017-06-16 10:57:42.657304+02, null, f, OK, 2017-05-23 11:00:16.455642+02, null, OK, OK, -1, 0, 0, 0, f7f04eed, null, FI, f).

(23 additional frame(s) were not displayed)
...
  File "comments/models.py", line 48, in on_delete_comment
    instance.content_object.post_delete_comment()
  File "sounds/models.py", line 625, in post_delete_comment
    self.save()

IntegrityError: new row for relation "sounds_sound" violates check constraint "sounds_sound_num_comments_check"
DETAIL:  Failing row contains (393776, 3481306, 2017-05-23 11:00:10.914712+02, NPR Humpty or Ned.mp3, /home/fsweb/freesound/freesound-data/sounds/393/393776_3481306.m..., 393776__breeze87__npr-humpty-or-ned, humpy dumpy is going through a lot of changes, one of them is hi..., null, 2, null, 27528, mp3, 60.2405, 345, 0, 22050, 305546, 1, f1acabdcfa9ca7e09dd495cd5404e5df, t, OK, 2017-06-16 10:57:42.657304+02, null, f, OK, 2017-05-23 11:00:16.455642+02, null, OK, OK, -1, 0, 0, 0, f7f04eed, null, FI, f).

```

ffont:
This happened to me when trying to delete a user account which had made comments to some sounds and the `num_comments` property of the `sound` object did not match `sound.comments.all().count()`

alastair:
[...]

PR #895, opened 2017-06-19
Count objects when comments are deleted

andrebola:
issue #894

### DIFF
--- a/sounds/models.py
+++ b/sounds/models.py
@@ -613,13 +613,13 @@ class Sound(SocialModel):
         # From django 1.8.4 is required to save the object before adding the relation
         comment.save()
         self.comments.add(comment)
-        self.num_comments += 1
+        self.num_comments += self.comments.count()
         self.mark_index_dirty(commit=False)
         if commit:
             self.save()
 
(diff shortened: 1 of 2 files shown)


ffont:
The proposed fix should work but I'm a bit worried about how expensive would it be when deleting a lot of comments in bulk. What do you think @alastair ?

Maybe we need to start thinking of providing ways to do operations with a single object vs with a list of objects. Eg. delete a single sound vs delete all the sounds from a pack. In the first case we delete the sound and then update the pack. In the second case we could delete all sounds and then update the pack, but right now we're updating after each delete. This is just an example, unrelated to this PR and issue, but something to have in find for the future. I guess Django provides a way to deal with these things too.

alastair:
I don't like this as a fix because it doesn't try and address the underlying issue as to why there may be a mismatched count with the number of comments.

Additionally, looking at the code around Comments, I noticed a number of other things that could be cleaned up while we are working on them. I added these as a list of suggestions in the linked issue. 
I'm going to close this so that we can re-approach the problem in a way that tries to solve the cause of the problem.

Issue #676, opened 2015-06-25
Optimise username lookup in messages


ffont:
The username autocompletion function when composing a message seems to generate slow queries. Should investigate how to optimise this.

alastair:
This should be technically quite similar to https://github.com/MTG/freesound/issues/614

andrebola:
The difference with #614 is that the query filters all the previous contacted users, and also filters by '__starts_with' which is much slower. 

But for my user the query wasn't slow, @ffont can you confirm if this is still an issue?

ffont:
I'm not sure, it might depend on the user.
the filter is starts_with because we want to autocomplete, so I guess we can't do much about it.
maybe we should think of ways to test that, like try response times for different users and see if this is still a problem.

bdejong:
Is this a username completion in the private messages? Because if so
perhaps we should o my look through the users that you wrote to before.

Or, just drop the autocomplete. It's not a great feature.

bdejong:
My apologies. Should have read the description first...

ffont:
This is for private messages yes, it only does autocomplete for users you previously wrote (or wrote you). It is in fact rather useful according to that users said when we implemented it. However, what we could do is replace it by a the same username check in registration, so we do an ajax check that the username exists. At the end I think this is to avoid misspelling usernames. Or maybe we can have both, autocomplete for usernames contacted in the past and username check for others. We could simply compute the list of previously contacted users and send it to the template to do the js autocomplete completely in the client (although this list might get large...).

ffont:
The autocomplete works fine but the username check should not be done for an empty username.

PR #896, opened 2017-06-19
No checking of username on new message when is empty

andrebola:
Following with issue #676

### DIFF
--- a/templates/messages/new.html
+++ b/templates/messages/new.html
@@ -20,16 +20,18 @@
 
       // When the user enters a username we check if it's available
       $("#id_to").focusout(function(){
-        $('#spinner').show();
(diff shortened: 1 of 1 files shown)


ffont:
The fix works, but the code is super similar to https://github.com/MTG/freesound/blob/master/templates/accounts/registration.html#L14 (the username check on the registration page). Maybe this could be abstracted somehow? Moved in to a util function?

andrebola:
It's true that they call to the same API method but after that they do different things, since one checks that the user exist and the other checks that the user don't exist. So if we want to move this to a more general method at the end we are not winning much since we might reduce from 10 lines to 6 or 7 lines.

Issue #839, opened 2017-04-04
Choose sound of the day and send emails using a cron job


alastair:
Currently we choose the sound of the day on page load if the item does not exist in the cache. This means that for one unlucky person per day (or more, if we clear the cache), page load takes an additional 500ms (#593) and then an email send is triggered (to tell the user that their sound was chosen)

It's perhaps not a good idea to do this kind of processing on page load, and we should think about moving it to an external cron job so that 1) external users are not affected, and 2) we don't have a random view sending unrelated emails - we can conceptually group these together in a single command (https://github.com/MTG/freesound/commit/3ed41a746c475997a4a302175d2a60ead7bc6d06#commitcomment-21546471)

Some possible issues:

* If we don't perform a check on page load, there may be no sound of the day
* If we perform the cron job too infrequently then there might be a gap between the sound expiring from the cache and choosing a new one
* If the cache is cleared then there will be no sound

Some solutions
* Make the script choose tomorrow's sound instead, so we always have an extra one waiting
* Think about using Redis instead of memcache - redis can serialise to disk, and can also have separate namespaces. We could have a volatile namespace for things like page caches and a more permanent one for sound of the day
* Think about putting sound of the day history in the database - this lets us be smarter about not choosing the same sound twice, or sounds by the same author too often (so that they don't receive too many emails!)

bdejong:
This actually happens to multiple pages. /Sounds/ probably too...

On top of that I think you can have multiple regenerations at the same
time. I.e. cache miss -> 500ms -> multiple more misses.

It might be interesting to cron this in a generalized way... I.e. have one
cron which can generate multiple memcached-stored HTML fragments...

alastair:
In the case of the sound of the day only an sound ID is stored in memcache, we're not storing any rendered data. 

I only bought up the page caches because everything is stored in memcache in a single namespace. I was suggesting that we think about separating items which are completely stateless (such as the cache of a template, we can just recompute it at any stage), and things that require a bit more state (such as the sound of the day ID, which should theoretically stay the same for a 24h period...)

bdejong:
Aha, okay...

Btw the tricky part might be the cache warming... If we have parts of the
website which are only generated via a cron, what do we do when we start
from a cold cache? Not display those parts? Does that make sense for
everything?

ffont:
[...]

PR #897, opened 2017-06-19
Use new model to store the random sound of the day

andrebola:
issue #839


sounds/management/commands/create_random_sound.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def handle(self, *args, **options):
+        print 'Create new RandomSound task'
+        random_sound_id = Sound.objects.random()
+        sound = Sound.objects.get(id=random_sound_id)
+        rnd = RandomSound.objects.create(sound=sound)
+        gm_client = gearman.GearmanClient(settings.GEARMAN_JOB_SERVERS)
+        gm_client.submit_job("email_random_sound", str(rnd.id),
+                wait_until_complete=False, background=True)
+        print "Create new RandomSound task ended"
```
alastair: let's either use a logger here, or start thinking about python3 and use `__future__.print_function`

sounds/models.py
```diff
@@ -681,6 +681,11 @@ class Meta(SocialModel.Meta):
         ordering = ("-created", )
 
 
+class RandomSound(models.Model):
```
alastair: Is this a `RandomSound` or a `SoundOfTheDay`?

sounds/management/commands/create_random_sound.py
```diff
@@ -0,0 +1,41 @@
...
+
+class Command(BaseCommand):
+
+    help = 'Add new RandomSound'
+
+    def handle(self, *args, **options):
+        print 'Create new RandomSound task'
+        random_sound_id = Sound.objects.random()
+        sound = Sound.objects.get(id=random_sound_id)
+        rnd = RandomSound.objects.create(sound=sound)
```
alastair: I think we should use the database to verify that we only have a `RandomSound` for each day. Currently it will create a new entry every time we run this command

sounds/views.py
```diff
@@ -73,11 +73,9 @@ def get_random_sound():
     cache_key = "random_sound"
     random_sound = cache.get(cache_key)
     if not random_sound:
-        random_sound = Sound.objects.random()
+        rnd = RandomSound.objects.order_by('created').last()
```
alastair: Part of my comments in #839 was suggesting that we create sounds of the day in advance, so that if we "forget" to run this script, we still have some sounds for the next few days.

alastair: What happens if there are no `RandomSound` objects?

sounds/management/commands/create_random_sound.py
```diff
@@ -0,0 +1,41 @@
...
+        gm_client = gearman.GearmanClient(settings.GEARMAN_JOB_SERVERS)
+        gm_client.submit_job("email_random_sound", str(rnd.id),
```
alastair: I wonder if it's a good idea to have an extra flag on the model to say that the email has been sent. We should send the email at the beginning of the day that the user's sound becomes the sound of the day.

sounds/management/commands/create_random_sound.py
```diff
@@ -0,0 +1,41 @@
...
+from sounds.models import Sound, RandomSound
+
+
+class Command(BaseCommand):
+
+    help = 'Add new RandomSound'
+
+    def handle(self, *args, **options):
+        print 'Create new RandomSound task'
+        random_sound_id = Sound.objects.random()
```
alastair: Should we have some checks to make sure that we don't choose a sound by the same author more than once a month, and the same sound more than once every year? (https://github.com/MTG/freesound/issues/839#issuecomment-291824353)

ffont: Yes, now that we're at it I'd put a limit like 1 sound per user/month. Also we should filter out sounds which are explicit, and I would only add those with average rating > X and more than X votes.

ffont:
We should take this opportunity to add a couple more restrictions to the way we select the sound. Besides the restriction of sounds by the same user in a month, we should filter out sounds which are marked as `is_explicit`, flagged, and also remove those average rating < X or less than X votes.

For example:
```
Sound.objects.exclude(is_explicit=True).filter(flag=None, avg_rating__gt=6, num_ratings=3)
```

Also exclude sounds that were already chosen in the past.

alastair:
I made some changes and is now ready for second review - @ffont ?

ffont:
Once this is deployed we have to add a run of the `create_random_sound` command in crontab  right?

ffont:
The tests for random sound of the day fail (https://travis-ci.org/MTG/freesound/builds/275903832?utm_source=github_status&utm_medium=notification) it looks like a cache is set somehow. When I run them locally it works fine, so it probably has to do with the travis setup.

alastair:
Merged and released!

Issue #574, opened 2014-03-04
HTTPS rollout


alastair:
Some notes for things we need to fix before we can use https
- [x] Make captcha on registration page work (use `//` protocol-less address)
- [x] Make nginx take a round-robin server out of rotation if it's not available
- [x] Set http host on round robin
- [x] Do not show images from non-https sites

qubodup:
Once https is done, please also only use https://freesound.org and make https://www.freesound.org, http://freesound.org, http://www.freesound.org use https://freesound.org

alastair:
www -> nowww redirect is https://github.com/MTG/freesound/issues/664


The last thing to be done here is make linked images secure. This is extra work, probably requiring us to find images, download them, and cache them ourselves. Perhaps we should open a new ticket for it.

ffont:
@alastair @bdejong @andrebola  What do you think we should do with this ticket?
I think it would be nice to move completely to https (still allow http in the API though).
We should find a solution for the images problem.

bdejong:
This would be awesome especially if we could always forward ppl to HTTPS...

The image thing, ... Local caching? This could also speed up rendering
pages with external images and remove dependencies on external servers.

bdejong:
Alternative/easier: change the HTML parsing:

If an image URL does not start with https, turn the image link into a URL.
Even easier: disable image linking, but this might make some people angry.

alastair:
I like this option, we should just show images which are on an https link and otherwise show a link. Not sure how much of a change we need to make to our parser. I don't want to start the discussion about changing our comments and descriptions to Markdown, but perhaps we should put it on the list.... ;)

bdejong:
@alastair using markdown would not solve this particular problem though ;-)

More importantly: keep in mind that all HTML is valid markdown... So by switching to markdown you actually open up to even more chaos ;-)

qubodup:
Hi,

> disable image linking

Totally OK for comments, personal messages and descriptions. A bit uncool for forum posts.

> just show images which are on an https link and otherwise show a link

Sounds totally OK. Ideally paired with a notification to the poster "Freesound only supports embedded images from https:// domains. The image embed code was replaced with a link"

alastair:
I think we're agreed that images are an integral part of the freesound experience, so I don't think we want to remove them.

We'll work towards filtering out non-https images, with a warning message.

alastair:
[...]

PR #898, opened 2017-06-20
Replace unsecure images for links on forum

andrebola:
issue #574


forum/templatetags/filter_img.py
```diff
@@ -0,0 +1,17 @@
...
+from os.path import basename, splitext
+
+register = template.Library()
+
+
+@register.filter(is_safe=True)
+def replace_img(string):
+    soup = BeautifulSoup(string)
+    for img in soup.findAll('img'):
+        if 'https://' != img['src'][:8]:
```
bdejong: ```if not img['src'].lower().startswith("https")```

bdejong: also make sure this doesn't break where there is no "src" ;)

andrebola: If there is no src it doesn't break because in not a regular dict

forum/templatetags/filter_img.py
```diff
@@ -0,0 +1,17 @@
+from django import template
+from BeautifulSoup import BeautifulSoup, Tag
+from os.path import basename, splitext
+
+register = template.Library()
+
+
+@register.filter(is_safe=True)
+def replace_img(string):
+    soup = BeautifulSoup(string)
```
alastair: When I first thought of this, I considered just using string matching instead of parsing the html. Is this a viable solution? In hindsight I guess bs is a good solution for this, but we process html at submission time anyway, to make it only use a whitelist (https://github.com/MTG/freesound/blob/144a536168bf8763a2ab401d30e62775fc84eec4/utils/text.py#L209-L211).
Do we know what the processing time impact is of computing this each time a post is displayed?

Another thing to consider is that bs3 is no longer maintained. As far as I understand, bs3 and bs4 can be installed side-by-side. Should we consider slowly migrating new functionality to bs4?

forum/templatetags/filter_img.py
```diff
@@ -0,0 +1,17 @@
...
+    for img in soup.findAll('img'):
+        if 'https://' != img['src'][:8]:
+            a = Tag(soup, "a")
+            a['href'] = img['src']
+            a.string = "(Unsecure image)"
```
alastair: We should also add `rel="nofollow"`

forum/templatetags/filter_img.py
bdejong: bs is quite slow IIRC... :|

why don't we process the stuff on input and remove any HTTP(-not-s) links there?

ffont:
After discussion in Hipchat, things to do:

- Measure the speed of the templatetag with some example posts with images (busing BeautifulSoup)
- Add a check like `if "<img" in text and "http:" in text` to optimize speed
- The `(Unsecure image)` text is not needed, just put the URL and maybe we can surround the image in brackets.
- The templatetag should also be applied to sound comments (sound page), sound descriptions (sound page), pack descriptions (pack page), user about text (profile page)

I think that's it...

ffont:
Also regarding BeatifulSound3 and BeatifulSound4 maybe this is a god opportunity to start using 4 and compare the speed.

andrebola:
This is the comparison for different posts using Bs3 and using bs4 (time in seconds):


BS3 | bs4 | with image |
| ------------- |:-------------:|-------------:|
9.53674316406e-07 | 9.53674316406e-07 | no
1.90734863281e-06 | 9.53674316406e-07 | no
9.53674316406e-07 | 9.53674316406e-07 | no
9.53674316406e-07 | 9.53674316406e-07 | no
0.0 | 0.0 | no
9.53674316406e-07 | 9.53674316406e-07 | no
2.14576721191e-06 | 1.90734863281e-06 | no
3.09944152832e-06 | 2.14576721191e-06 | no
1.19209289551e-06 | 9.53674316406e-07 | no
9.53674316406e-07 | 9.53674316406e-07 | no
1.19209289551e-06 | 9.53674316406e-07 | no
0.000348091125488 | 0.000723838806152 | yes (http) 
1.19209289551e-06 | 9.53674316406e-07 | yes (https)

alastair:
I don't understand what the columns are. Can you explain what no/yes/http/https means?
Is each row a different comment? Do they have different text, or the same? Are they longer/shorter than each other?
Can you do a test with a number of different comment lengths, a differing number of imgs in each (maybe 1-3 or 1-5?)
Did you run each of the tests just once? Using `timeit` with 100 or 1000 iterations might help here to get a better idea of runtimes. Make sure that the timescale of these executions is the same (some are e-6 and some e-7 and some e0!)

andrebola:
Each row on the table corresponds to a different post of a forum.
It shows the time in seconds that took to execute the templatetag for the different posts. Each post has different length, some of them with secure image, some of them with unsecure images, and some without image tag.

The left column indicates the time using BeautifulSound3 and the third column indicates the time for BeautifulSound4, the last column indicates if the post has image or not and if this image is secure or not.

media/js/freesound.js
```diff
@@ -154,3 +154,19 @@ function afterDownloadModal(show_modal_url, sound_name){
             openModal();
         });
 }
+function unsecureImageCheck(input) {
+    var unique = '#unsecure_img_check' + new Date().valueOf();
+    var div = $("<div>", {style: "color: red"});
+    div.append("Sorry, images must be from a secure uri.");
```
ffont: Hmm, yeah ok that's the idea but the message could be more accurate, something like "Warning: you are adding an  image from a non-secure location. Only images from secure locations (i.e. the URL starts with https://...) will be shown in Freesound."

ffont:
Added the template tag and the image check in some places where it was missing. Also improved the design of the message. Now I think it is ready for deploy, but @alastair wanted to have a look at the warning message wording. Go ahead ;)

general/templatetags/filter_img.py
```diff
@@ -0,0 +1,19 @@
...
+
+register = template.Library()
+
+
+@register.filter(is_safe=True)
+def replace_img(string):
+    if not "<img" in string or not "http:" in string:
+        return string
+
+    soup = BeautifulSoup(string)
```
alastair: bs4 should have a parser argument: https://www.crummy.com/software/BeautifulSoup/bs4/doc/#installing-a-parser

general/templatetags/filter_img.py
```diff
@@ -0,0 +1,19 @@
...
+    for img in soup.findAll('img'):
```
alastair: bs4 uses `find_all`

general/templatetags/filter_img.py
```diff
@@ -0,0 +1,19 @@
...
+        if not img['src'].lower().startswith("https"):
+            a = soup.new_tag("a", href=img['src'])
+            a.string = img['src']
+            img.replaceWith(a)
```
alastair: replace_with

media/js/freesound.js
```diff
@@ -154,3 +154,25 @@ function afterDownloadModal(show_modal_url, sound_name){
...
+function unsecureImageCheck(input) {
+    var unique = '#unsecure_img_check' + new Date().valueOf();
+    var div = $("<div>", {class: "unsecure_image_warning"});
+    div.append("<b>Warning</b>: you are adding an image from a non-secure location. Only images from secure locations (i.e. the URL starts with https://...) will be shown in Freesound.");
+    div.insertBefore(input);
+    div.hide();
+
+    function show_message_if_insecure(){
+        var txt = input.val();
+        var regular_expression = new RegExp('.*<img.+src=.http:.*', 'i');
```
alastair: you don't need quotes around an attribute, so `` should work too. This means the regex could better be: `'.*<img.+src=.?http:.*'`

media/js/freesound.js
```diff
@@ -154,3 +154,25 @@ function afterDownloadModal(show_modal_url, sound_name){
             openModal();
         });
 }
+function unsecureImageCheck(input) {
+    var unique = '#unsecure_img_check' + new Date().valueOf();
```
alastair: what's this date? Is it needed?

Issue #902, opened 2017-06-21
Add logging of API access domain and http/s status


alastair:
We have configured our load balancer to redirect www.freesound.org to freesound.org, unless the URL is for `/apiv2` (https://github.com/MTG/freesound/issues/664).
We would love to redirect this for everything, but it causes issues with POST requests to the API. We want to see how many requests to the API are made to www and how many are made to no-www, grouped by api client. We can add these fields to the messages sent to graylog.
Once we have this data, we can work out the best approach to try and get all API clients to connect with no-www.

alastair:
The second part of this issue should be to make a decision about how to contact api users and get them to change.

We should try and contact all API client users to get them to change the access domain to freesound.org without www. However, if there is released software actively being used by users which accesses www, it's likely that we can never add this redirect in place (unless we decide that we want to break the site for x% of users)

ffont:
Reopening this issue as we only added logging for the protocol (http/https), but not for the domain (www/non-www).

PR #920, opened 2017-07-06
Add to api logger if the protocol used by the client is secure

andrebola:
issue #902


ffont:
To me it looks good, were you able to test it?

alastair:
We also wanted to add it the API was being accessed via freesound.org or www.freesound.org

ffont:
@alastair is right, I'd be good to add this too.

PR #921, opened 2017-07-07
After download modals ready to deploy

ffont:
In this PR we worked a bit more on the after download modals that ask users for donations. Most important changes include:

- Store modal settings (i.e. how often should be shown, etc.)  in the DB so we can edit them without deploying new code (we took care  of catching settings to avoid many DB hits)
- Update design of the modal to make it more attractive
- Add text in the donate page to explain "why should users donate"

Besides this changes, we also removed the donations campaign widget in the front page and simply left the donate button.


ffont:
For the code review, I'm specially interested in @alastair's opinion for the wording in the modal (`templates/sounds/after_download_modal_donation.html`) and in the donate page (`templates/donations/donate.html`).

ffont:
I deployed it to test and set settings to show the modal (almost) always.
However you need a user without uploaded sounds to see it. Here is a screenshot:

screen shot 2017-07-07 at 14 19 28

templates/donations/donate.html
```diff
@@ -5,20 +5,54 @@
...
-      <br>Please use the form below to make a donation. Note that donations are managed through the Phonos Foundation, an institution linked to the Music
-         Technology Group for the promotion of social and cultural initiatives. Click here to see the <a href="{% url 'donors' %}">full list of Freesound donors and and donations</a>.
-  </p>
-  <!-- donations -->
+    <p>
+        Thanks for supporting Freesound!
+    </p>
+    <p>
+        Please use the form below to make a donation. Note that donations are managed through the <b>Phonos Foundation</b>, an institution linked to the <b>Music
+        Technology Group</b> for the promotion of social and cultural initiatives. Click here to see the <a href="{% url 'donors' %}">full list of Freesound donors and and donations</a>.
```
alastair: I would remove "and donations"

templates/donations/donate.html
```diff
@@ -5,20 +5,54 @@
...
+    </p>
+    <p>
+        Are you asking yourself "<b>why should I donate to Freesound?</b>"
+        <a href="javascript:void(0);" onclick="show_donate_reasons();">Here are some reasons</a>, let's see if we can
+        convince you :)
+    </p>
+
+    <div id="why_should_donate" style="display:none;">
```
alastair: why hidden? Can't we just list the reasons?

templates/donations/donate.html
```diff
@@ -5,20 +5,54 @@
...
+            and machine) to maintain and keep improving Freesound.
+        </p>
+        <p>
+            We need your donations in order to be able dedicate more resources to Freesound and not only maintain it but also
+            make it better, add new features, improve existing ones, etc. Because of the nature of Freesound and the philosophy
+            of openness, sharing and reuse that we promote, we think that the sustainability model that best fits our case
+            is one based on user donations (like Wikipedia).
+        </p>
+        <p>
+            Considering the huge amount of users that <a href="http://blog.freesound.org/?p=666" target="_blank" rel="nofollow">
```
alastair: why nofollow? we control both of these sites so I don't think it's necessary

templates/donations/donate.html
alastair: This section has good information, but it's hidden in some quite large paragraphs. What about being a bit more upfront with the reasons?

### Why support freesound? 
* **We are growing and need more resources.** [something about how we started and how we are maintained]
* **We don't like ads.** [because of the ethos of freesound we don't like ads and would prefer to ask for donations from the people who find value from the site]
* **We don't need many donations.** [just a small donation from everyone who uses freesound would help us grow and improve]

templates/sounds/after_download_modal_donation.html
```diff
@@ -3,9 +3,15 @@
         Support Freesound
     </div>
     <div class="modal-body">
-        Thanks for downloading <b>{{ sound_name }}</b>!<br />
-        Please consider supporting us by
-        <a href="{% url 'donate' %}">making a donation</a> if you have not contributed yet.<br />
+        <p>Thanks for downloading <b>{{ sound_name }}</b>!</p>
+        <p>It looks like you have been donwloading a singificant number of sounds recently.
```
alastair: It looks like you have downloaded a significant number....

check spelling - 2 errors here

templates/sounds/after_download_modal_donation.html
```diff
@@ -3,9 +3,15 @@
...
+        If Freesound is being useful to you, please consider supporting us by
```
alastair: If freesound is useful to you,

templates/sounds/after_download_modal_donation.html
```diff
@@ -3,9 +3,15 @@
...
+        <a href="{% url 'donate' %}">making a donation</a>.
         Help us to keep Freesound open and free :)
+        </p>
+    </div>
+
+    <div style="display: flex; justify-content: center;">
+        <a class="donate-button" href="{% url 'donate' %}">Donate Now :)</a>
```
alastair: Some smiles are OK, but I wouldn't put 2 in subsequent paragraphs

donations/management/commands/donations_mails.py
```diff
@@ -39,59 +39,90 @@ def handle(self, **options):
...
-                                }, None, user.email)
-                        user.profile.last_donation_email_sent = datetime.datetime.now()
-                        user.profile.save()
-                        logger.info("Donation request sent to user %i" % user.id)
+                        'donations/email_donation_request.txt', {
+                            'user': user,
+                            }, None, user.email)
+                    user.profile.last_donation_email_sent = datetime.datetime.now()
+                    user.profile.save()
+                    logger.info("Donation request sent to user %i" % user.id)
```
ffont: The text of the log message should be similar to the aboce.
Could be "Sent donations email (max downloads) to user:#"
Then the one above could be changed to: "Sent donations email(reminder) to...".
In this way the reason is included in the message.

Issue #709, opened 2016-02-17
Don't use javascript for search filtering


alastair:
On the search page we use javascript to add tag/license/other filters to the search query, however all this does is add some text to the search bar and submit the form again. We could replace this by making all links go to the search page with the query field updated. It should be easy to make a templatetag that performs any necessary text filtering.

PR #925, opened 2017-07-07
Not use js for applying filters on search

andrebola:
issue #709


alastair:
This is a good start for the specific task listed in the issue, but there is much much more javascript in this file that does other stuff that we could do in the template: https://github.com/MTG/freesound/blob/e2d95ad87353e325256311aaa612ca82ca65b3bc/templates/search/search.html#L41-L270
I think we should change this code as well.

andrebola:
Now the only Js on the search page refers to the advanced search which is a separate form

Issue #912, opened 2017-07-05
Store sound license in download object


ffont:
We should add a column to the downloads table to store the license with which sounds are downloaded.

Related discussion: https://github.com/MTG/freesound/issues/840

Issue #840, opened 2017-04-04
Store downloads when a user downloads a sound or pack more than once


alastair:
We currently have a unique index on (user, pack_id) or (user, sound_id) in the downloads table. This means that we only store one record of a user downloading something.
We've decided that we want to store records of all downloads. To do this we need to:

* Remove the unique indexes on the download table
* Create a new Download record each time a user downloads a sound or pack
* Check that all uses of the `Download` class treat these duplicates in a reasonable way
  * Users who downloaded a pack or sound
  * Attribution list
  * Sounds downloaded by user

We may also want to consider how we show the sound attribution page when a user downloads a sound twice, but the uploader has changed the license in the meantime. We might want to store the license in the Downloads table (perhaps uses too much data), or in a history table so that we can find the correct license that the user downloaded the sound under

bdejong:
Perhaps it would make sense to break the pack downloads into individual
sound downloads (I.e. a single pack download creates multiple sound
download rows) and have a bool "as pack" in the list.
Currently we don't keep track of the changes in packs. If a sound changes
from one pack to another this could create weirdness if you'd analyse the
pack downloads.
That said this probably need some analysis to see how badly the table would
explode in size.

ffont:
Currently licenses can not be changed, however I think it is good to start thinking about it and I think its a good idea to add a "license" field in the download table. It's just a foreign key to the licenses table, don't think is a problem in terms of data. Another option is that if we allow license change at some point and store history of licenses for a sound, then we can guess by timestamps the license for a particular download. But still storing the license with the download entry will make future queries faster and easier (I guess).

Regarding pack downloads I agree with Bram, that could be a good solution. I'm also concerned about the size though. We have an open issue to discuss options for moving downloads outside of database (https://github.com/MTG/freesound/issues/585). I guess this is all related. The average number of sounds per pack is 15 (just computed it now), and pack downloads represent 2.5% of total number of downloads. Therefore, my estimation is that the size of the table would be increased by 35%*.

* (num_total_pack_downloads*15 + num_total_sound_downloads) / num_total_downloads

ffont:
For this issue we should simply remove the constraint which prevents us from storing repeated downloads of the same object-user pair.

Out of the discussion happening in this issue, we opened others:
- https://github.com/MTG/freesound/issues/912 (store licenses in downloads table)
- https://github.com/MTG/freesound/issues/914 (store sound downloads for packs separately)

ffont:
[...]

Issue #608, opened 2014-04-09
License change list


ffont:
We should store changes in sound licenses with timestamps.

ffont:
We should have a deeper look into this as it could become a major license-handling refactor.
Also, if we want to update to CC4.0 licences it would be nice to implement that before the migration.

Issue #914, opened 2017-07-05
Store which sounds are downloaded when downloading packs


ffont:
Currently when a pack is downloaded we simply add a row to the downloads table which indicates which pack has been downloaded. We should consider storing also information about which sounds were downloaded inside the pack. One option is to create a row in the downloads table for each sound (see discussion related in https://github.com/MTG/freesound/issues/840), but maybe is not the best one (it would require also changes in the attribution views, etc...). The other option is to add a new (JSON?) field to the download model and store list of sounds there. We can also add a new `PackDownload` model that stores this is a new table. Other options are welcome :)

ffont:
After discussion in Hipchat and in person we agreed the following:

- We will create a `PackDownload` table in which we will store information about pack downloads including all the sounds downloaded for each pack.
- It is not yet clear whether we should store individual sounds as rows in the `PackDownload` (with the table having columns like `user_id,sound_id,license_id,pack_id,date`), or if we should store each pack download in a single row and then store the list of downloaded sounds and their licenses in a JSON field of that row (then columns would look like `user_id,pack_id,sounds_data,date`). As part of this issue, an experiment should be done comparing the two cases. Lets create both tables and evaluate their size and the efficiency of queries like "get all pack_ids downloaded by a user", "get sounds and licenses of a pack at the time user x downloaded it".
- Whatever strategy we choose, we'll have to write a command to copy pack downloads data from the current `Download` table to `PackDownload` table. This command might be slow and might take days to complete so we have to take that into account and think how are we going to avoid. If we don't do it right, it can happen that `PackDownload` ends up having rows whose date is not consecutive with the row id (like we are inserting information from new downloads as people download stuff and from old downloads as the command runs). We want to avoid for Postgres performance reasons. One possible solution is to have a flag in settings to disable storing new downloads to the `PackDownload` table, and have it set to false while we run the command to copy data. Then when the data is up to date we enable the flag.
- For this issue we are not making any changes to `Download` table. We will continue storing pack downloads into the current `Download` table, just as we are doing now, and not change the way we display lists of pack/sound downloads and attribution. This will happen in another issue after this one to "refactor downloads table".
- The last step of this issue should be to create a new one "Refactor Download table" which explains what to do next (rename the table, remove packs info from the table, changes in fronted, etc.)

ffont:
Updating this issue after progress in #1043:

[...]

PR #928, opened 2017-07-12
License refactor and changes in downloads

andrebola:
1) Added model to store License history for a sound, also changed license attribute on Sound for last_license.

2) Removed from the SQL files the unique constraint for the Download table (this has to be done manually on the database). Also changed the code to support multiple creation of Downloads for the same user and sound.

3) Added License field to Download class, and store the current license when downloading

issue #608 #840 #912


sounds/migrations/0008_download_license.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('sounds', '0007_auto_20161018_1809'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='download',
+            name='license',
+            field=models.ForeignKey(blank=True, default=None, null=True, on_delete=django.db.models.deletion.CASCADE, to='sounds.License'),
```
andrebola: We discussed this with @ffont, we thought that would be better to distinguish which of them had an historical from those that we didn't.

ffont: after **live** conversation we agreed that we should populate the license field for all downloads. However, because such population could take a lot of time, we propose to proceed as follows:

1. Keep the migration as it is now (creating the column but not populating it)
2. Modify the code in the attribution page to avoid breaking if license info is null (just show the license of the sound)
3. Deploy these changes and then make a script which populates the downloads table. This can do it bit by bit, and it's not a problem if it takes even days to complete.
4. (when the script has finished). Update the code in the attribution page to show the license of the download object.

We should add a note in the migration which explains that the script must be run.

alastair: Ideally the license column should be `NOT NULL`, but while we have packs and sounds in the same table this may not be possible. This will have to be done as part of #914

accounts/views.py
```diff
@@ -155,6 +155,8 @@ def bulk_license_change(request):
         if form.is_valid():
             selected_license = form.cleaned_data['license']
             Sound.objects.filter(user=request.user).update(license=selected_license, is_index_dirty=True)
+            for sound in Sound.objects.filter(user=request.user).all():
+                SoundLicenseHistory.objects.create(sound=sound, license=selected_license)
```
andrebola: added atomic transaction decorator

sounds/views.py
```diff
@@ -291,7 +291,7 @@ def sound_download(request, username, sound_id):
     sound = get_object_or_404(Sound, id=sound_id, moderation_state="OK", processing_state="OK")
     if sound.user.username.lower() != username.lower():
         raise Http404
-    Download.objects.get_or_create(user=request.user, sound=sound)
+    Download.objects.create(user=request.user, sound=sound, license=sound.license)
```
andrebola: added atomic transaction decorator

sounds/models.py
```diff
@@ -890,3 +893,10 @@ class RemixGroup(models.Model):
                                     related_name='remix_group',
                                     blank=True)
     group_size = models.PositiveIntegerField(null=False, default=0)
+
+
+class SoundLicenseHistory(models.Model):
+    """Store history of licenses related to a sound"""
+    license = models.ForeignKey(License)
+    sound = models.ForeignKey(Sound)
+    created = models.DateTimeField(db_index=True, auto_now_add=True)
```
andrebola: done on 8655c98cae0dff639a7ac7f63645c544dad01fe5

ffont:
After talk with @andrebola , steps for deployment should be:

- Merge PR
- Before deploy, remove constraints `sounds_download_user_sound` and `sounds_download_user_pack` from downloads table
- After deploy, run `update_downloads_license` regularly in a cron job until all license values have been updated

sounds/migrations/0008_download_license.py
```diff
@@ -0,0 +1,21 @@
...
+        ),
```
alastair: We should add the following sql commands to this migration:

```
create index if not exists sounds_download_user_sound on sounds_download(user_id, sound_id) WHERE pack_id IS NULL;
create index if not exists sounds_download_user_pack on sounds_download(user_id, pack) WHERE sound_id IS NULL;
drop index if exists sounds_download_user_pack_unique;
drop index if exists sounds_download_user_sound_unique;
```

ffont:
@andrebola alastair and I will make the necessary fixes and then deploy.

ffont:
Closing this PR as it's contents were replaced by https://github.com/MTG/freesound/pull/964

Issue #876, opened 2017-06-06
Can't reset password with email address


alastair:
Most of our account recovery pages let you enter a username or email address, but reset password only lets you enter a username. Expand to allow an email address too.

andrebola:
I don't understand since the form validates the emails, like this:
https://github.com/MTG/freesound/blob/878a435a1e0a7921e6dd523c3dc09192ce8fed6a/accounts/tests.py#L849:L850

And the UI is also restricted to emails:

<img width="760" alt="captura de pantalla 2017-06-26 a la s 12 34 59" src="https://user-images.githubusercontent.com/991305/27547318-f77c5942-5a6b-11e7-9a59-64c743dd658e.png">

Issue #875, opened 2017-06-06
Don't notify that email addresses exist


alastair:
We have a number of forms which let someone enter an email address to perform a task:

* re-send auth code
* Find username
* change email
* password reset

In these forms, we should make sure that we don't show an error message if an account does or doesn't exist. Just show a generic message saying "If you have an account you will get an email". This prevents people from being able to use these forms to see if an account with an email address exists in Freesound.

PR #930, opened 2017-07-14
Reset password form accept username and email

andrebola:
issue #876


alastair:
Can you rebase this so that it doesn't have the javascript commits?

andrebola:
This also fixes part of #875 regarding reset password form

accounts/forms.py
```diff
@@ -345,8 +352,31 @@ class EmailSettingsForm(forms.Form):
     )
 
 
-class FsPasswordResetForm(PasswordResetForm):
-    def get_users(self, email):
+class FsPasswordResetForm(forms.Form):
```
ffont: Isn't there a way that we could reuse more from Django's `PasswordResetForm`?
For example, the `send_email` and `save` methods look the same after a quick inspection.
Why do we need to create the form "from scratch"?

ffont: After talking with @alastair he clarified that you tried to reuse as much as possible but had to reimplement it anyway. I'll add a comment to the code and an entry to the [Migration to a new version of Django](https://github.com/MTG/freesound/wiki/Migration-to-a-new-version-of-Django) wiki doc to check for changes in this code when upgrading.

PR #932, opened 2017-07-17
log if user has accessed with www or without it

andrebola:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

I'm not 100% if this header is currently set by haproxy when does the redirect, in the case that haproxy is not setting it I think we should do it.

issue #902


alastair:
why not use `META['HTTP_HOST']`? - https://docs.djangoproject.com/en/1.11/ref/request-response/#django.http.HttpRequest.META

Issue #842, opened 2017-04-11
ConnectionError: HTTPSConnectionPool(host='www.paypal.com', port=443): Max retries exceeded with url: /cgi-bin/web...


alastair:
A failure during paypal stuff is probably bad - we want to make sure that this happens as little as possible.

You can do retries in requests: http://stackoverflow.com/questions/15431044/can-i-set-max-retries-for-requests-request
Is this operation something that has to be done in the user workflow (e.g., we have to show something to the user afterwards?) or can we do it in a worker process and retry until it succeeds?

https://logserver.mtg.upf.edu/sentry/freesound-web/issues/624/

```
ConnectionError: HTTPSConnectionPool(host='www.paypal.com', port=443): Max retries exceeded with url: /cgi-bin/webscr (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x6b91650>: Failed to establish a new connection: [Errno -2] Name or service not known',))
(4 additional frame(s) were not displayed)
...
  File "donations/views.py", line 25, in donation_complete
    req = requests.post(settings.PAYPAL_VALIDATION_URL, data=params)

ConnectionError: HTTPSConnectionPool(host='www.paypal.com', port=443): Max retries exceeded with url: /cgi-bin/webscr (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x6b91650>: Failed to establish a new connection: [Errno -2] Name or service not known',))
```

andrebola:
When someone makes a donation we receive a notification from paypal, then we verify that the information we received with paypal's server ( in this last step occurs this error) and if the verification is fine we save the donation and notify the user.

The notification of the donation from paypal is sent many times for the same donation, in this case the donations has been validated and saved one minute later:
https://www.freesound.org/donations/donors/

I think for some reason paypal didn't respond to our request, this could be fixed by setting a number of retries but I'm not sure if this is required since paypal sends the same notification many times.

alastair:
OK, to clarify: This URL is accessed by paypal, and then we get the body and send a confirmation to them to say that we have it. If we don't confirm, they see it again?

So this error won't be seen by an end user.

What is paypal's policy on retrying? Will they retry if we send anything other than HTTP 200? Only retry if we send 5xx? 

I think it would be good that we do at least the following
 * Use a session in requests to set `max_retries` for the POST,
[...]

PR #937, opened 2017-07-19
Change exception when paypal does not respond

andrebola:
issue #842

Since this error happened again I checked that the donation was
successfully stored on the DB, so this means the problem is that paypal
sometimes doesn't respond the requests.


alastair:
The specific 'name or service not known' is due to UPF DNS being bad, and not paypal.
The change to catch `ConnectionError` is fine, however we could also consider using requests sessions to retry 2 or 3 times: https://stackoverflow.com/a/15431343

```python
from requests.adapters import HTTPAdapter
s = requests.Session()
s.mount(settings.paypal_url, HTTPAdapter(max_retries=5))
s.post(settings.paypal_url)
```

andrebola:
@alastair is it really necessary? Since Paypal invoques multiple times the endpoint with the same information.

ffont:
It is Paypal ho sends a request to us right? so I'm not sure I understand well you suggestion @alastair

alastair:
yes, we should be programming defensively. If we can prevent a python exception from being raised by adding a retry loop then we should. We already know that this specific error (cannot resolve hostname) is caused by the UPF infrastructure and that retrying causes it to succeed.

andrebola:
Ok, I made the change and tested in test.freesound.org

Issue #631, opened 2014-09-15
id field in solr schema is not sortable


ffont:
Currently the id field in solr schema is not sortable, which means that numeric range filters over sound ids in apiv1 and apiv2 do not work properly. We should either index the field as a sortable integer ('sint' instead of 'integer') or document that numeric range filter can not be applied to sound ids. In any case, it does not make much sense to filter by ranges in ids...

andrebola:
Since this requires a change on solr schema we will do it together with #593

Issue #593, opened 2014-03-12
Use solr for selecting a random sound


alastair:
If we add a `random` field to the solr schema then we can use its random ordering to select our random sounds. This should be significantly faster than the current 500ms that it takes to query the database.
Something like this: http://www.params.me/2011/04/solr-random-results.html

alastair:
There are some really good comments here about doing this in postgres using some approximations: http://stackoverflow.com/questions/8674718/best-way-to-select-random-rows-postgresql

andrebola:
This is very related with #839

andrebola:
Now we create the random sounds querying the database in a cron job, so I think we don't need to use solr.

andrebola:
As we already talked, if we want to get a random sound faster we could user solr or improve the way we do it with postgres.

Here I did a comparison between the different ways to get the number of sounds:
```
freesound=# SELECT reltuples AS ct FROM pg_class WHERE oid = 'sounds_sound'::regclass;
   ct   
--------
 347636
(1 row)

Time: 0.265 ms
freesound=# SELECT count(*) from sounds_sound;
 count  
--------
 351626
(1 row)

Time: 101.880 ms
```

I also tried to get a random sound with solr, here is the result, look at Qtime (in milliseconds):

```
{
  "responseHeader": {
    "status": 0,
    "QTime": 132,
    "params": {
      "fl": "id",
      "sort": "headword_random_2158 asc",
      "indent": "true",
      "q": "*:*",
      "_": "1500449463017",
      "wt": "json",
      "rows": "1"
    }
  },
  "response": {
    "numFound": 347129,
    "start": 0,
    "docs": [
      {
        "id": 43423
      }
    ]
  }
}
```

PR #938, opened 2017-07-20
Solr changes to get random sound

andrebola:
This PR is linked with the issues #631 and #593

Regarding changes on Solr to allow range queries on the id field for the API, also added some other attributes to solr to get a list of random sounds.


utils/search/search_general.py
```diff
@@ -39,6 +41,7 @@ def convert_to_solr_document(sound):
     document["description"] = remove_control_chars(sound.description)
     document["tag"] = list(sound.tags.select_related("tag").values_list('tag__name', flat=True))
     document["license"] = sound.license.name
+    document["is_explicit"] = sound.is_explicit
```
alastair: does this have to be added to the schema?

andrebola: yes!

sounds/views.py
```diff
@@ -125,10 +127,18 @@ def remixed(request):
...
 
 def random(request):
-    sound_id = Sound.objects.random()
-    if sound_id is None:
+    sounds = get_random_sounds_from_solr(1)
+    sound_obj = None
+    if sounds:
+        try:
+            sound_obj = Sound.objects.get(id=sounds[0]['id'])
+        except ObjectDoesNotExist:
```
alastair: `Sound.DoesNotExist`

sounds/views.py
```diff
@@ -125,10 +127,18 @@ def remixed(request):
 
 
 def random(request):
-    sound_id = Sound.objects.random()
-    if sound_id is None:
+    sounds = get_random_sounds_from_solr(1)
```
alastair: Will this method be used in any other places? I wonder if we don't need to get multiple random sounds - what if this method only returns 1 sound?

sounds/views.py
```diff
@@ -125,10 +127,18 @@ def remixed(request):
...
+    sound_obj = None
+    if sounds:
+        try:
+            sound_obj = Sound.objects.get(id=sounds[0]['id'])
+        except ObjectDoesNotExist:
+            pass
+    if sound_obj is None:
+        sound_id = Sound.objects.random()
+        sound_obj = Sound.objects.get(pk=sound_id)
+    if sound_obj is None:
```
alastair: The line above is going to raise `Sound.DoesNotExist` if it doesn't work, not return `None`

sounds/views.py
```diff
@@ -125,10 +127,18 @@ def remixed(request):
...
-    if sound_id is None:
+    sounds = get_random_sounds_from_solr(1)
+    sound_obj = None
+    if sounds:
+        try:
+            sound_obj = Sound.objects.get(id=sounds[0]['id'])
+        except ObjectDoesNotExist:
+            pass
+    if sound_obj is None:
+        sound_id = Sound.objects.random()
```
alastair: This method has been updated to return an object, not an id: https://github.com/MTG/freesound/blob/b66b51d7c5294eca737fcc44f3c55aad1e786c33/sounds/models.py#L103

It'd be nice to have a test for this view which we know would catch this change.

andrebola:
After this changes the sound accessed after going to http://freesound.org/browse/random should be retrieved using solr and not the DB. There should not be any change on the behavior of the random sound shown in the frontpage

alastair:
This looks better :+1:
After talking with @ffont we decided to release this schema on appserver3, then we can reindex and switch over without downtime, then I will upgrade search to debian jessie and we can move the index back. I'll take care of the release.

Issue #818, opened 2017-03-17
IntegrityError: duplicate key value violates unique constraint "ratings_rating_user_id_key"


alastair:
Appears similar to #742, we should make sure we do this work in a transaction to guard against people double-clicking

This code could be updated to use `get_or_create` too

https://logserver.mtg.upf.edu/sentry/freesound-web/issues/319/

```
IntegrityError: duplicate key value violates unique constraint "ratings_rating_user_id_key"
DETAIL:  Key (user_id, content_type_id, object_id)=(329858, 20, 125520) already exists.

(11 additional frame(s) were not displayed)
...
  File "ratings/views.py", line 40, in add
    rating_object = Rating.objects.create(user=request.user, object_id=object_id, content_type=content_type, rating=rating)

IntegrityError: duplicate key value violates unique constraint "ratings_rating_user_id_key"
DETAIL:  Key (user_id, content_type_id, object_id)=(329858, 20, 125520) already exists.

```

PR #939, opened 2017-07-21
Changed rating view to use transaction also changed query to use get_or_create

andrebola:
issue #818

### DIFF
--- a/ratings/tests.py
+++ b/ratings/tests.py
@@ -54,10 +54,9 @@ class RatingsTestCase(TestCase):
         """ Change your existing rating. """
         loggedin = self.client.login(username="testuser1", password="testpass")
         self.assertTrue(loggedin)
-        r = ratings.models.Rating.objects.create(object_id=self.sound.id, content_type=self.ct, user_id=self.user1.id, rating=4)
 
         resp = self.client.get("/ratings/add/%s/%s/%s/" % (self.ct.id, self.sound.id, 5))
-        newr = ratings.models.Rating.objects.get(id=r.id)
+        newr = ratings.models.Rating.objects.first()
         self.assertEqual(ratings.models.Rating.objects.count(), 1)
         # Ratings in the database are 2x the value from the web call
         self.assertEqual(newr.rating, 10)
--- a/ratings/tests.py
+++ b/ratings/tests.py
@@ -55,6 +55,8 @@ class RatingsTestCase(TestCase):
         loggedin = self.client.login(username="testuser1", password="testpass")
         self.assertTrue(loggedin)
 
+        r = ratings.models.Rating.objects.create(object_id=self.sound.id, content_type=self.ct, user_id=self.user1.id, rating=4)
+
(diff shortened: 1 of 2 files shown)


ratings/views.py
```diff
@@ -24,20 +24,20 @@
...
     if rating in range(1,6):
         # in order to keep the ratings compatible with freesound 1, we multiply by two...
         rating = rating*2
-        try:
-            rating_object = Rating.objects.get(user=request.user, object_id=object_id, content_type=content_type)
-            rating_object.rating = rating;
-            rating_object.save()
-        except Rating.DoesNotExist: #@UndefinedVariable
-            rating_object = Rating.objects.create(user=request.user, object_id=object_id, content_type=content_type, rating=rating)
+        rating_object = Rating.objects.get_or_create(
```
alastair: `get_or_create` returns `object, created`. If you're not going to use the variable, don't set it

ratings/views.py
```diff
@@ -24,20 +24,20 @@
...
+                user=request.user,
+                object_id=object_id,
+                content_type=content_type, defaults={'rating': rating})
```
alastair: This method should an existing rating if it exists. This change won't do that - `defaults` is only used if the object doesn't exist and it's created. You'll need to set `rating` if the object already exists

ratings/tests.py
```diff
@@ -54,10 +54,9 @@ def test_rating_change(self):
         """ Change your existing rating. """
         loggedin = self.client.login(username="testuser1", password="testpass")
         self.assertTrue(loggedin)
-        r = ratings.models.Rating.objects.create(object_id=self.sound.id, content_type=self.ct, user_id=self.user1.id, rating=4)
```
alastair: This test is supposed to check that a rating for a user is updated if the rating already exists and then the user triggers the rating URL. If you delete this line then you are only checking that a rating is added, not updated

andrebola:
After this changes the sound accessed after going to http://freesound.org/browse/random should be retrieved using solr and not the DB. There should not be any change on the behavior of the random sound shown in the frontpage

ffont:
Looks good to me now. @alastair will you approve the PR or should I directly merge it?

ratings/tests.py
```diff
@@ -54,10 +54,11 @@ def test_rating_change(self):
         """ Change your existing rating. """
         loggedin = self.client.login(username="testuser1", password="testpass")
         self.assertTrue(loggedin)
+
         r = ratings.models.Rating.objects.create(object_id=self.sound.id, content_type=self.ct, user_id=self.user1.id, rating=4)
 
         resp = self.client.get("/ratings/add/%s/%s/%s/" % (self.ct.id, self.sound.id, 5))
-        newr = ratings.models.Rating.objects.get(id=r.id)
+        newr = ratings.models.Rating.objects.first()
```
alastair: this is OK because in the next line we check that we only have 1 `Rating` object, but I preferred the original line because we explicitly check that the ratings object is the same based on its ID.

alastair:
good! 👍

PR #940, opened 2017-07-24
Force paypal to display english interface

andrebola:
We got some messages of users telling that the form on paypal was in
spanish, this is because of the country on the seller's account.


donations/views.py
```diff
@@ -102,7 +102,8 @@ def donate(request):
                         "business": settings.PAYPAL_EMAIL,
                         "item_name": "Freesound donation",
                         "custom": returned_data_str,
-                        "notify_url": return_url
+                        "notify_url": return_url,
+                        "lc": 'en_US'
```
alastair: two types of quotes on the same line? :scream:

Issue #922, opened 2017-07-07
use display_raw_sound html template in display_sound templatetag


alastair:
We have two templatetags for displaying sounds: https://github.com/MTG/freesound/blob/209d4bdc33f484720432f02e2b6f9b5be89e3dbe/sounds/templatetags/display_sound.py
`display_sound` takes a `Sound` object, while `display_raw_sound` takes the result of `Sound.objects.bulk_query`. Each templatetag has its own html template file.
These template tags show the same thing, but having two templates is dangerous because we could forget to make a change in both of them. We should make a change so that the `display_sound` templatetag uses the `display_raw_sound.html` template, creating the necessary context.

Additional work, see if the `display_sound` tag is actually needed. How much work is it to move all instances to `display_raw_sound`?

PR #941, opened 2017-07-24
Unify templatetags to display sound from index or database

andrebola:
issue #922


sounds/templatetags/display_sound.py
```diff
@@ -52,6 +52,8 @@ def display_sound(context, sound):
      'sound_id':     sound_id,
      'sound':        sound_obj,
      'sound_tags':   sound_tags,
+     'sound_user':   sound.user.username,
+     'license_name': sound.license.name,
```
alastair: can we call select_related on user and license so that we don't have to make extra queries here?

Issue #812, opened 2017-03-06
IntegrityError: duplicate key value violates unique constraint "tags_taggeditem_tag_id_key"


alastair:
https://logserver.mtg.upf.edu/sentry/freesound-web/issues/578/

I wonder if this is because the user double-clicks on the submit button? We could do the check/add all in a single transaction, or otherwise use `get_or_create` instead of `create` when making the tag items.

```
IntegrityError: duplicate key value violates unique constraint "tags_taggeditem_tag_id_key"
DETAIL:  Key (tag_id, content_type_id, object_id)=(25232, 20, 383023) already exists.

(12 additional frame(s) were not displayed)
...
  File "sounds/views.py", line 340, in sound_edit
    sound.set_tags(data["tags"])
  File "sounds/models.py", line 436, in set_tags
    tagged_object = TaggedItem.objects.create(user=self.user, tag=tag_object, content_object=self)

IntegrityError: duplicate key value violates unique constraint "tags_taggeditem_tag_id_key"
DETAIL:  Key (tag_id, content_type_id, object_id)=(25232, 20, 383023) already exists.

```

PR #942, opened 2017-07-25
Add atomic transation to sound edit view

andrebola:
issue #812


ffont:
I could locally check that the error is caused by double-clicking the submit button and that adding the `@transaction.atomic()` decorator fixes the problem. Not clear in the explanation of the issue or the PR if it was tested that this was the problem. I'm wandering, can something like that be tested in a unit test? Also, the "double-click" problem affects any form right? What about the API views for editing content (like sound edit) shouldn't we always use `@transaction.atomic()` for views that write to DB? Am I being paranoid now?

ffont:
We've been talking with Alastair about the potential problems of not using `@transaction.atomic()`. It is a real problem and already "fixed" it in a number of places by using `@transaction.atomic()`. At some point we should review if there are other places where we should add it, but in any case it is out of the scope of this PR. I opened an issue to further deal with these problems https://github.com/MTG/freesound/issues/967.

PR #943, opened 2017-07-25
Don't notify that email addresses exist

andrebola:
issue #875

Note:
When a user changes the email for another which is already on the database we don't
inform the user that the email is already used


accounts/forms.py
```diff
@@ -287,23 +268,6 @@ class EmailResetForm(forms.Form):
...
-    def clean_email(self):
-        email = self.cleaned_data["email"]
-        try:
-            get_user_by_email(email)
-            raise forms.ValidationError(_("A user using that email address already exists."))
-        except User.DoesNotExist:
-            pass
-        return email
-
-    def clean_password(self):
```
ffont: Why is this clean function also removed?

accounts/views.py
```diff
@@ -222,8 +222,14 @@ def resend_activation(request):
     if request.method == 'POST':
         form = ReactivationForm(request.POST)
         if form.is_valid():
-            user = form.cleaned_data['user']
-            send_activation(user)
+            username_or_email = form.cleaned_data['user']
```
ffont: So now activation works with usernames and emails right? This is a new feature not explained in the PR or issue description right?

accounts/views.py
```diff
@@ -935,38 +946,45 @@ def old_user_link_redirect(request):
 @login_required
 def email_reset(request):
     if request.method == "POST":
-        form = EmailResetForm(request.POST, user=request.user)
```
ffont: This changes depend on whether `EmailResetForm` keeps the `clean_password` method (see my comment above).

accounts/forms.py
andrebola: I think this was removed by mistake, I will add it again

accounts/views.py
andrebola: The clean_user method of the form used to check email and username, now this is made in the view.

bdejong:
This is a relatively big change on user/email validation so I think this could really use a test so we know we're not breaking anything here.

ffont:
I improved the tests and fixed a problem that was spotted after improving the tests. I'll merge the PR as now I'm quite positive it's ready to deploy.

Issue #926, opened 2017-07-11
IntegrityError: insert or update on table "tickets_ticket" violates foreign key constraint "tickets_ticket_sound_...


ffont:
https://logserver.mtg.upf.edu/sentry/freesound-web/issues/689/

```
IntegrityError: insert or update on table "tickets_ticket" violates foreign key constraint "tickets_ticket_sound_id_97fa5496_fk_sounds_sound_id"
DETAIL:  Key (sound_id)=(396700) is not present in table "sounds_sound".

(11 additional frame(s) were not displayed)
...
  File "tickets/views.py", line 435, in moderation_assigned
    tc.save()
  File "tickets/models.py", line 138, in create_ticket_message
    instance.ticket.save()

IntegrityError: insert or update on table "tickets_ticket" violates foreign key constraint "tickets_ticket_sound_id_97fa5496_fk_sounds_sound_id"
DETAIL:  Key (sound_id)=(396700) is not present in table "sounds_sound".

```

andrebola:
We think this might be related with a migration that took some time, only look it if this error happens again since today

andrebola:
This error happened more times last month, I think the problem is that when the moderator deletes a sound  it gets deleted before the ticket is updated.

ffont:
This seems to make sense, did you check?

andrebola:
All the events on sentry that I checked were related with delete actions of the moderators, in my local I couldn't reproduce the error though.

PR #947, opened 2017-09-04
Update loaded reference of sounds for the tickets

andrebola:
After reproducing the error in my environment I fix the problem by
removing the reference to the sounds just after deleting the sounds.

issue #926

### DIFF
--- a/tickets/views.py
+++ b/tickets/views.py
@@ -401,7 +401,6 @@ def moderation_assigned(request, user_id):
                     if ticket.sound.pack:
                         packs_to_update.add(ticket.sound.pack)
                 Sound.objects.filter(ticket__in=tickets).delete()
-                tickets.update(sound_id=None)
                 notification = Ticket.NOTIFICATION_DELETED
 
             elif action == "Whitelist":
@@ -419,6 +418,7 @@ def moderation_assigned(request, user_id):
                     tickets""" % ", ".join(users))
 
             for ticket in tickets:
+                ticket.refresh_from_db()
                 if action != "Delete":
                     # We only fill here users_to_update and packs_to_update if action is not
                     # "Delete". See comment in "Delete" action case some lines above
--- a/tickets/views.py
+++ b/tickets/views.py
@@ -401,6 +401,10 @@ def moderation_assigned(request, user_id):
                     if ticket.sound.pack:
                         packs_to_update.add(ticket.sound.pack)
                 Sound.objects.filter(ticket__in=tickets).delete()
+                # After we delete sounds that these tickets are associated with,
+                # we refresh the ticket list so that sound_id is null and this does
+                # not affect the TicketComment post_save trigger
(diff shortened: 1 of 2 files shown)


tickets/views.py
```diff
@@ -400,6 +401,7 @@ def moderation_assigned(request, user_id):
                     if ticket.sound.pack:
                         packs_to_update.add(ticket.sound.pack)
                 Sound.objects.filter(ticket__in=tickets).delete()
+                tickets.update(sound_id=None)
```
ffont: `Ticket` model has a foreign key to `Sound` [with `on_delete=models.SET_NULL`](https://github.com/MTG/freesound/blob/master/tickets/models.py#L58), therefore the manual update of the field should not really be necessary right? Did you understand why this is happening? Isn't the `transaction.atomic()` bit already fixing the problem?

andrebola: Note that the error was in line https://github.com/MTG/freesound/pull/947/files#diff-f0c47805ccfcbaead49a83d1ce646e68R437. So I think the problem was that the local objects either were not refreshed with the data in the database or they were modified after.
Now, by calling update on the queryset the values are updated immediately for all the Ticket.

andrebola: I just checked that only using `transaction.atomic()` doesn't fix the problem

ffont: I see, makes sense :)

ffont: Then we should be using maybe something like `obj.refresh_from_db()` but for a queryset right? (if it exists)

ffont: In any case, the solution seems ok but a comment should be left there explaining why we need to explicitly call `tickets.update(sound_id=None)`

alastair: After talking this through with Frederic I finally understand the issue - I didn't realise that it happened in the post_save trigger for `TicketComment`. My preference here would be to use `Ticket.objects.filter` again instead of the `update`. Perhaps with a comment like

    # After we delete sounds that these tickets are associated with, we refresh the ticket list so that sound_id is null and this does not affect the TicketComment post_save trigger

I think that the comment is important because we both looked at the code and intuitively thought that the update wasn't needed because of the `on_delete` setting. Having an explicit explanation is a good idea

andrebola: I changed it for refresh_from_db as @ffont suggested. If you think is still not clear I can change it

alastair:
looks OK now

ffont:
Regarding this commit https://github.com/MTG/freesound/pull/947/commits/3c19e173b87e901add0b7a1137facddcbb611dcb which disabled button after submitting, it would be nice to think of an strategy that could be easily applied to all the places where it could be helpful. I kind of like the way you do in JS on document ready. Maybe we could have a function in the general freesound JS on document ready which finds submit buttons with a specific class and then finds their parent form and "overwrite" the submit function just as you do. Then we should simply add a class like `disable-on-submit` to all the buttons that should implement the same behaviour. What do you think?

Issue #816, opened 2017-03-17
OSError: [Errno 2] No such file or directory: '/home/fsweb/freesound/freesound-data/uploads/5937039/Simple...


alastair:
This may have been related to our swapping of storage locations. In any case, we should probably wrap this in a try/catch.

https://logserver.mtg.upf.edu/sentry/freesound-web/issues/290/

```
OSError: [Errno 2] No such file or directory: '/home/fsweb/freesound/freesound-data/uploads/5937039/Simple Gunshot.wav'
  File "accounts/views.py", line 500, in describe_sounds
    os.remove(forms[i]['sound'].full_path)

OSError: [Errno 2] No such file or directory: '/home/fsweb/freesound/freesound-data/uploads/5937039/Simple Gunshot.wav'
```

ffont:
Make sure in our Sentry logs that this error always occurs at `os.remove(...)`. If that is the case, for this issue we simply need to handle the exception and pass (or log).

alastair:
Also happening in another part of freesound

https://logserver.mtg.upf.edu/sentry/freesound-web/issues/685/

This code is the same, we should break it out into a helper and call it from both places (and also check if we remove a sound like this in other places).

alastair:
@ffont pointed out that the second error might be related to double-click on the describe page. We should check if this seems to be the case.
If we want to start disabling buttons to fix double-click then we should open a ticket for it and make some documentation to explain how we can easily apply this to many pages without duplicating code.

alastair:
https://logserver.mtg.upf.edu/sentry/freesound-web/issues/705/

PR #948, opened 2017-09-05
Add try/catch to os.remove when file not exists

andrebola:
issue #816


accounts/views.py
```diff
@@ -429,8 +430,14 @@ def describe(request):
...
             elif "delete_confirm" in request.POST:
                 for f in form.cleaned_data["files"]:
-                    os.remove(files[f].full_path)
-                    remove_uploaded_file_from_mirror_locations(files[f].full_path)
+                    try:
+                        os.remove(files[f].full_path)
+                        remove_uploaded_file_from_mirror_locations(files[f].full_path)
+                    except OSError as e:
+                        if e.errno == errno.ENOENT:
+                            logger.error("Failed removing file " + str(e) + " (no such file or directory)")
```
alastair: This is really picky, but...
- with loggers we should use query parameters - `logger.error("foo %s", param)`
- `str(e)` is already going to include "no such file or directory" so we probably don't need the additional comment in brackets
- Grammar - "Failed to remove file"

ffont: Fixed https://github.com/MTG/freesound/pull/948/commits/bbd0f63e19780cb8ec64b65fdec1f1ea8846ec2e :)

Issue #616, opened 2014-04-22
Support m4a file format


bdejong:
Should be relatively easy to add (ffmpeg does it)...

ffont:
When this is updated, we should update all the infos which say which formats are allowed, etc.

PR #951, opened 2017-09-11
Support m4a

andrebola:
Issue https://github.com/MTG/freesound/issues/616

### DIFF
--- a/_docs/api/source/resources.rst
+++ b/_docs/api/source/resources.rst
@@ -60,7 +60,7 @@ was_remixed: 		boolean
 pack: 			string
 pack_tokenized: 	string, tokenized
 is_geotagged: 		boolean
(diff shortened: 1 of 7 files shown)


andrebola:
I tested this on test.freesound.org:

http://test.freesound.org/people/testandre/sounds/274380/

When deploying on production we need to install the following dependency for converting m4a to wav:
`sudo apt-get install faad`

Initially I was using ffmpeg  for the conversion but on some versions of debian ffmpeg is harder to compile and install, so I think is safer to use faad.

ffont:
I installed `faad` in processing 1 and processing 2. Also added the dependency to the install scripts in freesound deploy repository.

The conversion falls back to `ffmpeg` if `faad` fails  (that's how we do it for other formats, their specific encoder falls back to `ffmpeg` if problems occur, `ffmpeg` is already installed in our servers).  We should consider getting rid of all specific encoders and use only `ffmpeg`.

PR #953, opened 2017-09-18
Catch error on Comment delete when sound is already deleted

andrebola:
Fixes: https://logserver.mtg.upf.edu/sentry/freesound-web/issues/795/

Also added test for this case.


comments/models.py
```diff
@@ -41,5 +42,8 @@ class Meta:
 
 
 def on_delete_comment(sender, instance, **kwargs):
+    try:
         instance.sound.post_delete_comment()
+    except ObjectDoesNotExist:
```
alastair: `Sound.DoesNotExist`
It would be nice to have a clarifying comment here, like we do in a few other post_delete triggers. Maybe something like
"If this comment is deleted as a result of its parent sound being deleted, the sound will no longer exist so we don't need to update it"

sounds/tests.py
```diff
@@ -239,6 +240,7 @@ def test_sound_delete(self):
         sound = sounds[0]
         sound.change_processing_state("OK")
         sound.change_moderation_state("OK")
+        c = Comment.objects.create(user=user, sound=sound)
```
alastair: I guess we want to check somewhere that the comment no longer exists too. What about using `Sound.add_comment` to add this comment?

comments/models.py
andrebola: Comment added

ffont:
Now this PR looks good to me, do you want to have a second look @alastair or should I just merge it?

PR #955, opened 2017-09-21
Notify a user if their sound has been chosen as sound of the day

alastair:
Fixes missing functionality from #839


alastair:
tests will pass when merged into master, due to #956

PR #959, opened 2017-09-25
Added redirect_authenticated_user parameter to LoginVIew

andrebola:


alastair:
This commit should have a description explaining what it does and why it's needed - as it is, I don't understand why we need to make the change and what it fixes.

bdejong:
If you click this link: http://www.freesound.org/apiv2/apply
You end up at: http://freesound.org/home/ instead of http://freesound.org/apiv2/apply

ffont:
I agree there is not enough information here. In fact I'm not sure I fully understand the problem. I'm trying to reproduce it in production and I can't do it anymore. Tried being logged in/logged out and both www/non-www, http/https (note that this page is under the API domain so our www/https redirects do not apply) and I had no problems...

Can anyone further reproduce the problem?

andrebola:
The problem is that we are currently ignoring the next parameter, for example: http://freesound.org/home/login/?next=/monitor/stats/

By adding the parameter redirect_authenticated_user to LoginView the 'next' parameter is used, by default is ignored.

bdejong:
@ffont I can reproduce it 100% of the time :-) If you want me to make a movie, let me know

bdejong:
Just read #961 - we should probably reject this one...

alastair:
We fixed #961 with a different method

Issue #793, opened 2017-01-31
UnicodeDecodeError: 'utf8' codec can't decode byte 0xe9 in position 21: invalid continuation byte


ffont:
https://logserver.mtg.upf.edu/sentry/freesound-web/issues/313/

```
UnicodeDecodeError: 'utf8' codec can't decode byte 0xe9 in position 21: invalid continuation byte
(17 additional frame(s) were not displayed)
...
  File "rest_framework/request.py", line 193, in user
    self._authenticate()
  File "rest_framework/request.py", line 316, in _authenticate
    user_auth_tuple = authenticator.authenticate(self)
  File "apiv2/authentication.py", line 48, in authenticate
    super_response = super(OAuth2Authentication, self).authenticate(request)
  File "oauth2_provider/ext/rest_framework/authentication.py", line 18, in authenticate
    valid, r = oauthlib_core.verify_request(request, scopes=[])
  File "oauth2_provider/oauth2_backends.py", line 167, in verify_request
    valid, r = self.server.verify_request(uri, http_method, body, headers, scopes=scopes)
```

ffont:
Apparently you can reproduce the problem sending a query like this:
```
http://xxxx/apiv2/search/text/?query=ambient&filter=tag:(rain%20OR%CAfe)
```
I'm not sure if this covers "all of it". We should probably raise an issue in the rest framework repo and see what they think. Still we should first make sure that we're not doing any processing to this data which might be causing this. I don't think so but I remember we overwritten some methods in rest framework classes.

alastair:
For data that we control (e.g. query parameters like @ffont's example) we should validate it before we use it. However, I think this error is caused by rest framework + oauthlib trying to verify the data before we get access to it in the view. So we need to work out if this is a bug in oauthlib that should be reported upstream, or if we need to wrap our entire auth function in a try/catch for unicode errors.

Also: See what happens on python3

andrebola:
See: https://github.com/evonove/django-oauth-toolkit/issues/202

It looks like we should update oauthlib to 2.0

andrebola:
After updating django-rest-framework to 3.6 and oauthlib to 2.0 the error still occurs, so I think the related [issue](https://github.com/evonove/django-oauth-toolkit/issues/202) says it has been fixed for oauthlib 2.0 using python 3.

alastair:
It looks like this is happening with some other parameters too - 

https://logserver.mtg.upf.edu/sentry/freesound-web/issues/974/

[...]

PR #966, opened 2017-09-28
Catch string decode error on api request

andrebola:
issue #793

### DIFF
--- a/apiv2/authentication.py
+++ b/apiv2/authentication.py
@@ -39,7 +39,7 @@ class OAuth2Authentication(Oauth2ProviderOauth2Authentication):
     @property
     def authentication_method_name(self):
         return "OAuth2"
-    
+
     def authenticate(self, request):
         """
         We override this method to check the status of related ApiV2Client.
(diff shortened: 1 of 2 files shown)


ffont:
Can we add a unit test for this?

apiv2/authentication.py
```diff
@@ -45,7 +45,12 @@ def authenticate(self, request):
         We override this method to check the status of related ApiV2Client.
         Check that ApiV2Client associatied to the given acess_token has not been disabled.
         """
-        super_response = super(OAuth2Authentication, self).authenticate(request)
+        try:
+            super_response = super(OAuth2Authentication, self).authenticate(request)
+        except UnicodeDecodeError:
+            # Catch excpetion when data can not be decoded
```
alastair: one small thing here - it's great that we have a comment! However it would be nice to have this a little more explicit as to the reason, e.g.
```py
# If the request contains html entities that don't decode to valid UTF8, an exception is raised during oauth validation, even if it's on a field/parameter unrelated to oauth (#793)
# TODO: Check if this is still needed with Python3/oauthlib2
```

Issue #913, opened 2017-07-05
Use django signal instead of DB trigger to increment Sound.num_downloads


ffont:
Currently we use a database trigger to increment `Sound.num_downloads` when a sound is downloaded.
We should use a Django signal instead. Same thing applies to packs.
In this signal, sounds should also be marked as index dirty so they get reindexed.

andrebola:
I will make this change after #840 is merged with master

Issue #464, opened 2012-11-09
Redesign functionalities currently in triggers


ghost:
Triggers cause problems when deleting objects because they often tr to modify the object being deleted

ffont:
Consider reimplementing triggers as django signals

ffont:
Update of num_sounds in user profiles is now carried out in django and not as a trigger.
Current db triggers include downloads count, forum post/thread/last_post and sound ratings.

ffont:
sound.num_comments, user.num_sounds and pack.num_sounds are already handled by django now.
Remaining triggers are for ratings (and computing average rating) and for forum stuff.

ffont:
Related to this one (specifically for the case of downloads): https://github.com/MTG/freesound/issues/913

andrebola:
The following triggers can be deleted after merging the branch [remove-db-triggers](https://github.com/MTG/freesound/tree/remove-db-triggers):

forum_thread_delete
forum_post_delete
forum_post_insert
forum_thread_insert
forum_thread_update

download_post_insert
download_post_delete

rating_post_delete
rating_post_insert
rating_post_update

Issue #674, opened 2015-06-25
Optimise forum latest posts


ffont:
The query to get forum latest posts is quite slow. Could be probably optimised using custom SQL.

andrebola:
Forum has a reference to the last_post, but the method get_last_post doesn't use it:
https://github.com/MTG/freesound/blob/c8ecaf2c36d3db4784c0242f2e0051a48d451dcf/forum/models.py#L67-L68

@ffont Is this what you refer? Or there is a specific part that it's slow?

ffont:
For what I see in the code you're right that `get_last_post` could simply use the stored one. I guess this method was implemented before `last_post` was stored in the model. Maybe this method could be refactored so that if `last_post` is null, then `Thread.set_last_post()` is called before returning `self.last_post`. In this way we'll avoid extra queries.

alastair:
I'm not sure I like the solution to this in 38f6336
* Because we do the potential lookup in a function, we can't add `last_post` to the forum with `select_related`. This means that we're always doing 1 extra db lookup for each forum that we have
* All of our forums already have a `last_post` set. What's the case when it might be null? When we have a new forum - in this case we don't have a post anyway, so `set_last_post` won't have any affect anyway
* In the template we check against `if forum.last_post.id` but then use `get_last_post` when printing the username and time of the post. This could be simplified to just use `last_post`
* `set_last_post` has some commented code which isn't used, and some badly formatted text layout. Since we're working on this code we could clean it up

I think we should remove `get_last_post` and make sure that when we select forums we do `select_related`. Can we override the forum manager to always do this for us?

alastair:
Looking at the title of this issue again I wonder if it's actually talking about http://freesound.org/forum/latest_posts/ (which is already pretty fast, I don't think there's any further optimisation we need to do here)

PR #968, opened 2017-09-29
Remove db triggers

andrebola:
issues #464 #913 

This PR replaces the following triggers:  
```
forum_thread_delete
forum_post_delete
forum_post_insert
forum_thread_insert
forum_thread_update

download_post_insert
download_post_delete

rating_post_delete
rating_post_insert
rating_post_update
```


alastair:
I think it would make sense to have a migration to remove triggers here - `DROP TRIGGER IF EXISTS`

andrebola:
Also added change requested on #674

sounds/models.py
```diff
@@ -1013,6 +1014,29 @@ class Meta:
         ordering = ("-created",)
 
 
+@receiver(post_delete, sender=Download)
+def update_num_downloads_on_delete(**kwargs):
+    download = kwargs['instance']
+    if download.pack:
```
alastair: Does django give us the instance object with any related objects loaded? In these two methods we're going to be doing 2 select methods to resolve `pack` and `sound` which will double the number of queries needed for adding a download object (insert, select, select, update) instead of the current insert, update from the trigger.

Can we replace the checks with `if download.pack_id`?

If this can't be done, let's at least put sound first, since it's the most common download object and use `elif` instead of two `if`s

ratings/models.py
```diff
@@ -42,3 +46,29 @@ def __unicode__(self):
...
+
+
+@receiver(post_delete, sender=Rating)
+def post_delete_rating(sender, instance, **kwargs):
+    try:
+        instance.content_object.num_ratings = F('num_ratings') - 1
+        rating = Rating.objects.filter(
+                content_type_id=instance.content_type_id,
+                object_id=instance.object_id).aggregate(Avg('rating')).values()[0]
+        if rating == None:
```
alastair: `is`

ratings/models.py
```diff
@@ -42,3 +46,29 @@ def __unicode__(self):
...
+            rating = 0
+        instance.content_object.avg_rating = rating
+        instance.content_object.save()
+    except ObjectDoesNotExist:
+        pass
+
+
+@receiver(post_save, sender=Rating)
+def update_num_ratings_on_post_insert(**kwargs):
+    instance = kwargs['instance']
```
alastair: There are some things that we should change here, but with #979 I don't think it's worth changing them twice. We should merge both patches before we do a release @ffont

forum/models.py
```diff
@@ -112,10 +115,31 @@ def __unicode__(self):
...
+            old_thread.forum.save()
+            instance.forum.num_threads = F('num_threads') + 1
+            instance.forum.save()
+
+
 @receiver(post_delete, sender=Thread)
 def update_last_post_on_thread_delete(**kwargs):
     thread = kwargs['instance']
     try:
+        thread.forum.refresh_from_db()
```
alastair: Transaction for this block.
There's a little bit of hidden magic here, where `set_last_post` performs the `.save()` that applies the update to `num_threads`. 
You have a similar comment at https://github.com/MTG/freesound/pull/968/files#diff-c1b18b9ac76c85da21bb14f4e0481a14R181, Do you think that we can improve this? Possible options:
 * @ffont has `commit=True` default parameters in some methods (https://github.com/MTG/freesound/blob/936b5d233fde0be045f48e40ee10793f166ae27e/sounds/models.py#L651) which allows us to force or ignore this and call it again separately
 * rename the method to not be `set_` so that it makes it clear that this sets and saves the method?
 * Documentation in `set_last_post` to say that it saves the object and that this behaviour is used elsewhere in the code?
 * Remove the save and require it to be added after any call to the method

alastair:
Ready!

PR #972, opened 2017-10-04
If the user is logged in the email is prefilled for stripe popup

andrebola:


templates/donations/donate.html
```diff
@@ -139,9 +139,11 @@
 
         });
         $('.donation_button_stripe').click(function(e) {
+          var email = "{{user.email}}";
```
ffont: Shouldn't this be `{{request.user.email}}`?

PR #973, opened 2017-10-06
Display specific error to user for stripe

andrebola:
If the error is related with the credit card, now we display the message
to the user.

### DIFF
--- a/donations/views.py
+++ b/donations/views.py
@@ -119,22 +119,26 @@ def donation_complete_paypal(request):
     This view listens to a notification made from paypal when someone makes
     a donation, it validates the data and then stores the donation.
     """
-    params = request.POST.copy()
-    params.update({'cmd': '_notify-validate'})
 
-    try:
-        req = requests.post(settings.PAYPAL_VALIDATION_URL, data=params)
-    except requests.exceptions.Timeout:
-        logger.error("Can't verify donations information with paypal")
-        return HttpResponse("FAIL")
+    if "mc_gross" in params:
+        # Paypal makes notifications of different events e.g: new suscriptions,
+        # we only want to save when the actual payment happends
+        params = request.POST.copy()
+        params.update({'cmd': '_notify-validate'})
 
-    if req.text == 'VERIFIED':
-        _save_donation(params['custom'],
-            params['payer_email'],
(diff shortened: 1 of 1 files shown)


andrebola:
I also added in this PR the fix for the issue: https://logserver.mtg.upf.edu/sentry/freesound-web/issues/806/
We only have to discard this kind of notifications and store when the actual payment is received.

donations/views.py
```diff
@@ -111,22 +119,26 @@ def donation_complete_paypal(request):
...
-        _save_donation(params['custom'],
-            params['payer_email'],
-            params['mc_gross'],
-            params['mc_currency'],
-            params['txn_id'],
-            'p')
+
+    if "mc_gross" in params:
+        # Paypal makes notifications of different events e.g: new suscriptions,
+        # we only want to save when the actual payment happends
```
ffont: Ok.
We should consider maybe storing information about these other notifications in the future.
It would be nice to know how many subscribers we have, etc. Maybe there is an API endpoint we can call to get this information. Same would apply for stripe, although we currently don't support subscriptions via stripe.

donations/views.py
```diff
@@ -111,22 +119,26 @@ def donation_complete_paypal(request):
     This view listens to a notification made from paypal when someone makes
     a donation, it validates the data and then stores the donation.
     """
-    params = request.POST.copy()
-    params.update({'cmd': '_notify-validate'})
```
ffont: Removing this line makes the tests fail, for some reason I did not see the tests were failing before merging the PR, but I saw before after running the tests locally. Fixed this in https://github.com/MTG/freesound/commit/6271a334995953e2e8f97b97539a4ba9a0a864c0

donations/views.py
```diff
@@ -99,9 +100,16 @@ def donation_complete_stripe(request):
                     donation_received = True
                     messages.add_message(request, messages.INFO, 'Thanks! your donation has been processed.')
                     _save_donation(form.encoded_data, email, amount, 'eur', charge['id'], 's')
+            except stripe.error.CardError as e:
+                # Since it's a decline, stripe.error.CardError will be caught
```
alastair: # If it's a decline

PR #977, opened 2017-10-10
Add string format parameter for log message

alastair:
Related to changes made in #973, but this change was missed


alastair:
Fixes https://logserver.mtg.upf.edu/sentry/freesound-web/issues/805/

PR #979, opened 2017-10-13
Remove generic relation in  rating with sound

andrebola:
This changes should be merged after https://github.com/MTG/freesound/pull/968


alastair:
This PR is difficult to understand, as it also has the commits from #968. Can you rebase against that branch again and change the base branch of this PR so that it only shows the 5 main commits of this PR?

ratings/models.py
```diff
@@ -21,24 +21,18 @@
 #
 
 from django.contrib.auth.models import User
-from django.contrib.contenttypes import fields
-from django.contrib.contenttypes.models import ContentType
 from django.db import models
 
 class Rating(models.Model):
```
alastair: This is now a `SoundRating`

alastair:
I merged master into this branch and fixed conflicts. It looks good to me now.

alastair:
* Renamed model name
* Alter the migration files so that we have a single migration that removes the generic relation and renames column names to match the new field names. This is much faster, but will break if you applied the original migrations in this PR previously
* Make some changes to the URL used to rate sounds
* Cleaned up the way that we show the vote stars on a sound to remove code duplication

@ffont a quick review on these points would be great, thanks.

PR #981, opened 2017-10-13
Remove old solr 3.3 core

alastair:


alastair:
I see that you deleted config for 1.4, but left 1.1! I'll delete it without a PR

alastair:
sorry, my mistake, I didn't see that you removed both old cores!

Issue #965, opened 2017-09-28
Web-based API browser doesn't work in www.freesound.org


alastair:
Continuing from #961

If a user visits freesound.org/apiv2 in a web browser they get an interactive documentation/api browser tool.
They must be logged in for this, but if they visit with www.freesound.org then they won't have a session cookie, and get an error.
We want to redirect webbrowser users from www. to nowww. We can do this in [`GenericAPIView.initial`](https://github.com/MTG/freesound/blob/91c1d206d1ec5b5b633c180eef1f18e54a393910/apiv2/apiv2_utils.py#L71) using the value of `contains_www` and if the renderer is set to `BrowsableAPIRenderer`. If so, we should issue a redirect to nowww.

ffont:
After deploying to production we manually tested this functionality and it is not working when users are not logged in. @andrebola do you remember checking if `finalize_response` is called even if users are not logged in? I'll remove the "merged" label as this issue needs more work.

andrebola:
@ffont the problem is that when the code of finalize_response was moved to a new method then the returned value is lost. 

So, this line should be something like `response = self.redirect_to_nowww_if_needed(request, response)`

https://github.com/MTG/freesound/pull/982/files#diff-7c0184a36b103db6a6f6af55a7e067d2R137

PR #982, opened 2017-10-13
Redirect to non-www on API browser

andrebola:
issue #965

The method GenericAPIView.initial raise an exception if the user is not logged before setting the value of self.accepted_renderer, so it can't be evaluated if the user is accessing through the browser.

Therefore I'm overriding the method finalize_response, where the value of self.accepted_renderer indicates if it's using the browser.


ffont:
This looks good to me, but because we don't have the redirect from `www` to `non-www`enabled now we can't test. Let's wait until we have it enabled again and then we can merge and test.

apiv2/apiv2_utils.py
```diff
@@ -79,6 +81,22 @@ def initial(self, request, *args, **kwargs):
...
+        """ This method is overriden to make a redirect when the user is using the interactive API browser and
+        with 'www' sub-domain. The problem is that we can't check if it's accessing through the interactive browser
+        inside the 'initial' method because it raises an exception when the user is not logged in, that exception is
+        handled by 'finalize_response' method of APIView.
+        """
+        response = super(GenericAPIView, self).finalize_response(request, response, *args, **kwargs)
+
+        # If the user is using the interactive API browser and the www in the domain we redirect to nowww.
+        host = request.get_host()
+        if type(response.accepted_renderer) == type(BrowsableAPIRenderer()) and 'www' in host:
```
alastair: `host.startswith('www.')`

apiv2/apiv2_utils.py
```diff
@@ -79,6 +81,22 @@ def initial(self, request, *args, **kwargs):
...
+            domain = "https://%s" % Site.objects.get_current().domain
```
alastair: let's use the scheme from `request` here - if it's http then we can rely on haproxy to redirect us again to https

by using the request scheme we can test locally more easily

apiv2/apiv2_utils.py
alastair: I prefer to use `isinstance` instead of `type`, because it works for subclasses too

ffont:
I'm working on this PR, I'll make changes as @alastair requests and some other small ones. @andrebola I'll take it from here.

Issue #789, opened 2017-01-23
Add 'delete post' option on forum moderation


ffont:
Currently when moderation forum posts there is only the option of approving a post or deleting the user (and the post). This works most of the time because posts which are not approved are mostly spam. But sometimes there might be some inappropriate posts that should not be approve but there is no need to delete the user. In that case it would be good to have a 3rd option to remove the post/thread only.

bdejong:
Do we have concrete examples of this? Otherwise I would say this is not
really needed...

ffont:
Hmm I remember finding myself in that situation sometimes and wishing I had this, that's why I finally opened the issue. Now I couldn't put a specific example but I will if it happens again.

PR #983, opened 2017-10-16
Add option on forum moderation to delete the post

andrebola:
Also added feedback to the user when the post has been deleted

issue #789

### DIFF
--- a/forum/views.py
+++ b/forum/views.py
@@ -18,24 +18,23 @@
 #     See AUTHORS file.
 #
 
+import re
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required, permission_required
-from django.core.paginator import Paginator, InvalidPage
-from django.urls import reverse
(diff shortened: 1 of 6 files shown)


forum/views.py
```diff
@@ -335,8 +336,12 @@ def moderate_posts(request):
             elif action == "Delete User":
                 try:
                     post.author.delete()
+                    messages.add_message(request, messages.INFO, 'The user has been successfully deleted.')
                 except: #someone deleted him already
                     pass
+            elif action == "Delete Post":
+                messages.add_message(request, messages.INFO, 'The post has been successfully deleted.')
```
ffont: I'll re-order these two lines not to add the message if deleting post breaks for some reason :)

ffont:
I saw that the code in forums views was a bit ugly so I cleaned it up a bit and also added basic tests for forums app. Even though tests could be more complete, this should ease further refactoring in the future.

Issue #915, opened 2017-07-05
Prepare Freesound development setup for implementing new front end


ffont:
We want to be able to start working on the new front end (which will be implemented from scratch) and be able to switch from the new front end to the other with a GET parameter (or some variable stored in DB on a user basis or something). The goal of this is that we can start updating the front end page by page. Just to provide some more context, we want to have the backend compatible with the two versions of the front end. We also thought of taking this opportunity to try using a different templating system like jinja. Some things to try/think/make decisions:

- How to specify which frontend to use (GET param?, session param?). I guess we basically need a middleware which sets a variable which is accessible then from the views.
- How make views choose one frontend or another? Maybe wrapper around the `render` function? In some cases it might happen that the view needs to redirect to some other view for the new frontend. 
- It can also happen that one frontend requires variables which another one does not. How to handle these cases? I guess if we have this middleware we can do if-elses for the backend type. Nevertheless the goal is to avoid that as much as possible, and have the two frontends share the same backend.
- The new frontend will most probably use some customised version of libraries like bootstrap or semantic-ui and we'll need to do some css pre-compiling. We should include this in the workflow somehow. We should see which is the best way to do integrate css/js/whatever pre-compilers with Django.

andrebola:
Django supports multiple template engines, and it can be specified which to use by a name. This name is specified in the settings: https://github.com/MTG/freesound/blob/f22640188eb80cabc31d4b8a8cacfa51e569f5c0/freesound/settings.py#L334-L355

By defining our own render method we pass the name of the template engine to use:
https://github.com/MTG/freesound/blob/c7c98b552487a54148315d5a1e7757305bf9f54a/freesound/render.py#L1-L10

By doing it in this way we can decide later if we want to use django template engine or jinja2 and the only thing to change is the setting: 
https://github.com/MTG/freesound/blob/f22640188eb80cabc31d4b8a8cacfa51e569f5c0/freesound/settings.py#L337

Also note currently I created the folder templates2 for the new interface but this could be changed to use a template folder inside each app.

bdejong:
@andrebola switching the templates might not be enough though: we also need to be able to switch the view functions... I guess you could do some crazy stuff with middleware and try to dynamically switch from one view to another.. I.e. define

```python
def my_view_function(...):
     return bleep

def my_view_function_newdesign(...):
    return bloop
```

and dynamically choose if you want to use my_view_function or my_view_function_newdesign?

bdejong:
[...]

PR #986, opened 2017-10-20
Development setup for new frontend

ffont:
Related issue: https://github.com/MTG/freesound/issues/915

This PR provides a development setup for the new Freesound frontend. We want to be able to start working on the new front end (which will be implemented from scratch) and be able to switch from the new front end to the other at any time. The goal of this is that we can start updating the front end page by page. Just to provide some more context, we want to have the backend compatible with the two versions of the front end. 

This PR includes:

- A way for Django to decide which templates to load. A **session parameter** is used to store which frontend to be used. This session parameter can be set at any time by adding a request parameter like `?fend=XX` to any request. Depending on the frontend to be used, Django will decide to look up for template files in `templates` or in `templates2` folder (this is configured in `settings.TEMPLATES` *). When configured to look up templates in `templates2` (the new interface), if a template is not found, then Django will fall back to the old frontend. In this way we can incrementally add more pages to the new frontend and still use the site. Because of the way we have implemented this, Django **will never mix templates** from the two frontends in a single request (e.g. it won't take `base.html` from `templates2` and `sound.html` from `templates`). To activate the new interface, use `?fend=bw`. To activate again the old interface use `?fend=ng` (or type any string other than `bw`).

- A number of util functions in `utils.frontend_handling` to facilitate server-side changes during the development of the new frontend. The util functions include decorators for redirecting views (`redirect_if_beastwhoosh` and `http_redirect_if_beastwhoosh`) and functions to easily check which frontend is being used (`redirect_if_beastwhoosh` and `using_nightingale`). All these functions are documented in the code so check `utils.frontend_handling` for more information.

- A `webpack` setup to deal with new static files. This is under the `static` folder. Because in our current version of Freesound we are not using the static files app and folders (we have static content in `media`), we can directly use `static` for all the things related to the new interface. We include a JS server to serve live-generated static files during development. See `static/README.md` for instructions on how to work with static files (i.e. how to run the server, produce bundles, etc.). 

In this proposed development setup the generated static files with webpack are not added to the repository (are ignored). My idea is that our deployment script will generate this files at deployment time (in the local machine) and then place them in the remote servers (i.e. using Fabric's `put` command). But I'm not 100% sure this is the best way to go. Please share your thoughts :)

Also, the current setup means that the JS server needs to be running all the time for doing development with freesound (unless `USE_JS_DEVELOPMENT_SERVER` is set to `False`). Maybe we could think of some hook in Django's `runserver` which starts the JS server if it is not running or something like this.

BTW, the new version of Freesound is called *Beast Whoosh*, you can listen to it here: https://freesound.org/people/martian/sounds/403973/

* We can override `settings.TEMPLATES` in production local settings to completely disable the new frontend while we still don't want to make it publicly accessible through the request parameter.


freesound/middleware.py
```diff
@@ -77,31 +63,45 @@ def __call__(self, request):
...
+class FrontendPreferenceHandler(object):
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        """
+        This middleware sets a session variable when the parameter 'new_frontend' is received.
+        The 'render' method will use this session variable to display the new/old frontend
+        """
+        if request.GET.get(settings.FRONTEND_CHOOSER_REQ_PARAM_NAME, None):
```
bdejong: shouldn't we at least cast this?

freesound/middleware.py
```diff
@@ -77,31 +63,45 @@ def __call__(self, request):
...
 
         if request.user.is_authenticated \
-            and not 'tosacceptance' in request.get_full_path() \
-            and not 'logout' in request.get_full_path() \
-            and not 'tos_api' in request.get_full_path() \
-            and not 'tos_web' in request.get_full_path() \
-            and not 'contact' in request.get_full_path() \
-            and not 'bulklicensechange' in request.get_full_path() \
-            and not request.get_full_path().startswith(settings.MEDIA_URL):
+                and 'tosacceptance' not in request.get_full_path() \
```
bdejong: I don't like mixing features and refactoring in the same PR, but whatever :)

freesound/settings.py
```diff
@@ -334,14 +339,65 @@
...
+# settings were loaded). We add this variable to the context processor and use it in base.html as a
+# parameter for the url of all.css and freesound.js files, so me make sure client browsers update these
+# files when we do a deploy (the url changes)
+LAST_RESTART_DATE = datetime.datetime.now().strftime("%d%m")
+
+# Followers notifications
+MAX_EMAILS_PER_COMMAND_RUN = 1000
+NOTIFICATION_TIMEDELTA_PERIOD = datetime.timedelta(days=7)
+
+PASSWORD_HASHERS = [
```
bdejong: Why are there things in here merged from the main branch. Shouldn't we be rebasing so these things don't show up in the PRs?

static/.babelrc
```diff
@@ -0,0 +1,8 @@
+{
```
bdejong: This probably shouldn't be in here

static/.eslintignore
```diff
@@ -0,0 +1,3 @@
+src/vendors/
```
bdejong: this probably shouldn't be in here

static/.eslintrc
```diff
@@ -0,0 +1,11 @@
+{
```
bdejong: this probably shouldn't be in here

utils/frontend_handling.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def using_nightingale(request):
+    """
+    Returns True if currently used frontend is Nightingale
+    """
+    return using_frontend(request, settings.FRONTEND_NIGHTINGALE)
+
+
+def redirect_if_beastwhoosh(redirect_to_view):
```
bdejong: I think "redirect" is a badly chosen name, what with HttpRedirect existing...

How about "defer" ? Maybe Alastair can think of a better name. Also the variable name "redirect_to_view" sounds alot like you are really HTTP redirecting to another view...

utils/frontend_handling.py
```diff
@@ -0,0 +1,159 @@
...
+        def _wrapped_view(request, *args, **kwargs):
+            if not using_beastwhoosh(request):
+                return view_func(request, *args, **kwargs)
+            else:
+                return redirect_to_view(request, *args, **kwargs)
+        return _wrapped_view
+    return decorator
+
+
+def http_redirect_if_beastwhoosh(redirect_url_name, kwarg_keys=None, query_string=''):
```
bdejong: Ok, this makes it even worse 😄

freesound/middleware.py
ffont: Why? I mean if the param is not correctly passed we just show the default. This only needs to be passed once so I think we're fine.

freesound/middleware.py
ffont: I know, but we also have to be practical and improve things.

freesound/settings.py
ffont: I think because I moved things arround in settings and then merged master back to the branch.

utils/frontend_handling.py
ffont: Then this one can be simply `redirect_redirect_if_beastwhoosh`

freesound/middleware.py
bdejong: Hmm, considering that, perhaps it's complete overkill to do this in Middleware? Why not make a small view function that is `def switch_to_beastwoosh(request)` rather than the thing with parameter in any page?

Issue #731, opened 2016-04-13
describe sounds: automatically expand accordions of sounds with errors


alastair:
So that we don't have to find the red ones and expand them

Issue #729, opened 2016-04-13
describe sounds: on describe page cursor should turn into pointer when hovering over sound in list


alastair:
On the describe page there are instructions " Click on the sound filenames to display the form for describing each sound." but we should make it look more like the filenames are links.

Some ideas to make this more understandable:
- have arrows > which show that it can be expanded
- The cursor should change to a link pointer when we mouse over it.
- use an `<a>` tag so we get underline and text darkening when we hover (need to work out the colours for the "error" case)

andrebola:
I think this is already done, should we close it?

PR #987, opened 2017-10-23
Expand accordion on sound upload when it has errors

andrebola:
issue #731 

Also #729: Visual changes on sound describe page to sound headers. Note that the following icons are used, we need to credit the author: https://www.flaticon.com/free-icon/keyboard-right-arrow-button_60758


ffont:
I tested it locally and when there are errors the accordion is expanded but then there are no arrows. I guess this is a bug. 

Regarding the arrows I'm not sure where should we put the attribution. I guess we add a "credits" paragraph in the about section. Nevertheless, aren't there any other arrow signs that we could use? This is just going to be provisional until we release the new UI anyway...

alastair:
We could use a simple http://www.fileformat.info/info/unicode/char/25b6/index.htm or http://www.fileformat.info/info/unicode/char/203a/index.htm for an arrow instead of an additional dependency

alastair:
(fontawesome also has a nice arrow, but I don't think that we should add it if we don't already use it. Something to think about for the next version? - http://fontawesome.io/icon/chevron-right/)

andrebola:
@ffont Make sure to clear the cache of the browser before, since I changed all.css file.

Issue #710, opened 2016-02-18
Don't try and blindly encode unicode in paginator


alastair:
We should only encode parameters when they're unicode

https://github.com/MTG/freesound/blob/master/general/templatetags/paginator.py#L52-L57

I also don't like the way of manually constructing parts of the query parameters. An alternative way of doing this would be to generate the first/last/next/prev links in code and render them directly in the template

PR #991, opened 2017-10-24
Move generation links of paginated pages outside templates

andrebola:
issue #710

Now the links to the first/last/next/last page are generated from python
code and passed to the templates using the context.

### DIFF
--- a/general/templatetags/paginator.py
+++ b/general/templatetags/paginator.py
@@ -48,13 +48,22 @@ def show_paginator(context, paginator, page, current_page, request, anchor="", n
 
     # although paginator objects are 0-based, we use 1-based paging
     page_numbers = [n for n in range(min_page_num, max_page_num) if n > 0 and n <= paginator.num_pages]
-    
+
     params = urllib.urlencode([(key, value.encode('utf-8')) for (key, value) in request.GET.items() if key.lower() != u"page"])
-    
+
     if params == "":
         url = request.path + u"?page="
     else:
         url = request.path + u"?" + params + u"&page="
+
+    url_prev_page = None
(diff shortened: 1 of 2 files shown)


general/templatetags/paginator.py
```diff
@@ -48,14 +48,31 @@ def show_paginator(context, paginator, page, current_page, request, anchor="", n
...
     else:
         url = request.path + u"?" + params + u"&page="
 
+    url_last_page = None
+
+    # Page could be a dict or an object
+    if isinstance(page, dict):
+        url_prev_page = url + str(page['previous_page_number'])
+        url_next_page =  url + str(page['next_page_number'])
+        url_first_page = url + '1'
```
ffont: Why we don't have `url_last_page` in this case?
I tested locally and the first thing I tried (clicking on the last page) used `page=None` which obviously does not work...

general/templatetags/paginator.py
```diff
@@ -48,14 +48,31 @@ def show_paginator(context, paginator, page, current_page, request, anchor="", n
...
+
     if params == "":
         url = request.path + u"?page="
     else:
         url = request.path + u"?" + params + u"&page="
 
+    url_last_page = None
+
+    # Page could be a dict or an object
+    if isinstance(page, dict):
```
ffont: Ok so apparently we can have `page` as an object or a dict. We should look into why this happens and maybe change it if it makes sense. If it does not make sense to change it, then we should document this here (like explain the dict and object cases).

general/templatetags/paginator.py
andrebola: Ok, solved!

Issue #596, opened 2014-03-14
Revise pack editing for large packs


ajaysmurthy:
I am working with the pack http://www.freesound.org/people/ajaysm/packs/14056

I get a 404 Not Found Error whenever I edit the pack description and submit the edits. The pack in question has 236 sounds and was uploaded to FS using a script. 

It might be a django timeout.

alastair:
We still need pagination. I'm not sure if we completely fixed the saving too - I believe sound saving/pack generation can still be made faster

ffont:
After some more implementation work and more discussion we concluded there are a number of things which are not right in the pack edit page:

- There is no pagination for the current sounds of the pack nor for the available user sounds
- The code for saving the pack is a bit messy and not optimised: https://github.com/MTG/freesound/blob/4e4b91fa45adabb5294487eeed384596425fd8bb/sounds/forms.py#L154-L182

The first point is mostly for UI purposes, but simple implementations of a paginator won't work as we need to be able to keep on adding/removing sound ids to a js variable while switching pages. UI-wise, the best would probably to do an ajax paginator here. Nevertheless this is not a commonly used feature, and for small packs the current UI will be ok. Considering that we're updating the whole interface soon then we decided there is no need to work on the pagination issues now.

Regarding the second point it is still worth looking at it, but it is not main priority.

ffont:
The pack editing page has changed with the new UI. It is still problematic to edit large packs, but I'll close this issue an open a new one which can state the problem better.

PR #997, opened 2017-10-26
Add pagination to sounds in pack edit page

andrebola:
issue #596

### DIFF
--- a/sounds/views.py
+++ b/sounds/views.py
@@ -495,7 +495,7 @@ def pack_edit(request, username, pack_id):
             return HttpResponseRedirect(pack.get_absolute_url())
     else:
         form = PackEditForm(instance=pack, initial=dict(pack_sounds=pack_sounds))
-        current_sounds = list(Sound.objects.bulk_sounds_for_pack(pack_id=pack.id))
+        current_sounds = Sound.objects.filter(pack_id=pack.id)
 
     paginate_data = paginate(request, current_sounds, settings.SOUNDS_PER_PAGE)
     paginator = paginate_data['paginator']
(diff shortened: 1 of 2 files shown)


alastair:
I'm not sure why you closed #996, but in this case it's OK to force push to overwrite the commits if no one has reviewed the PR yet.

andrebola:
I closed #996 because I accidentally mixed commits from other branch. I was avoiding to force push, but it might be the right thing.

sounds/views.py
```diff
@@ -495,11 +495,19 @@ def pack_edit(request, username, pack_id):
             return HttpResponseRedirect(pack.get_absolute_url())
     else:
         form = PackEditForm(instance=pack, initial=dict(pack_sounds=pack_sounds))
-        current_sounds = Sound.objects.bulk_sounds_for_pack(pack_id=pack.id)
+        current_sounds = list(Sound.objects.bulk_sounds_for_pack(pack_id=pack.id))
```
ffont: Why do we do the `list()` here? Is this because otherwise when we do `paginate` it throws some errors? Django's paginator is supposed to receive a queryset and then runs the query adding a range at the end. In our case we probably can't do that because `Sound.objects.bulk_sounds_for_pack()` does not accept a range at the end (am I right?). Anyway, if we turn this into a list, then the query is executed and we get the info for all sounds, not only for those needed for the `n-th` page.

The point is the paginator is to improve the speed of the page and also the UI. I think this solution only improves the UI but not the speed of the page. I'm not sure however how should be go about it.

andrebola: yes, you're right. I changed the bulk query for a regular django query

ffont:
Closing this PR because there are still many thinks to improve to the current pagination implementation and this is not a priority as we're changing the whole interface soon. For more details see new comments added in the issue: https://github.com/MTG/freesound/issues/596

PR #1001, opened 2017-10-26
Remove unused freesound-specific PermissionDenied exception

alastair:
A freesound-specific PermissionDenied exception had an exception handler
in some custom middleware, however this exception hadn't been used for
some time because an import in sounds/views.py called
  from sounds.forms import *
which overrode the previous PermissionDenied class with the Django
equilavent which had been imported in forms.py (in 14edc52a0)


ffont:
Closing this PR as custom `PermissionDenied` exception is removed in https://github.com/MTG/freesound/pull/1001

Issue #988, opened 2017-10-23
Improve speed of solr sound indexing


alastair:
The `manage.py reindex_solr` command sends data about all sounds to solr, however this process takes a long time. The biggest bottleneck is due to many queries being made to the database for each sound.

It looks like we tried to reduce the number of queries by using `select_related`, but we still need 2-3 queries per sound. Ideally we would do just one query for each block of sounds (default 4000).
We could use something like `bulk_query` in the `Sound` manager, but there are a number of fields which we need in solr that aren't needed in a normal sound display. For performance reasons, we shouldn't query these extra fields in a normal sound display so perhaps we need a variation of `bulk_query` for solr too.

andrebola:
I started to work on this, using bulk_query with 4000 sounds takes around 30 seconds to index each block of sounds in my local. So it takes around 50 minutes to index all the sounds.

Issue #989, opened 2017-10-23
Keeping the search index fresh


bdejong:
There are a lot of actions which don't invalidate a sound in the Solr index. For example: renaming a user, downloading a sound, rating a sound, ...

This means that the Solr index is almost always partly outdated.

It might be interesting thinking about some strategies on how to solve this...

PR #1002, opened 2017-10-27
Use bulk_query to reindex all sounds

andrebola:
issue #988

### DIFF
--- a/search/management/commands/post_dirty_sounds_to_solr.py
+++ b/search/management/commands/post_dirty_sounds_to_solr.py
@@ -42,10 +42,7 @@ class Command(BaseCommand):
         logger.info("Starting posting dirty sounds to solr. %i sounds to be added/updated to the solr index"
                     % num_sounds)
 
-        num_correctly_indexed_sounds = add_all_sounds_to_solr(sounds_to_index)
-
-        if num_correctly_indexed_sounds == num_sounds:
-            sounds_to_index.update(is_index_dirty=False)
+        num_correctly_indexed_sounds = add_all_sounds_to_solr(sounds_to_index, mark_index_clean=True)
 
         logger.info("Finished posting dirty sounds to solr. %i sounds have been added/updated"
                     % num_correctly_indexed_sounds)
--- a/search/management/commands/post_dirty_sounds_to_solr.py
+++ b/search/management/commands/post_dirty_sounds_to_solr.py
@@ -23,7 +23,7 @@ import logging
 from django.core.management.base import BaseCommand
 
 from sounds.models import Sound
-from utils.search.search_general import add_sounds_to_solr, delete_sound_from_solr, check_if_sound_exists_in_solr
+from utils.search.search_general import add_all_sounds_to_solr, delete_sound_from_solr, check_if_sound_exists_in_solr
 from utils.search.solr import SolrException
 
 logger = logging.getLogger("web")
@@ -36,26 +36,16 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
 
-        # Index all those which are processed and moderated ok
-        num_sounds = Sound.objects.filter(processing_state="OK", moderation_state="OK", is_index_dirty=True).count()
+        # Index all those which are processed and moderated ok that has is_index_dirty
+        sounds_to_index = Sound.objects.filter(processing_state="OK", moderation_state="OK", is_index_dirty=True)
+        num_sounds = sounds_to_index.count()
         logger.info("Starting posting dirty sounds to solr. %i sounds to be added/updated to the solr index"
                     % num_sounds)
 
-        num_correctly_indexed_sounds = 0
-        slice_size = 1000
-        for i in range(0, num_sounds, slice_size):
-            console_logger.info("Adding %i sounds to solr, slice %i", slice_size, i)
-            try:
-                # Get all sounds moderated and processed ok that has is_index_dirty
-                where = "sound.moderation_state = 'OK' AND sound.processing_state = 'OK' AND is_index_dirty = true " \
-                        "AND sound.id > %s"
-                order_by = "sound.id ASC"
-                sounds_qs = Sound.objects.bulk_query_solr(where, order_by, slice_size, (i, ))
-                add_sounds_to_solr(sounds_qs)
-                num_correctly_indexed_sounds += slice_size
-            except SolrException as e:
-                console_logger.error("failed to add sound batch to solr index, reason: %s", str(e))
-                raise
+        num_correctly_indexed_sounds = add_all_sounds_to_solr(sounds_to_index)
+
+        if num_correctly_indexed_sounds == num_sounds:
+            sounds_to_index.update(is_index_dirty=False)
 
         logger.info("Finished posting dirty sounds to solr. %i sounds have been added/updated"
                     % num_correctly_indexed_sounds)
--- a/search/management/commands/post_dirty_sounds_to_solr.py
+++ b/search/management/commands/post_dirty_sounds_to_solr.py
@@ -18,11 +18,16 @@
 #     See AUTHORS file.
 #
 
+import logging
+
 from django.core.management.base import BaseCommand
+
 from sounds.models import Sound
 from utils.search.search_general import add_sounds_to_solr, delete_sound_from_solr, check_if_sound_exists_in_solr
-import logging
+from utils.search.solr import SolrException
+
 logger = logging.getLogger("web")
+console_logger = logging.getLogger("console")
 
 
 class Command(BaseCommand):
@@ -42,7 +47,8 @@ class Command(BaseCommand):
             console_logger.info("Adding %i sounds to solr, slice %i", slice_size, i)
             try:
                 # Get all sounds moderated and processed ok that has is_index_dirty
-                where = "sound.moderation_state = 'OK' AND sound.processing_state = 'OK' AND is_index_dirty = true AND sound.id > %s"
+                where = "sound.moderation_state = 'OK' AND sound.processing_state = 'OK' AND is_index_dirty = true " \
+                        "AND sound.id > %s"
                 order_by = "sound.id ASC"
                 sounds_qs = Sound.objects.bulk_query_solr(where, order_by, slice_size, (i, ))
                 add_sounds_to_solr(sounds_qs)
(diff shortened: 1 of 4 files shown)


bdejong:
this also solves #989 actually, if we do this every day... :)

ffont:
First thing to do is to now why we're doing so many queries per sound, which means to check in [`convert_to_solr_document`](https://github.com/MTG/freesound/blob/master/utils/search/search_general.py#L38) which attributes that we request are not included in [`bulk_query`](https://github.com/MTG/freesound/blob/936b5d233fde0be045f48e40ee10793f166ae27e/sounds/models.py#L127) select. Also, there are some attributes which are already selected in `bulk_query` but we're asking for them again in `convert_to_solr_document` (see the tags for example). 

For those attributes which are not selected in `bulk_query`, we should evaluate if it makes sense to add them in this select or if we should simply do another query to get all of them at once at each iteration of indexing sounds (this is what @alastair proposes, I'd mean to maybe have 4-5 queries per loop instead of only 1).

Finally, If making changes to `convert_to_solr_document` make sure that it is always given a sound object returned from `bulk_query`, not from a normal select.

andrebola:
tags are selected in bulk_query:
https://github.com/MTG/freesound/blob/936b5d233fde0be045f48e40ee10793f166ae27e/sounds/models.py#L153-L157

But they are not used in`convert_to_solr_document` , this could be fixed easily, and make the same with the others like Comments.
The biggest problem that I see is with `locations()` because they are generated dynamically. Any idea about how to generate that without using the Sound object?

alastair:
I don't think we should use `bulk_query` here - the biggest reason is that this means that we would have to add Comments (and maybe other things) to it. We don't want to do this because it'll make it slower for every other case that we use this query (including all sound page loads).

Instead, I'm interested in seeing why we do ~4 queries per sound (1626 queries per 400 sounds) in the loop - what things are not selected in the `Sound.objects.select_related` that cause us to do a separate query? 

As an example, is it possible to select tags for all 4000 sounds in a single query, instead of doing it once per sound 4000 times? What other statements in `convert_to_solr_document` cause a database query to be run?

andrebola:
Yes, I also think we should use a different method in the case we want to use bulk queries. 

I can check what is not selected by select_related and see how many queries are done. Also note that we could use a data importer handler that connects directly to the database and makes the query:

https://lucene.apache.org/solr/guide/6_6/uploading-structured-data-store-data-with-the-data-import-handler.html

I think it also works for the version that we are currently using.

ffont:
To keep simplicity I'd not use the data importer handler and simply add more queries for the bits of information that are missing. I'd make these as extra methods for the manager (like `bulk_query`).

andrebola:
@alastair note that you have to use: `from django.db import reset_queries` otherwise the number of queries will always increment

search/management/commands/reindex_solr.py
```diff
@@ -18,20 +18,84 @@
...
+from utils.text import remove_control_chars
+from utils.search.search_general import delete_sound_from_solr
 from sounds.models import Sound
-from utils.search.search_general import add_all_sounds_to_solr, delete_sound_from_solr
+
+console_logger = logging.getLogger("console")
+
+def convert_to_solr_document(sound_obj):
+    document = {}
+    sound = sound_obj.__dict__
```
alastair: I'm not really a fan of this. Can you use `getattr` instead?

search/management/commands/reindex_solr.py
```diff
@@ -18,20 +18,84 @@
...
+    document["waveform_path_m"] = locations["display"]["wave"]["M"]["path"]
+    document["waveform_path_l"] = locations["display"]["wave"]["L"]["path"]
+    document["spectral_path_m"] = locations["display"]["spectral"]["M"]["path"]
+    document["spectral_path_l"] = locations["display"]["spectral"]["L"]["path"]
+    document["preview_path"] = locations["preview"]["LQ"]["mp3"]["path"]
+    return document
+
+
+def add_sounds_to_solr(sounds):
+    console_logger.info("adding multiple sounds to solr index")
```
alastair: we could say how many sounds we're adding here, instead of *multiple*

search/management/commands/reindex_solr.py
```diff
@@ -18,20 +18,84 @@
...
+    solr = Solr(settings.SOLR_URL)
+    console_logger.info("creating XML")
+    documents = map(convert_to_solr_document, sounds)
```
alastair: Not big, but we could use a list comprehension here instead of map

search/management/commands/reindex_solr.py
```diff
@@ -18,20 +18,84 @@
...
+import logging
+import math
+
+from django.conf import settings
 from django.core.management.base import BaseCommand
+from utils.search.solr import Solr, SolrException
+from utils.text import remove_control_chars
+from utils.search.search_general import delete_sound_from_solr
 from sounds.models import Sound
-from utils.search.search_general import add_all_sounds_to_solr, delete_sound_from_solr
```
alastair: did you delete this function from `search_general`? I don't think it's used anywhere else other than this command.

search/management/commands/reindex_solr.py
```diff
@@ -19,27 +19,79 @@
...
+from utils.search.solr import Solr, SolrException
+from utils.text import remove_control_chars
+from utils.search.search_general import delete_sound_from_solr
 from sounds.models import Sound
-from utils.search.search_general import add_sounds_to_solr, delete_sound_from_solr
-from utils.search.solr import SolrException
 
 console_logger = logging.getLogger("console")
 
+def convert_to_solr_document(sound_obj):
```
ffont: Now the function `convert_to_solr_document` is duplicated here and in `utils.search.search_general`. I think we should have only one version of this function and make sure that it works from whatever place it is called.

sounds/models.py
```diff
@@ -124,6 +124,53 @@ def random(self, excludes=None):
         else:
             return None
 
+    def bulk_query_solr(self, where, order_by, limit, args):
+        query = """SELECT
```
ffont: Please add documentation here to explain why we added this method, which is a version of `bulk_query` which selects all relevant information for indexing sounds to solr.

search/management/commands/reindex_solr.py
andrebola: First I did that, but then I moved it here because in one case I have to convert the sound object to a dict and in the other case I have to deal with the object. After that the logic if different for the 2 cases, so I think it makes more sense to have 2 methods.

search/management/commands/post_dirty_sounds_to_solr.py
```diff
@@ -18,30 +18,51 @@
...
-        num_correctly_indexed_sounds = add_all_sounds_to_solr(sounds_dirty_to_index, mark_index_clean=True)
+                    % num_sounds)
+
+        num_correctly_indexed_sounds = 0
+        slice_size = 1000
+        for i in range(0, num_sounds, slice_size):
+            console_logger.info("Adding %i sounds to solr, slice %i", slice_size, i)
+            try:
+                # Get all sounds moderated and processed ok that has is_index_dirty
+                where = "sound.moderation_state = 'OK' AND sound.processing_state = 'OK' AND is_index_dirty = true " \
```
ffont: Regarding the `where` clause, the part of `sound.moderation_state = 'OK' AND sound.processing_state = 'OK'` could be part of `Sound.objects.bulk_query_solr` because we're never going to send sounds which are not moderated or processed OK to Solr. Then what you pass in the function would be some sort of "extra where" clause.

Also the condition you use `AND sound.id > %s` is wrong because you are comparing sound IDs with the counter `i`. Sound IDs are not necessarily sequential (even though you sort them) as when sounds are deleted there might be missing IDs. Therefore what happens is that sound IDs grow faster than `i` (`num_sounds` < most recent sound ID). Once the loop is finished, there will still be some sounds not indexed, and also I think we will be sending some sounds twice. I indeed tried that locally and while postgres told me I have 300k sounds moderated and processed ok, after a full reindex I only had 270k in solr.

I understand that you use this strategy to avoid having to retrieve all sound IDs and doing some sort of `WHERE id IN ...`. I see two options:
* Retrieve all sound IDs and iterate them in chunks and do the query. In this case at each iteration we'll be sending 1000 unique sounds to solr.
* Get the id of the most recent sound and use this in the `for i in range(0, num_sounds, slice_size)` instead of `num_sounds`. In this case you won't have to get the IDs but at each iteration we might be sending some sounds that were already sent in the last iteration. 

I guess first option is the best. What do you think @alastair and @andrebola ?

Note that this comments affect both `post_dirty_sounds_to_solr` and `reindex_solr` commands.

search/management/commands/post_dirty_sounds_to_solr.py
```diff
@@ -36,26 +36,16 @@ class Command(BaseCommand):
...
-                sounds_qs = Sound.objects.bulk_query_solr(where, order_by, slice_size, (i, ))
-                add_sounds_to_solr(sounds_qs)
-                num_correctly_indexed_sounds += slice_size
-            except SolrException as e:
-                console_logger.error("failed to add sound batch to solr index, reason: %s", str(e))
-                raise
+        num_correctly_indexed_sounds = add_all_sounds_to_solr(sounds_to_index)
+
+        if num_correctly_indexed_sounds == num_sounds:
+            sounds_to_index.update(is_index_dirty=False)
```
ffont: Why do we do this here? If a batch of sounds can't be indexed for some reason (see `try/except` in `add_all_sounds_to_solr` ), then it can happen that `num_correctly_indexed_sounds != num_sounds` and we wont mark them as index dirty. I think we should use the `mark_index_clean` option passed to `add_all_sounds_to_solr` that we had before (https://github.com/MTG/freesound/blob/f17064bf5d5419a3657dcd3dacf7d094d0b227d0/utils/search/search_general.py#L101-L103).

search/management/commands/reindex_solr.py
```diff
@@ -34,19 +33,11 @@ class Command(BaseCommand):
...
-                sounds_qs = Sound.objects.bulk_query_solr(where, order_by, slice_size, (i, ))
-                add_sounds_to_solr(sounds_qs)
-            except SolrException as e:
-                console_logger.error("failed to add sound batch to solr index, reason: %s", str(e))
-                raise
+        # Get all sounds moderated and processed ok
+        sounds_to_index = Sound.objects.filter(processing_state="OK", moderation_state="OK")
+        console_logger.info("Reindexing %d sounds to solr", sounds_to_index.count())
+
+        add_all_sounds_to_solr(sounds_to_index)
```
ffont: I'd be good to add an option to the command to mark all sounds as index clean and make it default to `True`. Pass this option to `add_all_sounds_to_solr` (see comment above) so that sounds get marked as clean.

search/management/commands/post_dirty_sounds_to_solr.py
andrebola: The main reason of doing it in this way is to avoid doing one more query on each iteration. Also, I thought that if the indexing fails then it will fail for all the sounds so the condition was fine.

ffont:
I see. There are several reasons why indexing could fail. If it fails because of solr being down then you’re right. However it can also fail because one single sound fails converting to the xml (it happened in the past although we have been adding fixes for that). In that case we would not mark any sound as clean even though some would have been indexed ok.

Issue #999, opened 2017-10-26
Can only unsubscribe to a forum thread on notification emails


alastair:
If you are subscribed to a forum thread, you get an email when someone posts a reply in that thread. There is a link in the email to unsubscribe from the thread, but if you visit the thread there is no link to unsubscribe.
Similarly there is only the option to subscribe to a thread when you post a reply. Could we also have a link on threads to subscribe?

PR #1003, opened 2017-10-27
Add link to (Un)Subscribe to thread

andrebola:
issue #999


forum/views.py
```diff
@@ -111,11 +111,15 @@ def thread(request, forum_name_slug, thread_id):
     paginator = paginate(request, Post.objects.select_related('author', 'author__profile').filter(
         thread=thread, moderation_state="OK"), settings.FORUM_POSTS_PER_PAGE)
 
+    has_subscription = False
     # a logged in user watching a thread can activate his subscription to that thread!
     # we assume the user has seen the latest post if he is browsing the thread
     # this is not entirely correct, but should be close enough
     if request.user.is_authenticated:
-        Subscription.objects.filter(thread=thread, subscriber=request.user, is_active=False).update(is_active=True)
+        subscription = Subscription.objects.filter(thread=thread, subscriber=request.user)
```
alastair: This is not a big thing, but reading this code makes me unsure how many subscriptions a user can have for a thread.  In the `Subscription` model, `Meta` marks thread/subscriber as unique_together.
This means that 
```py
try:
    subscription = Subscription.objects.get(thread=thread, subscriber=request.user, is_active=False)
    subscription.is_active = True
    subscription.save()
except Subscription.DoesNotExist:
    pass
```
is also a valid way of doing this check, and makes it clearer that there is only one Subscription in this case.

andrebola: it's not exactly the same, if you want to do it with a get instead of a filter it would be:

```python
try:
    subscription = Subscription.objects.get(thread=thread, subscriber=request.user)
    if not subscription.is_active:
        subscription.is_active = True
        subscription.save()
    has_subscription = True
except Subscription.DoesNotExist:
    pass
```
I'm not sure if this code is better than the other.

alastair: I see the difference, you're right. I still prefer the example you just gave, because as I said it makes it explicit that there is only one subscription object per user+thread

alastair:
This is good to go!

PR #1005, opened 2017-10-30
Replace faad2 for ffmpeg for processing m4a files

andrebola:
issue #1004

Tested in test.freesound.org and it works fine for the same m4a file:

http://test.freesound.org/people/test/sounds/274383/

### DIFF
--- a/utils/audioprocessing/processing.py
+++ b/utils/audioprocessing/processing.py
@@ -494,7 +494,7 @@ def convert_to_pcm(input_filename, output_filename):
     elif sound_type == "flac":
         cmd = ["flac", "-f", "-d", "-s", "-o", output_filename, input_filename]
     elif sound_type == "m4a":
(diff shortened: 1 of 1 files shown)


ffont:
Hi,

I think this is not what is discussed in https://github.com/MTG/freesound/issues/1004.
In my opinion the fix is to remove the `--ac 1` in the default `ffmpeg` converter. 

If we want to stop using `faad2` for m4a files and use `ffmpeg` instead, we could do the same change for all other formats. I'm nor sure if there is a reason why we use the individual encoders but I would continue with the same approach as we did in the past. In the future we could consider to only use `ffmpeg`.

andrebola:
Yes, it's true. I created a new PR #1006

ffont:
Closing this PR as it is invalid after https://github.com/MTG/freesound/pull/1006

PR #1006, opened 2017-10-30
Change default conversion parameter of ffmpeg for 2 channels

andrebola:
issue  #1004


ffont:
What I would do is to not specify any number of channels so that the conversion respects the original file. I'll update the code.

PR #1011, opened 2017-10-31
Remove locals() from geotags

alastair:
Perform style cleanups on geotags code (formatting, pep8, etc).

I'm adding this PR now to get feedback on my changes made in 5fb6c568dda34f2cb09e6608820931330524923c. I like the idea of moving the `GOOGLE_API_KEY` setting out of the context processor, but I'm not sure of the best way of providing the template tag. One way is as I've commited, where `{% google_maps_js %}` will include a full
`<script type="....." src="//..........js"></script>` string. However, it's a bit opaque as to exactly what is being inserted with this tag. An alternative would be to have the tag only return the URL, so that we would use it as
```js
<script type="text/javascript" src="{% google_maps_js %}"></script>
```
which immediately makes it clear that this tag inserts a URL of some js that we have to load.

Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
 - pip install django-nose
 - pip install -r requirements.txt
 script:
-- python manage.py test --noinput --settings=freesound.test_settings accounts apiv2 bookmarks donations follow forum ratings search sounds support tags tickets utils wiki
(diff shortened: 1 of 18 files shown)


ffont:
I like the second option:

```js
<script type="text/javascript" src="{% google_maps_js %}"></script>
```

alastair:
Ready for review @ffont

Issue #669, opened 2015-06-04
Solr 7


alastair:
solr 5 is out. It's new and shiny and doesn't support being run as a .war any more. We could consider building new packages based on this. My maven template is out of date now, I'm not sure what the best way to build with additional dependencies is.

alastair:
solr 6 is out too!
It appears to have geospatial search: https://lucene.apache.org/solr/guide/6_6/spatial-search.html, which might be able to replace our additional dependency.
We should check that the query syntax is the same, as we pass this syntax from the FS API to solr directly. Double-check if anyone actually uses this search, or perform a rewrite of the query syntax before passing to solr.

alastair:
Looks like polygon search still needs JTS: https://lucene.apache.org/solr/guide/6_6/spatial-search.html#SpatialSearch-JTSandPolygons

andrebola:
Last week solr 7 was released, I will try to set up everything for that version.

alastair:
:man_facepalming:

ffont:
Closing this issue as is no longer relevant after all changed we've been doing related to solr

PR #1012, opened 2017-11-02
Solr 7.1.0 support

andrebola:
issue #669

In order to use solr 7.1.0, once is downloaded and started, I used the web admin to create a new core for which I used the config files that in utils/search/solr7.1.0/solr.home/fs2/

### DIFF
--- a/utils/search/solr.py
+++ b/utils/search/solr.py
@@ -426,7 +426,6 @@ class Solr(object):
 
         if response.status != 200:
             raise SolrException, response.reason
-        
         return response
 
     def select(self, query_string, raw=False):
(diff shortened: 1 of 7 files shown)


utils/search/solr.py
```diff
@@ -426,7 +426,6 @@ def _request(self, query_string="", message=""):
 
         if response.status != 200:
             raise SolrException, response.reason
-        
         return response
 
     def select(self, query_string, raw=False):
```
alastair: We've been thinking about replacing this file with a maintained 3rd party solr library instead of our custom connection code. Do you think it's a good time replace it as part of this PR, or should we release as-is and then change the library at a later stage?

andrebola: For what I could see this works without much changes, so I think it would be better to do it in two steps.

alastair:
In our existing setup we have JTS loaded as a dependency by using https://github.com/MTG/freesound-deploy/tree/master/solr, which generates a war with the dependency enabled.
As far as I understand, new versions of solr don't run as a war any more. How did you set this up with the additional dependency?  It would be nice to have easy installation for the new version of solr (including in https://github.com/MTG/freesound-deploy/blob/master/deploy/search.sh if possible) - that is, we shouldn't have to perform manual steps to install and configure the new solr.

alastair:
Has anything changed between these two versions with regards to ranking, etc? Do the same queries on both cores give the same results?  How much of a test suite do we have for solr currently? @ffont?

utils/search/solr.py
alastair: OK, cool. Can you open a new issue for this?

andrebola:
I check the order returned for the queries and they are not the same, even that the configuration is the same and the query is the same there are some differences in the order returned for the sounds. 

Looking deeply in the sounds returned by solr 4.5 I don't understand why gives more priority to the sound filename if in the query is the least important: `id^4 tag^4 description^3 username^1 pack_tokenized^2 original_filename^2`

For example, look the second item for the query "explode": http://freesound.org/search/?q=explode
In that case the description doesn't contain the word, if I run the same query with solr7.1 that sound is shown in page 4.

andrebola:
There is another difference that I found regarding filters. Previously when you use something like: `fq=tag:music type:mp3` it used to work as an AND of the two filters but now it works as an OR. We use to do that in the search filters from the frontend and also we allowed that syntax using the api (as is [documented](http://freesound.org/docs/api/resources_apiv2.html#search-resources)).

Something that used to work and still does is to do multiple fq, the previous example would be ` fq=tag:music&fq=type:mp3` that works as an AND of the filters. So, my idea was to replace internally when we receive something of the form `fq=filter1 filter2 filter3` for something like `fq=filter1&fq=filter2&filter3` 

The problem that I see here is that either we change how the API works now adapting to how solr 7 works or we make changes on the query of the user and we remove from the docs where we say that we follow the syntax of solr. I see adv. and disadv. in both cases. 

We discussed previously the idea of having a apiv3 where we change the way people make queries, I just mention that as a third option, but I don't think we need to take such a drastic solution.

andrebola:
About geospacial filters, I found that basic queries work without using JTS, like this one:

`{!geofilt sfield=geotag pt=<LATITUDE>,<LONGITUDE> d=<MAX_DISTANCE_IN_KM>}`

but more complex queries are not supported like:

`geotag:"Intersects(<MINIMUM_LONGITUDE> <MINIMUM_LATITUDE> <MAXIMUM_LONGITUDE> <MAXIMUM_LATITUDE>)"`

ffont:
After discussion we decided to set this PR on hold for now. There are many big changes coming and apparently newer versions of SOLR might help us with the spatial search "problems" so let's come back to this in some time from now. I'm not closing the PR, just adding [WIP] at the title so that we don't accidentally remove the branch.

ffont:
Hi @andrebola,
We have finally been working on updating solr and have reused some of your work here. The discussion here were also very helpful to work again on Solr. We're now running with Solr5 in production, but are already working on Solr9. I'll close this PR because code has diverged and we'll not use it anymore. Thanks for your 5 years old efforts @andrebola, they have been actually very helpful in the past months :)

Issue #1017, opened 2017-11-07
Disable grouping by pack doesn't work


andrebola:
If you disable the option "Group sounds by pack" in the advanced options of sound search and then submit the results are still grouped by pack the checkbox appears checked again.

ffont:
This is probably because in my recent refactorings in the search view I forgot to add some variable to `tvars`. I'd be nice to implement a test for this.

PR #1018, opened 2017-11-08
Fix grouping by pack

ffont:
Fixes issue https://github.com/MTG/freesound/issues/1017

### DIFF
--- a/templates/search/search.html
+++ b/templates/search/search.html
@@ -216,7 +216,13 @@
         $("#filter_query").val(cleaned);
     }
 
-
+    function set_hidden_grouping_value(){
+        if ($("#grouping_geotagged").is(':checked')){
+            $("#grouping_geotagged_hidden").val("1")
+        }else{
+            $("#grouping_geotagged_hidden").val("")
+        }
+    }
 
 </script>
 
--- a/templates/search/search.html
+++ b/templates/search/search.html
@@ -79,21 +79,6 @@
 
     });
 
-    function showAdvancedSearchOptions()
-    {
-        $("#advanced_search_hidden").val(1);
-        $("#enableShowButton").hide();
-        $("#advanced_search_options").show('blind', {}, 500, function() {});
-
-    }
-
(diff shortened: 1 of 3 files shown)


ffont:
The problem was that a JS function was accidentally removed in commit 0840048.

templates/search/search.html
```diff
@@ -216,7 +216,13 @@
         $("#filter_query").val(cleaned);
     }
 
-
+    function set_hidden_grouping_value(){
```
alastair: Part of the reason for doing the original commit where this was removed was to get rid of javascript that isn't strictly required for some search features. In this case, restoring this function makes the checkbox work again, but after looking at the code I'm confused as to why we need it
https://github.com/MTG/freesound/blob/4e4b91fa45adabb5294487eeed384596425fd8bb/templates/search/advanced_search_form.html#L63-L64
Why don't we update it to make the checkbox the field named `g` instead of the hidden field? This will let us get rid of this extra js.
While looking at the change to `actual_grouping`, I also wonder why it's used. To me I don't see a reason to assign another variable and use it instead of `grouping`:
https://github.com/MTG/freesound/blob/e289acff01bf9159cb9c61d30fbfd8afef0c45f3/search/views.py#L162

ffont: Yes you're right, I see no apparent reasons why these changes could not be done. I didn't do them before because I did not want to spend more time with the fix, but given that these are simple and easy to test (manually), I'll implement them.

ffont:
I did some refactoring and cleaning. The thing with the hidden grouping input I did not change because the problem is that we want the checkbox to be on by default, and there is no way in Django of differentiating from a checkbox which is explicitly turned off or a checkbox that "does not exist". HTML forms only send checkbox values when these are checked. The use of a hidden sister field is a common approach. The alternative now would be to change the semantic meaning of the checkbox to "Don't group by pack" and have it off by default, but I don't want to introduce such changes now, specially considering that new interface is coming.

Issue #1016, opened 2017-11-07
DataError: value too long for type character varying(254)


alastair:
We need to make sure that either the form knows about our maximum lengths in the database and rejects it at the validation stage, or that we catch these SQL errors when we save objects.

We should check this behaviour for all fields in the database that have length restrictions, not just the email field.

https://logserver.mtg.upf.edu/sentry/freesound-web/issues/822/

```
DataError: value too long for type character varying(254)

(13 additional frame(s) were not displayed)
...
  File "utils/session_checks.py", line 36, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "accounts/views.py", line 191, in registration
    user = form.save()
  File "accounts/forms.py", line 203, in save
    user.save()

DataError: value too long for type character varying(254)

```

PR #1023, opened 2017-11-13
Email validate

andrebola:
issue #1016 

I think the best way of handling validation of length in char fields is to pass max_length in the definition of forms fields, in this way the length of the field is validated by javascript and also in the backend automatically.

The only way to make sure that we always use the max_length when the model is restricted is to use a Model Form, but sometimes this makes the code unnecessarly complex for simple cases.

For this reasons I manually checked all the length restrictions in the models and checked that the corresponding forms had the max_length parameter.


andrebola:
I think the tests are related with the changes I did in the code. Probably the name of the PR is not very accurate because the changes involve other fields as well.

Issue #1025, opened 2017-11-15
Re-Request password for self-delete


bdejong:
When a user deletes himself we don't request the password again.

This could lead to people deleting your account if you forgot to "log out" on a shared computer...

ffont:
I think this should be relatively easy by adding a `password` field in the `DeleteUserForm ` form which validates to check user password.

PR #1026, opened 2017-11-15
Add password to delete account form

andrebola:
issue #1025

Added password field to account delete form. Also changed how the form is validated and the error messages are displayed.

### DIFF
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -312,8 +312,15 @@ class DeleteUserForm(forms.Form):
     encrypted_link = forms.CharField(widget=forms.HiddenInput())
     delete_sounds = forms.ChoiceField(choices = DELETE_CHOICES,
             widget=forms.RadioSelect())
+    password = forms.CharField(label=_("Your password"), widget=forms.PasswordInput)
 
-    def clean_encrypted_link(self):
+    def clean_password(self):
+        user = User.objects.get(id=self.user_id)
+        if not user.check_password(self.cleaned_data["password"]):
+            raise forms.ValidationError(_("Incorrect password."))
+        return self.cleaned_data['password']
+
+    def clean(self):
         data = self.cleaned_data['encrypted_link']
         if not data:
             raise PermissionDenied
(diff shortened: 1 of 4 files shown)


bdejong:
YEAH! 👍

accounts/forms.py
```diff
@@ -312,8 +312,15 @@ class DeleteUserForm(forms.Form):
     encrypted_link = forms.CharField(widget=forms.HiddenInput())
     delete_sounds = forms.ChoiceField(choices = DELETE_CHOICES,
             widget=forms.RadioSelect())
+    password = forms.CharField(label=_("Your password"), widget=forms.PasswordInput)
```
bdejong: `Confirm your password`
perhaps?

accounts/forms.py
```diff
@@ -312,8 +312,15 @@ class DeleteUserForm(forms.Form):
...
 
-    def clean_encrypted_link(self):
+    def clean_password(self):
+        user = User.objects.get(id=self.user_id)
+        if not user.check_password(self.cleaned_data["password"]):
+            raise forms.ValidationError(_("Incorrect password."))
+        return self.cleaned_data['password']
```
bdejong: I guess you should return `None` or `""` here not to have the password be played back to the user? If the user messed up the password is still going to be repeated to the user from the backend... That's a bit messy.

accounts/tests.py
```diff
@@ -668,6 +668,17 @@ def test_user_delete_remove_sounds(self, delete_sound_solr):
...
+        # This should set user's attribute active to false and anonymize it
+        user = self.create_user_and_content(is_index_dirty=False)
+        a = self.client.login(username=user.username, password='testpass')
+        form = DeleteUserForm(user_id=user.id)
+        encr_link = form.initial['encrypted_link']
+        resp = self.client.post(reverse('accounts-delete'),
+                {'encrypted_link': encr_link, 'password': 'testpass', 'delete_sounds': 'delete_sounds'})
+
+        self.assertEqual(User.objects.get(id=user.id).profile.is_deleted_user, True)
+
```
bdejong: Should probably add a failing test too, especially one where the password is mistyped...

templates/accounts/delete.html
```diff
@@ -15,15 +15,15 @@
...
     <form method="post">
-        {% if error_message %}
-        <ul class="errorlist nonfield">
-            <li>{{ error_message }}</li>
-        </ul>
-        {% endif %}
       {% csrf_token %}
+      {{ delete_form.non_field_errors }}
       {{delete_form.encrypted_link}}
       {{delete_form.delete_sounds}}
```
bdejong: This is weird... Why is {{delete_form}} itself not in here?

accounts/tests.py
ffont: Yes, this test should be added.

Issue #967, opened 2017-09-29
Use @transaction.atomic() where required


ffont:
During the last months we've fixed some "violates unique constraint" errors which happen because requests that access the DB *more than once* are not done in a single transaction. In such situations, it can happen that two requests that are processed simultaneously (e.g. if users double-click on a submit form) cause problems. Some of this errors have been reported and fixed:

 * Add ratings: issue https://github.com/MTG/freesound/issues/818, pr https://github.com/MTG/freesound/pull/939
 * Edit tags: issue https://github.com/MTG/freesound/issues/812, pr https://github.com/MTG/freesound/pull/942
 * Create new user: issue https://github.com/MTG/freesound/issues/742, pr https://github.com/MTG/freesound/pull/944

There are also very old tickets that might be related:
 * Duplicate md5: https://github.com/MTG/freesound/issues/389

We should go through all our views and evaluate whether or not we should wrap them into `@transaction.atomic()`decorators (or find other solutions like in https://github.com/MTG/freesound/issues/894, with the use of [F-expressions](https://docs.djangoproject.com/en/dev/ref/models/expressions/#f-expressions)).

alastair:
Further to this, we are considering disabling buttons on form submit to prevent double-clicks, as the transaction protection won't prevent this error especially if we don't have a unique index on the thing that we create. More thoughts at https://github.com/MTG/freesound/pull/947#issuecomment-337842036

PR #1031, opened 2017-11-17
Disable form submit by adding class disable-form-submit

andrebola:
Moved to `freesound.js` the code to disable of the forms with class `disable-form-submit`  after submit.

Also added the class to the registration form so we can finally solve the problem:
https://logserver.mtg.upf.edu/sentry/freesound-web/issues/669/

issue #967

### DIFF
--- a/media/js/freesound.js
+++ b/media/js/freesound.js
@@ -65,6 +65,11 @@ function d()
 $(document).ready( function() {
     setupStarRatings();
     switchFormSubmits();
+
+    $(".disable-on-submit").submit(function () {
+       $(this).find('input[type="submit"]').attr("disabled", true);
+       return true;
+    });
 });
 
 var voted = {};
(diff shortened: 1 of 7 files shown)


ffont:
In this PR we are disabling form submit in two cases:

- Moderation tickets
- Registration page

But what happens with the other cases in which we've seen problems with double clicks or there are potential problems? I think we should either:

- go through all the forms and evaluate whether they need the disable on submit or not
- just add the disable on submit to all forms in Freesound (do it in `freesound.js` like `$('form').submit(...)`)

I feel inclined for option 2 now. Reasons for not doing this?

ffont:
After discussion on Hipchat we decided to add it to some known forms as doing it for all of them at once could be problematic. Let's do it for:

- Moderation tickets [done]
- Registration page [done]
- Sound describe
- Add comment to sound
- Post in forums (new topic and reply)

PR #1037, opened 2017-11-20
Add sentry logger for sound/pack download

andrebola:
issue #840

Also added the required handler in the logger configuration.


alastair:
Looks good to me

Issue #933, opened 2017-07-18
Can't create a new empty database with `manage.py migrate`


alastair:
If you run `python manage.py migrate` you get the following error:

```
django.db.utils.ProgrammingError: relation "django_site" does not exist
LINE 1: ..."django_site"."domain", "django_site"."name" FROM "django_si...
```

This is due to the API trying to do a DB query when the URLs are parsed:
```
  File "/code/accounts/urls.py", line 31, in <module>
    import apiv2.views as api
  File "/code/apiv2/views.py", line 67, in <module>
    docs_base_url = prepend_base('/docs/api')
  File "/code/apiv2/apiv2_utils.py", line 317, in prepend_base
    return "http://%s%s" % (Site.objects.get_current().domain, rel)
```
https://github.com/MTG/freesound/blob/52710d4ab640de50a3d69f9a7232fc7a81f8d6c6/apiv2/apiv2_utils.py#L314-L317

ffont:
This happens because the function [`prepend_base`](https://github.com/MTG/freesound/blob/d47d5302fcca6ae10ffe32a2c2581707f1b3e5cc/apiv2/apiv2_utils.py#L370) makes a DB query and we use it in several places in `apiv2.views` which are executed before DB is ready. This includes the use when we set [`docs_base_url`](https://github.com/MTG/freesound/blob/d47d5302fcca6ae10ffe32a2c2581707f1b3e5cc/apiv2/views.py#L63) and each call to `get_formatted_examples_for_view` when setting the docstring of all API views.

The problem with `docs_base_url` can be easily fixed by removing `prepend_base` (just assign `docs_base_url = '/docs/api'`). This is because here we just use a relative link to the corresponding docs page and don't show the URL to the user. However, in the case of `get_formatted_examples_for_view` we generate a number of example URLs that the user should be able to copy and paste and we need to show the full URL. The way to fix this is that instead of directly assigning `__doc__` to something which calls a function, we should populate `__doc__` when we access it (lazy). We could do this with a decorator. Then, we can store the generated contents in a global variable so that examples are not recomputed every time a user makes a request to the same API view.

PR #1041, opened 2017-11-20
Add exception handling when Site row doesn't exists

andrebola:
isue #933


ffont:
The fix you propose works but we had a closer look to the problem and decided that we should eliminate it from its root, therefore we're not merging this PR. I'll comment on the issue (https://github.com/MTG/freesound/issues/933) and close this PR for now.

Issue #779, opened 2016-12-29
Renaming packs does not clear the old SOLR index


bdejong:
http://freesound.org/search/?q=cuba+havana

see the first two packs mentioned (next to the ➕  sign): same username, same pack name, one is real the other isn't.

bdejong:
This is the correct pack:
http://freesound.org/people/kyles/packs/20401/

This is rather weird:
http://freesound.org/people/kyles/packs/3339/
exists, but contains no sounds (wuh?)

Could this mean there is there no unique(user_id, pack.name) index in the db?

andrebola:
I think the second pack was deleted but it wasn't updated on Solr. So, we should set index_dirty=True when the pack is deleted at line:

https://github.com/MTG/freesound/blob/7f64cdf864eee0f8af9d461232e72a9fb4ab5780/sounds/models.py#L838

ffont:
Well, I don't think the pack was deleted because the object is there, but apparently the sounds were moved from the old pack (3339) to the new one (20401, probably using pack edit feature (https://github.com/MTG/freesound/blob/master/sounds/views.py#L435). What might be happening is that once a pack is edited, the sounds assigned to the pack are set to reindex, but the ones that were removed from the pack don't seem to be set to reindex. We should set all affected sounds to reindex and I guess this will fix the problem.

andrebola:
I don't think that is the problem since the old and new sounds are marked as dirty on the save method of the form:

https://github.com/MTG/freesound/blob/master/sounds/forms.py#L177

So I don't see why the new sounds are marked in the view as well: https://github.com/MTG/freesound/blob/master/sounds/views.py#L465

ffont:
According to @andrebola comment we don't think this issue is still valid.  However, we seem to be doing the mark index dirty two times, at pack edit view and at pack form save.

andrebola:
I check the code again and some of the sounds might be marked as dirty two times but there are others sound related to the pack that will be marked dirty only once (in the edit view method).

andrebola:
I just tested this in my local and it worked fine. Changing pack name sets is_index_dirty as True and then the management command post_dirty_sounds_to_solr indexed the sound.

andrebola:
This issue is very related with #989. After merged the PR #1002 this could also be fixed by running the full index periodically.

andrebola:
[...]

PR #1042, opened 2017-11-21
Not index deleted packs

andrebola:
I noticed that we indexed deleted packs, it's possible (but not completely sure) that the issue #779 is because of this problem.

After this change it took more time to do the full index, @alastair maybe we need a index in the DB for the field is_deleted of the Pack table?

### DIFF
--- a/sounds/models.py
+++ b/sounds/models.py
@@ -169,7 +169,7 @@ class SoundManager(models.Manager):
         FROM
           sounds_sound sound
           LEFT JOIN auth_user ON auth_user.id = sound.user_id
-          LEFT JOIN sounds_pack ON sound.pack_id = sounds_pack.id
+          LEFT JOIN sounds_pack ON (sound.pack_id = sounds_pack.id AND sounds_pack.is_deleted = False)
           LEFT JOIN sounds_license ON sound.license_id = sounds_license.id
           LEFT JOIN geotags_geotag ON sound.geotag_id = geotags_geotag.id
         WHERE
--- a/utils/search/search_general.py
+++ b/utils/search/search_general.py
@@ -46,7 +46,7 @@ def convert_to_solr_document(sound):
     document["tag"] = getattr(sound, "tag_array")
     document["license"] = getattr(sound, "license_name")
 
-    if getattr(sound, "pack_id"):
+    if getattr(sound, "pack_name"):
         document["pack"] = remove_control_chars(getattr(sound, "pack_name"))
         document["grouping_pack"] = str(getattr(sound, "pack_id")) + "_" + remove_control_chars(getattr(sound, "pack_name"))
     else:


ffont:
So what will happen after applying this PR is that sounds whose pack has been deleted will be indexed as if they had no pack right?

andrebola:
Yes, deleted packs will be ignored.

utils/search/search_general.py
```diff
@@ -46,7 +46,7 @@ def convert_to_solr_document(sound):
     document["tag"] = getattr(sound, "tag_array")
     document["license"] = getattr(sound, "license_name")
 
-    if getattr(sound, "pack_id"):
+    if getattr(sound, "pack_name"):
```
ffont: Why this change? Is this a bug that was in here before?

sounds/models.py
```diff
@@ -169,7 +169,7 @@ def bulk_query_solr(self, sound_ids):
         FROM
           sounds_sound sound
           LEFT JOIN auth_user ON auth_user.id = sound.user_id
-          LEFT JOIN sounds_pack ON sound.pack_id = sounds_pack.id
+          LEFT JOIN sounds_pack ON (sound.pack_id = sounds_pack.id AND sounds_pack.is_deleted = False)
```
ffont: I did a quick test with a pack marked as deleted and I get the same results adding the extra `AND sounds_pack.is_deleted = False` than when I don't add it (which is `pack_id` and `pack_name` being null in the query). If I understand all of this correctly, this means that we are not really indexing deleted packs because when we delete a pack, the foreign key in sound is set to null.

However, if for some reason the foreign key is not set to null and the sound still points to pack X, then this extra clause will set the `pack_name` prop to null when `pack_id` points to a (deleted) pack. In that case we need the improvement in this PR to avoid indexing the deleted pack. Hence, this improvement fixes the problem when the data is not fully consistent in the DB, but does not look at the root problem which is to understand why we're getting sounds pointing to deleted packs (if any). In fact, I just checked now in production DB and running `Sound.objects.filter(pack__is_deleted=True).count()` returns 7 sounds, and these are 6 sounds from pack 20169 and 1 sound from pack 20836.

I'm now sure how to proceed, but I'm inclined to think that we should not merge this patch because it will potentially hide the problem of sounds pointing to deleted packs. I'd fix the ones which are wrong in DB now, and look again in this problem only if it reappears in the future.

utils/search/search_general.py
ffont: Ok I understand it now, this is because the updated query will set `pack_name` to null if pack has been deleted therefore we need to check on `pack_name` to decide whether to index the pack field or not.

andrebola: yes, that is the case

sounds/models.py
andrebola: I never though about that, I guess at some point we could remove the reference between the sound and the deleted pack, but then the question is why we keep the deleted pack?.

ffont: I don't remember why we keep it, maybe to show the "deleted pack" page. The reference between the sound and the deleted pack is supposed to be removed, that is fine.

ffont: I checked and we keep the reference to show the pack as deleted. This is fine.
As mentioned above, this fix only solves the problem in a very specific case. I'll do the following:
 * Fix current sounds pointing to deleted packs and reindex these sounds
 * Close this PR without merging
 * Check in some time from now (i.e. six months) if the problem appeared again

If the problem appears again then we further investigate the causes and (possibly) apply this PR plus some other required changes.

PR #1043, opened 2017-11-22
Add two PackDownload different classes to compare performance

andrebola:
According to issue #914, created PackDownload class and PackDownloadJson to compare the performance. Also added a command to copy all the Downloads of packs to the new models.

Before changing the code in each place that a Download of a Pack is created or accessed we should decide which of the models we will use.


ffont:
How do you want to go about testing this @andrebola?
Are you planning on doing some speed tests on local DB (like those mentioned in the issue) , or you want us to deploy this somewhere?

alastair:
I'd like to run it on fs-test with the database connected to an up-to-date mirror on fs-db1 so that we can replicated it as close as possible to production.
This week I can make a dump/mirror and make sure that test can connect to db1. Then we can deploy this branch on test.
It might make sense to have some test scripts already to check access times - as I said in chat we should check the execution times of all existing pages that we have on fs that accesses the downloads table

andrebola:
@alastair suggested to deploy it on test.freesound.org, with a replicate of the db2 in db1. I think it's a good idea. Then we can do some queries and measure the time compared to the current implementation. Also we should look at the size of the tables.

ffont:
What's the status of that? I guess we need to set up the replica of production db for test right? Is this on your list @alastair ? Should we plan it?

andrebola:
Taking the advice of @alastair I improved the script to copy the downloads. The copy script now processes around 100 downloads per second, so it will take around 7 hours to complete.

andrebola:
I wanted to migrate the downloads in the test server but there is a problem with the database and the fs is not working there.

ffont:
Was this PR finally ever deployed in test? @andrebola are you planning on deploying this in test and taking measurements or what is the current plan? If you are waiting for something on our side please let us know.

andrebola:
Yes, I will change this in test and leave the command running this (long) weekend

andrebola:
Everything related with Downloads was on pause, because there was an issue with the use of disk because of the index of the table, so I didn't move forward with this. But if you want I can do it in test anyways.

andrebola:
The command for copying pack downloads to the new models finished at test.fs. 

The queries for the model `packdownload` work faster since we don't need the join with the table `sounds_packdownloadsound` because we don't show that information anywere, but the table `sounds_packdownloadjson` is much smaller than the other two combined:

```
 table_name             |   total    |   index    |   toast    |  table  
sounds_packdownload               | 413 MB     | 284 MB     |            | 129 MB
 sounds_sound                     | 391 MB     | 184 MB     | 1112 kB    | 207 MB
 sounds_packdownloadsound         | 9521 MB    | 6672 MB    |            | 2849 MB
```

In conclusion I think it's better to keep the table `sounds_packdownloadjson`. @alastair @ffont  the data is in fs-db1 if you want to make your own tests.

alastair:
Can you add more disk usage examples? I can't see the json one, I don't know why sounds_sound is there.

I recall that we talked about how to split the packdownload table - we decided to split packdownloadsound into a separate one so that we didn't duplicate the user and created columns, right?

In the non-json table we have indexes for the foreign keys (implicitly) - sound & pack download table, but not for License: https://github.com/MTG/freesound/pull/1043/files#diff-415a01a6b052965e5200562c20c4836aR1086
Note that we have no indexes for items in the json table - we should create these also and see what this does to disk usage to be a fair comparison.

Do we have a list of queries that we currently need to run on this table, and a list of queries that we think that we might want to run? I guess we might want to show "Sounds that were included when user x downloaded this pack". Will we ever want to show on the sounds download page "this sound was downloaded by user x as part of pack y"?

Out of interest, do you think that we are able to check the size of the sounds download table now that packs are removed (compare with db2, and see if any queries are faster?)

ffont:
Current uses and potential uses of the data (this list might not be complete):

- List of downloaded packs by a user
- For a specific pack download, list of sounds that were downloaded and with which licenses
- For a given sound x and user y, list how many times it was downloaded including downloads as part of a pack and with which licenses (this is just guessing, but it would be nice to show a list of times the sound was downloaded by the visiting user. so if you're visiting a sound page and you downloaded the sound in the past, we could show that info.)

I don't see other uses right now...

alastair:
> For a given sound x and user y, list how many times it was downloaded including downloads as part of a pack and with which licenses 
This will require an index on the sound field inside the json document.

The first two queries don't need an index on the License field.

The first query - "list of packs downloaded by a user" - should be much faster than the current system with either table style.

alastair:
I'm just thinking out loud - if we have on the "packs downloaded by user" page a list of sounds, this requires a join with the two-table system, but not with the json system, although with the two table system we can join with the sounds table in the database too - with json we'll have to do 1 query for the download objects and then one query with the sound ids to get the sound objects

alastair:
the largest packs have over 1000 sounds (though the average is probably much smaller). We also need to think about how we will show this data on the page. I don't think that it make sense to say "user x downloaded these packs, containing these sounds" if one pack has 1000 sounds in it.

what is the user's goal on this page?  if we give them a list of 1000 sounds are they going to be able to find the information that they need?

What if on attribution we said "you downloaded pack x containing n sounds CC0, m sounds CC-BY" with some way of them being able to click through and get more information if they want it.
We could also be smart about showing this data based on a threshold of number of sounds (10?)

andrebola:
From what we discussed probably the best would be to keep the `sounds_packdownload` since it will be easier to do the queries using the ORM and it will the also efficient if we need to get only the downloads or only the sounds related with a single PackDownload.

ffont:
So we decided not to use the JSON version?
Would be nice to put the numbers we computed here for reference. The different queries for the things we mentioned above and how long it takes in each case.

andrebola:
User Packs downloads page (1 result) 

freesound=> select * from sounds_packdownload where user_id = YY order by created desc;
Time: 2.297 ms
freesound=> select * from sounds_packdownloadjson where user_id = YY order by created desc;
Time: 2.986 ms
freesound=> select * from sounds_download where user_id = YY AND pack_id is not null order by created desc;
Time: 68.742 ms


Pack downloads page (232 results) 
freesound=> select * from sounds_download where pack_id = XX order by created desc;
Time: 1608.748 ms
freesound=> select * from sounds_packdownloadjson where pack_id = XX order by created desc;
Time: 732.839 ms
freesound=> select * from sounds_packdownload where pack_id = XX order by created desc;
Time: 6.545 ms

ffont:
I'm not sure if I understand to above numbers correctly (I edited the format a bit to make it easier to follow):

1) user packs downloads page -> list of packs downloaded by a user? what does "1 result" mean?
2) pack downloads page -> given a pack, list the sounds that were downloaded inside this pack

if 2) is correct, then this information can't be obtained with current `sounds_download` table right? therefore I understand that the comparison is  not really appropriate with the current table. Nevertheless this is not too relevant because we are interested in the comparison between `sounds_packdownload` and `sounds_packdownload_json` right?

There is this other use case that it would be nice to know the comparison which is when we want to get all the times a sound has been downloaded (including pack downloads). Can you run queries regarding this and see how it compares when using `sounds_packdownload` and `sounds_packdownload_json`? Also I'd do the case of 'list all downloads of sound X by user Y (including inside packs)' and 'count all downloads of sound X (including inside packs)'.

andrebola:
1. 1 result means that for the query I used there was only one pack downloaded for that user.
2. I don't understand why this is not possible to obtain with current `sounds_download`, note that I didn't deleted the pack downloads from that table yet. But yes, I think it's not very relevant is just to have an idea of the current time.
 
#### Number of Downloads of a sound: 

```
select count(*) from sounds_packdownloadsound where sound_id=10005 ;
 count 
-------
  1884
(1 row)

Time: 3741.590 ms

freesound=> select count(*) from sounds_packdownloadjson where sounds @> '[{"sound_id": 10005}]' ;
 count 
-------
  1884
(1 row)

Time: 25540.329 ms
```

#### Number of Downloads of a sound from a user: 

```
freesound=> select count(*) from sounds_packdownloadjson where sounds @> '[{"sound_id": 10005}]' AND user_id = 662748;
 count 
-------
     1
(1 row)

Time: 2.592 ms

freesound=> select count(*) from sounds_packdownloadsound, sounds_packdownload where sound_id=10005 AND sounds_packdownloadsound.pack_download_id = sounds_packdownload.id  AND user_id = 662748;
 count 
-------
     1
(1 row)

Time: 51.568 ms
```

ffont:
> result means that there was only one pack downloaded for the user_id I queried.
ok

> I don't understand why this is not possible to obtain with current sounds_download, note that I didn't deleted the pack downloads from that table yet. But yes, I think it's not very relevant is just to have an idea of the current time.
I think I misunderstood what you're doing in the "pack downloads page" queries. I now understand it and it makes sense.

Looking at your last numbers it looks like for *Number of Downloads of a sound:* the `sounds_packdownloadsound` table is faster than `sounds_packdownloadjson`. Although if I understand correctly here you're only getting the number of downloads for a sound as part of a pack, not including the "normal" downloads. But anyway I guess this is not relevant here because we're only looking at the pack related things. Nevertheless, for the case of *Number of Downloads of a sound from a user:* it looks like performance is inverted. I guess this is because you to use the two tables `sounds_packdownload` and `sounds_packdownloadsound`. If included `user` in `PackDownloadSound` then that would be much faster right? 

Also this made me think that the two implementation options that you provide here are:

- use  `PackDownloadJson` model, or
- use `PackDownload` and `PackDownloadSound` models in combination

Is this right? In the description of the issue it was not contemplated that option of using two models like `PackDownload` and `PackDownloadSound` in combination, but only to use one model with the current fields in `PackDownloadSound` plus `user` (just as I suggest above, to make a query faster). Was it discussed somewhere whether it would be best to use two separated models or a single one? I imagine using two separated models will speed up the queries in *Pack downloads page* and *User Packs downloads page* and that is the reason for doing so. Any comments on that?

andrebola:
I remember discussing this on Hipchat, if you want I can also add the user_id field to  `PackDownloadSound`, and compare the time. But I guess that it will be harder to use it in that way from the ORM. 

I also think that it makes more sense to have them separated because of what you mentioned, the only time that we will do the join is for the popup.

ffont:
> I also think that it makes more sense to have them separated because of what you mentioned, the only time that we will do the join is for the popup.

What do you mean the 'popup'?

In any case I agree having the two tables separated has clear advantages. I'm just unsure whether we should have `user_id` in both tables. What would be harder to use? To get the times a sound is downloaded by a user as part of a pack Id' be just `PackDownloadSound.objects.filter(sound_id=SID, user_id=UID)` right?

andrebola:
When I say the popup refer to what we discussed yesterday on hipchat, that we will never show all  the downloads of a pack together with all the sounds of each pack. We said that we will show all the downloaded packs and then if the user selects one we will display the sounds in a Popup.

About the ORM, for that case is not difficult but for the (most common case) when you want to get the downloads of a pack for a user, you need to do a group by unless we keep both tables with the user_id as you mentioned. But I don't like having the userd_id in both tables since it will consume much more space and it will be used only in the "popup"

ffont:
Ok, now I fully understand. Then I agree `PackDownload` and `PackDownloadSound` as you implemented them is the best option. If I'm not wrong next steps would be:

- Remove `PackDownloadJson` model form this PR
- Remove/update management command (?) - we only want to prefill `PackDownload` table (not `PackDownloadSound` because we don't have info form the past)
- Merge this PR
- Continue working  in https://github.com/MTG/freesound/issues/914 (basically add the logic to fill in  `PackDownload` and `PackDownloadSound` tables) - in a new PR

One #914 is done (see the description there), we can run the command to prefill `PackDownload` and deploy changes to start storing pack downloads in the new tables.

bdejong:
Silly comment, but in this case naming it PackSoundDownload makes more sense though... Especially if you pluralise it: PackSoundDownloads makes sense, PackDownloadSounds is... weird

ffont:
> Silly comment, but in this case naming it PackSoundDownload makes more sense though... Especially if you pluralise it: PackSoundDownloads makes sense, PackDownloadSounds is... weird

Hmmm I'm not sure. Surely it pluralizes worse, but the meaning is more accurate as each row in this table will be a "sound", not a "download". So these fill be "sounds" of a "pack download".

PR #1048, opened 2017-11-24
Add sound id/pack id as extra_data in sentry instead of in title

alastair:
Otherwise we get a sentry notification for every download. This way
sentry will suppress notification emails for all downloads


ffont:
Looks great :)

PR #1050, opened 2017-11-24
Replace sentry log of downloads for graylog

andrebola:
Issue #840

Also changed logging.py in freesound-deploy to use GelfHandler which sends messages to Graylog using UDP (https://github.com/severb/graypy#usage)


ffont:
Looks good,
I'll rename `FEATURE_LOG_SOUND_DOWNLOADS` to `LOG_DOWNLOADS` and add a comment when I merge.

Issue #500, opened 2012-12-20
</> (HTML tags) and more stripped out of Messages without warning


qubodup:
Hello,

Today I wrote a message to a user that uses < and > in a non-HTML manner.
http://i.imgur.com/zuz0q.png

The result was that the content and the following content was deleted.
http://i.imgur.com/krhBG.png
1. The content after the <> symbols should not have been deleted (my opinion, quite certain)
2. The <>-tag-like content should have been warned about or not be deleted (my opinion, not certain)

Ideas for solutions on #2:
a) Whenever freesound decides to strip out tags, it should replace <> with &lt; and &gt; instead.
b) Warn the user that <> is filtered by displaying a text in the editor. Mockup:
http://i.imgur.com/VeQVz.png
c) Implement some kind of preview, which should attract attention (pop-up, big red bg warning message text)  or else people might close the window after clicking "submit" thinking the message got sent.

I hope this is a clear report (tell me if not). I will gladly think more about solutions and make more mockups if you would like me to.

Thanks for freesound!

PR #1051, opened 2017-11-24
Replaced BeautifulSoup for Bleach for cleaning html

andrebola:
The original idea was to replace everything like `<something>` for `&lt;something&gt;` but with all the libraries that I could check the problem is the same, they recognize it as a tag and breaks it. 

Solution I could find is to replaced with a regex only for the links like : `<http:/...>` for `< http:/.. >` (with a space), this way the library doesn't recognize it as a tag. For the rest of the well structured tags this PR works following the original idea, e.g: `<h1>Something</h1>` will be `&lt;h1&gt;something&lt;/h1&gt;`

Now we don't deleted the tags that are not in the allowed_list because they will not be parsed as html.

I also changed the old code which used BeautifulSoup to use Bleach. I think is much easier to understand now.

Finally I added unit tests for the clean_html method.

issue #500


andrebola:
Note that the build fails because of an unrelated problem, I have the same test failing in master branch.

requirements.txt
```diff
@@ -1,5 +1,6 @@
 BeautifulSoup==3.2.0
```
bdejong: These probably need to be removed...

utils/tests.py
```diff
@@ -224,3 +225,65 @@ def test_should_suggest_donation_probabilty_0(self):
             created=datetime.datetime.now())
         # Change downloads date again to be recent (however modal won't show because probability is 0.0)
         self.assertEqual(utils.downloads.should_suggest_donation(user, times_shown_in_last_day), False)
+
+    def test_clean_html(self):
+        ret = clean_html(u'a b c d')
```
bdejong: There are no tests for images... We do allow images right now... Neither for ul/li, ...

In particular the forum says:

> To format your text you can use the following HTML tags. a, img, strong, b, em, li, u, p, br, blockquote, code.

I know the original clean_html doesn't have tests for these but we should add these. Especially as we don't allow linking to http:// images, only to https:// images

utils/text.py
```diff
@@ -20,6 +20,9 @@
 
 import re
 import unicodedata
+import bleach
+from html5lib.filters.base import Filter
+from functools import partial
 from BeautifulSoup import BeautifulSoup, Comment
```
bdejong: shouldn't this be removed too?

utils/tests.py
andrebola: There is a test for ul, the thing is that we did not support it previously (so the text in the forum was wrong). Now I added other ul and img tests.

Note that the image tag with http or https source is allowed in this method, then is removed when we display it with a templatetag.

requirements.txt
andrebola: I think is used in other places, like messages/views.py

andrebola:
@ffont what you mention in the first bullet is what I tried to explain in my first comment https://github.com/MTG/freesound/pull/1051#issue-276663104, this is also related with your last bullet. 

I will try to explain it again. I don't think we should convert `<support>`  to `&gt;Subtitle&lt;`, as you mentioned this is how github deals with it, since it not a valid tag. As an exception to that case we support `<http://...>` only in that case I add the space with a regex because is possible to have a pattern that catch this case without leaving anything outside. And if we add this space then Bleach doesn't remove it.

ffont:
I don't understand your comment and maybe I did not express myself clearly anyway. What Github does is the same as what Freesound currently does (without this PR), they just remove the parts inside the bad tag. If we want to behave the same as Github, we could simply not merge this PR. Nevertheless, I'd still be good to add a message telling the user that unsupported tags will be stripped and that he should be careful with <> characters.

However, after discussing the issue I think we got to the conclusion that we should not remove the content and that we should either simply remove the wrong tags and leave the content inside or just leave everything but encoding the <> characters so that these are not interpreted as HTML. This is what I explain in the examples in the bullet points of my previous comment.

I understand now your comment about the URL case because this is the problem that Iwan reported in the original issue. But if instead of an URL he wrote something else inside the <> then the problem would happen again (it would be stripped). Therefore adding a solution which only works for the URL case I don't think is useful.

I don't really see this problem as something as important as deserving as much attention as we're giving it, but I think either we implement a "proper" solution that behaves the way I explained in my examples, or we just add a way through which users are warned when things are being stripped. The solution you propose in this PR is close to the "proper" solution but it still has some quirks as I described in my examples above.

I might be completely wrong. If that is the case, please show me some clear examples about how you think it should work (with input and stored/rendered output as I did in my previous comment).

andrebola:
What you propose is not possible to do with bleach, if you want to replace the tags that are not allowed it will also automatically close the tags and if you tell bleach to remove the not allowed tags then it will also remove the unclosed tags. 

The other important thing that I want to mention is that in github `<http://www.google.com>` is transformed to http://www.google.com, so I guess we will not support this either? 

I'm trying to avoid removing the tags since is difficult to display a message to the user telling him that we changed the text, it's not easy to identify the cases where the change it will be changed.

ffont:
Ok but then how do you think it should work? Show me some example because I don't understand. My example `1.` above, do you think this is correct? I think it should not add the closing tag. Also we can have a special case for URLs, I agree with that, did not realise about this in Github case.

Also the fact that Bleach does not implement the behavior I described out of the box does not mean we can't do it right? We could have a regex that gets all the tags and then me manually check if the tags are supported for us and if they are not we replace the <> for the encoded chars and then when we run Bleach to do the rest it won't touch this. Isn't that feasible?

It seems to me that I'm pushing to implement a specific behavior that I don't see reflected in the current PR, but it could be the case that it was not clear beforehand which behavior should have been implemented. Therefore before making any further change let's make sure that we agree on what we want to do. Here I'll put some examples of how I think this whole thing should behave. Please @andrebola @alastair @bdejong , feel free to answer proposing alternatives or provide your opinion so we finally decide on this. Then I'd add the agreed on examples as tests.

| # | Input | Ouput (as stored in DB) | Ouput (as rendered) |
| ------------- | ------------- | ------------- | ------------- |
| 1 | `<b>bold words</b> should be <supported>` | `<b>bold words</b> should be &lt;supported&gt;` | <b>bold words</b> should be &lt;supported&gt;  |
| 2 | `A URL in <http://inside.this> should be seen` | `A URL in &lt;http://inside.this&gt; should be seen` | A URL in &lt;http://inside.this&gt; should be seen |
| 3a | `an <script> unsuported tag </script> is stripped out, but the contents persist (?)` | `an unsuported tag is stripped out, but the contents persist (?)`  | an unsuported tag is stripped out, but the contents persist (?) |
| 3b | `an <script> unsuported tag </script> is stripped out, but the contents persist (?)` | `an &lt;script&gt; unsuported tag &lt;/script&gt; is stripped out, but the contents persist (?)`  | an &lt;script&gt; unsuported tag &lt;/script&gt; is stripped out, but the contents persist (?) |
| 4 | `<h2>an unclosed unsupported tag...` | `&lt;h2&gt;an unclosed tag...` | &lt;h2&gt;an unclosed tag... |
| 5 | `an <em> unclosed but supported tag should be stripped out?` | `an unclosed but supported tag should be stripped out?` | an unclosed but supported tag should be stripped out? |

Not really sure about cases `3a` and `3b` and `5`. I think It's just a matter to decide for one option.

andrebola:
Regarding the question in the second paragraph, I think is not a good idea to have a regex that replaces multiple tags as you propose, since you have to parse the html with a regex which is very complex, and difficult to cover all the cases. That is why we use an external library, so if the library doesn't support the behavior we want it's also a problem,. For some of the cases you listed it would be better to stick with BeautifulSoup. 

Also note that there are some cases in the current code that we need to still support, for example replacing urls for links (`<a> tags`), for the full list see: https://github.com/MTG/freesound/blob/master/utils/text.py#L141-L181

For the cases 3a and 3b, I prefer 3b. And about 4 and 5  (this also happens in 1),  I think if we keep unsupported tags which are not closed then we should also keep the supported that are not closed (but replaced).

ffont:
> Regarding the question in the second paragraph, I think is not a good idea to have a regex that replaces multiple tags as you propose, since you have to parse the html with a regex which is very complex, and difficult to cover all the cases. That is why we use an external library, so if the library doesn't support the behavior we want it's also a problem,. For some of the cases you listed it would be better to stick with BeautifulSoup.

I guess there can be different approaches. I'm proposing a two step process. First step is per-processing the text so that all `<>` from unsupported tags are replaced by `&lt; &gt;`. Then this text is fed to BeautifulSoup/Bleach to do the same processing it is doing now. I think finding the tags with regexp (or even with the BeautifulSoup/Bleach if they provide some way to return position of tags in text) should be feasible, although I don't know in depth. In any case let's first agree on which is the behavior we want and then we'll see the best way to do it. 

> Also note that there are some cases in the current code that we need to still support, for example replacing urls for links (<a> tags), for the full list see: https://github.com/MTG/freesound/blob/master/utils/text.py#L141-L181

My list of examples is not supposed to be complete, just some examples. The idea is that all supported tags should continue behaving like in these tests, and the links as well. The most important thing is to specify what to do for the tags which are not allowed. Also feel free to add more examples to the list with specific cases so it all becomes clear.

> For the cases 3a and 3b, I prefer 3b. And about 4 and 5 (this also happens in 1), I think if we keep unsupported tags which are not closed then we should also keep the supported that are not closed (but replaced).

I'm fine with option 3b. Regarding 4 and 5 what is it that you propose then? I agree to never strip tags, simply replace by `&lt; &gt;` when we don't want them to be processed.

andrebola:
What I propose is that it should works as in 4 (then 5 would be discarded). So, now that we agree on the behavior, I want to comment on how to do it, the code in this PR doesn't work like this and I don't think this will be possible with bleach version 2.+ 
I will try to do it with an older version, otherwise I will go back to BeautifulSoup

bdejong:
Hey guys, I might miss a bit of context, but... Shouldn't we be OK with changing the rules of our HTML cleaning a little? I don't necessarily see the benefit of HAVING to replace `<something>` with `&lt;something&gt;`. This use case probably happens VERY infrequently. I would think that the use case we are trying to prevent is thins like `<script>`.

So rather than trying to bend backwards to adapt the libraries we have at our disposal, why not makes the rules a bit more lenient?

For me the most important parts are:

- make sure that people don't break the website by restricting the HTML fields
- enable people to do some formatting by using html
- link-ify URLs

ffont:
Sure, that's why I'm insisting in agreeing how it should work first. Nevertheless I think this all started because of text like `<abc>dashjds dfjibs fdsjkbhsdf </abc>`, or `<df sdfk sdhsg>` or `<http://...>` would be removed with the current code as would be interpreted as HTML tags. The three points you mention are already covered with the code we had before this PR.

andrebola:
Yes we already covered this three points but now we support adding links like <http://...> and I think now the code is cleaner. 

The only big difference is the case `<something>` because this will be `<something></something>` and for what we talk it should be `&lt;something&gt;`.

ffont:
> The only big difference is the case <something> because this will be <something></something> and for what we talk it should be &lt;something&gt;.

Yeah I think that is right.

bdejong:
Imho:

- `<abc>dashjds dfjibs fdsjkbhsdf </abc>` should become `dashjds dfjibs fdsjkbhsdf`
- `<df sdfk sdhsg>` could be removed
- `<http://...>` could be removed

andrebola:
Isn't that exactly what we currently have in prod?

ffont:
The first point is not what we have in prod, it would all be removed as well.
To me if we really want to improve this, we should do what @andrebola and me discussed. Otherwise we can leave it as it is with the only little improvement of fixing the <URL> case (and also code cleanings by using Bleach). Also add a message in the help text that things between `<>` will be removed. This is an option as well.

andrebola:
> The first point is not what we have in prod, it would all be removed as well.

I think it currently works in that way, look the comment in: https://github.com/MTG/freesound/blob/master/utils/text.py#L148

For me both options are fine, I also agree this are probably not very frequently used.

ffont:
Aps, ok so then the `<unsuppored tag>XXX</unsupported tag>` is not a problem, which means that in production we have currently implemented version `3a` as I understand. Then the only "problem" we have I think is with `<XXX>` constructions. Andrés PR does:

- Support specific `<http...>` case by adding spaces around `<>`
- As Andrés explains in the description of this PR, it also replaces `<>` of unsupported (but well structured) tags for `&lt; &gt`.
- As a side effect, Bleach adds closing tags of what it considers to be a tag and is not closed (see what I described in my first comment, point 1.)

Therefore what I understand now is that this PR is fine, the only thing that should be fixed is the adding of a closing tag which is something that's going to feel really weird for users. So the two options are:

1. Fix this closing tag problem
2. Strip `<XXX>` like constructions (with the exception of `<http...>`)

Am I right? If I am right I would leave the decision to @andrebola because he's the one who has been really working with this. Just chose whatever feels best to you and makes the code clean enough. If you chose option 2., then the help text will need to be edited to include something like `Text between <> will be removed`.

ffont:
Nice, I tested it now and it behaved as expected. Merging!

Issue #1055, opened 2017-11-27
Add new fields on profile with number of Packs and Sounds downloaded


andrebola:
Add the fields: profile.num_sound_downloads / profile.num_pack_downloads
with the number of Downloads for each user.

Issue #935, opened 2017-07-18
Crazy amounts of spam in "personal description" fields


bdejong:
http://freesound.org/forum/freesound-project/40305/

maybe we should not be showing the description of people until they download at least one sounds?
these guys are all using freesound as nice spam platform...

andrebola:
I was planning to check the number of downloads before showing the description of the user but there is no reference to the the number of downloads of a user, so if I make the query it could be too heavy. 

I think we should add the field to the user and update it in the signal that counts the downloads for a sound: https://github.com/MTG/freesound/blob/master/sounds/models.py#L1087
But should we also count the downloads of packs?

bdejong:
Perhaps cross-reference with #952 ? it's somewhat related...

ffont:
Let's count both of them (packs and downloads), I think it makes sense.

andrebola:
Ok, then this issue should be done after #1055

PR #1057, opened 2017-11-27
Add fields on profile to store number of downloads

andrebola:
issue #1055 

Added the fields and updated the value in the signals, also added the check of this fields to the test of the signal. 

About the migration that fills this values, I think it could take to much time to be in a migration, maybe we can run it separately (we can check the time that takes the migration in the new replicated db!). 

Also I think we should try to schedule a maintenance downtime to deploy this changes, because 1) it will take more time to fill the  data if the database has load, 2) there will be some inconsistencies if the users continue to download sounds while we run the script or migration (I think this is less important).


ffont:
Looks good, but as you mention I'd just initialize all counts to 0 and put in the real data at a later stage with a management command. In fact, we should probably update the management command we have to sync the indexes to also do this numbers and then simply run it. It will take some time but we can better optimize the queries there. So what i propose is:

- Update this PR and remove the part of the migration which sets in the real data
- Update the `report_count_statuses` command to include these new fields (this I can do myself as I've been working with this command lately). I'd create a new issue for this and do it in a separate PR so this one can be merged.
- Then we should find in which places we could start using these new count fields and use them.

What do you think @andrebola @alastair ?

andrebola:
I think you're right. For me is better to update the data with a command if we are not using this field yet.

ffont:
Ok, let's do it this way then!
But I won't include this PR in the current deploy anyway, leave it for next week.
You can update the PR removing that part of the code.

alastair:
Even if we do the update of the count in a background process, consider optimising it. For example, you're doing 23 million sql queries in the update (select count downloads, select count packs, save). You can optimise this to do a single query for counts of sounds, another query for counts of packs, and only 6.5 million updates since the other 1.3 million users haven't actually downloaded anything.

accounts/migrations/0015_num_downloads.py
```diff
@@ -0,0 +1,26 @@
...
+def forwards_func(apps, schema_editor):
+    # We get the model from the versioned app registry;
+    # if we directly import it, it'll be the wrong version
+    Download = apps.get_model("sounds", "Download")
+    Profile = apps.get_model("accounts", "Profile")
+    db_alias = schema_editor.connection.alias
+    for profile in Profile.objects.all():
+        profile.num_sound_downloads = Download.objects.filter(user_id=profile.user_id, pack_id__isnull=True).count()
+        profile.num_pack_downloads = Download.objects.filter(user_id=profile.user_id, sound_id__isnull=True).count()
+        pack.save()
```
alastair: pack? profile? Did you test this migration?

andrebola:
Now it performs 3 queries (including the update) for each user. I think it will take around 24 hours to update all the users.

accounts/migrations/0015_num_downloads.py
andrebola: I tested the migration but probably I made some change before I committed, it was hard to fully test it because of the size of the download table.

andrebola:
I think it will be safer to do this changes is a maintenance window, now with this query takes less than 10 minutes.

andrebola:
Also note that now is important to set the default value to 0 after the command is executed otherwise there will be null values in this fields.

andrebola:
When I try to add the default value and not allow nulls it gives an error because of a bug: https://code.djangoproject.com/ticket/25105#comment:4

So I think the best is to keep the default=0 in the migration since is not much the difference of time and it will take the same time to replace the nulls for 0 values.

alastair:
As the `update set = 0 where is null` takes ~as long as the initial add column with default takes, let's just set the default at column creation time and have a slightly longer downtime. We aren't saving any time trying to be smart with the not null.

ffont:
I'm assigning this to @alastair so that we can try one more time in test before deploying to production.

alastair:
on it... waiting for disk increase in db servers to go ahead with it

ffont:
Hi @alastair, any plans for this? Now that we got space ;)

alastair:
OK, I'll run this on test to get an estimate of the downtime.

We also need to reboot production database server in order to finish disk switchover stuff. My plan is to do the reboot while we release this patch in production.

alastair:
I tested this on fs-db and it takes 4m46.962s to apply the migration and 20m31.458s to update counts for downloads and packs. We only refer to the fields in the django signals, so if we wanted to we could probably only have downtime during the migration and then run the updater later. Keep in mind that because the updater runs in just 2 queries it will probably have an impact in FS performace (compared to a less efficient updater that lets us sleep between queries).

We should upgrade the count check script to also check these numbers (https://github.com/MTG/freesound/issues/1113)

Have we decided where to use these fields yet? From a quick look around FS I can't find any places yet where we should use it - the public profile doesn't say "this user has downloaded n sounds"

alastair:
When we have downtime to apply the migration we need to reboot the db server to update the disk configuration after we increased its size.
This will cause slowdowns in freesound while we reload caches. I found some documentation at https://www.postgresql.org/docs/current/static/pgprewarm.html which might help us, but we need to know what the popular tables are. Perhaps we could do it for users, profiles, sounds, and downloads.

andrebola:
The idea was to use this fields to check if the user is spam or not: #1055

alastair:
Great, but that's a new feature, right? There are no existing places that we should update?

ffont:
@andrebola is right, #935 is the first issue for which we plan to use that. For the rest I'd be nice to estimate the impact of the 20m queries. Nevertheless, if we're going to have lower performance anyway because of the warm-up, we could tell freesound users that we're expecting 5 minutes total downtime and 30 minutes potential slowness and I think that will be fine. We could program it for tomorrow morning @alastair if that's ok for you, and make a blog post announcing the downtime.

ffont:
> Great, but that's a new feature, right? There are no existing places that we should update?

right, no existing places to update

alastair:
@ffont Sure, let's do it tomorrow morning

Issue #1047, opened 2017-11-23
Synchronise donations with Paypal and Stripe


ffont:
Currently when a donation is carried we are getting callbacks from Paypal and Stride which we use to create donation objects and store the data. Nevertheless when looking at the data in Paypal it seems we are not correctly synchronized (not all donations that appear in Paypal also appear in Freesound). I presume the same problem happens for Stripe, but it might not be the case. This might be because of people going directly to the Paypal page to make the donations instead of using our form. If that is the case, then maybe it is not a problem for the Stripe case (as there is no Stripe page to go and donate to Freesound). Nevertheless we should make sure that we create donation objects for all donations.

What I propose to do is to write a management command which runs daily and uses the Paypal/Stripe API to get a list of donations and create the objects (and send messages to Graylog) if these are not in Freesound. In this way we can be perfectly in sync.

One issue to look into is that if we do that, the timestamp for the messages we send to Graylog will be that of the command running and not the donation itself. We should see if there is a way to force a particular timestamp (I guess not). Otherwise we should add a new timestamp field to all donations send to Graylog.

PR #1058, opened 2017-11-28
Add management command to synchronize old paypal donations

andrebola:
issue #1047

Make sure the settings are placed in the local_settings file, the value on PAYPAL_USERNAME is also given by paypal (is not the regular username):

https://developer.paypal.com/developer/accounts/

Regarding Stripe, I don't  see how we could have donations not registered, since is our server that requests the donations. Maybe we can check first the amounts donated vs donations records before implementing it.

### DIFF
--- a/donations/management/commands/check_missing_donations.py
+++ b/donations/management/commands/check_missing_donations.py
@@ -1,0 +1,94 @@
+#
+# Freesound is (c) MUSIC TECHNOLOGY GROUP, UNIVERSITAT POMPEU FABRA
+#
+# Freesound is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Freesound is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#     See AUTHORS file.
+#
+
+import datetime
+import logging
+import requests
+import json
+from urlparse import parse_qs
+from django.conf import settings
+from donations.models import Donation, DonationCampaign
+from django.core.management.base import BaseCommand
+
+logger = logging.getLogger("web")
+
+
+class Command(BaseCommand):
+    help = 'Synchronize Paypal donations'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '-d', '--days',
+            action='store_true',
+            dest='days',
+            default=1,
+            help='Use this option to get the donations older than 1 day.')
+
+    def handle(self, **options):
+        days = options['days']
+        logger.info("Synchronizing donations from paypal")
+
+        td = datetime.timedelta(days=days)
+        start = datetime.datetime.now() - td
+
+        one_day = datetime.timedelta(days=1)
+        stop_date = datetime.datetime.now() - one_day
+        params = {
+            'METHOD': 'TransactionSearch',
+            'STARTDATE': start,
+            'ENDDATE': start + one_day,
+            'VERSION': 94,
+            'USER' : settings.PAYPAL_USERNAME,
+            'PWD': settings.PAYPAL_PASSWORD,
+            'SIGNATURE': settings.PAYPAL_SIGNATURE
+        }
+        while stop_date >= start:
+            req = requests.post(settings.PAYPAL_PAYMENTS_API_URL, data=params)
+            raw_rsp = parse_qs(req.text)
+            if raw_rsp['ACK'] == ['Success']:
+                campaign = DonationCampaign.objects.order_by('date_start').last()
+                Donation._meta.get_field('created').auto_now_add = False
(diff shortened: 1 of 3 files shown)


donations/management/commands/check_missing_donations.py
```diff
@@ -0,0 +1,94 @@
...
+            'PWD': settings.PAYPAL_PASSWORD,
+            'SIGNATURE': settings.PAYPAL_SIGNATURE
+        }
+        while stop_date >= start:
+            req = requests.post(settings.PAYPAL_PAYMENTS_API_URL, data=params)
+            raw_rsp = parse_qs(req.text)
+            if raw_rsp['ACK'] == ['Success']:
+                campaign = DonationCampaign.objects.order_by('date_start').last()
+                Donation._meta.get_field('created').auto_now_add = False
+                for i in range(100):
```
ffont: Why this 100 here?

ffont:
Nice :)
I'll look at the stripe case when I review.
Did you checked if it is possible to force the timestamp in the graylog message? Probably not but just to check. Also if it's not possible it will be good to have "created" field also in the donation data. I see you add it here, but is it also added for the normal donations?

I haven't look at the code carefully yet, but also if there is any code that could be reused to store/log the normal donations and these ones it would be nice to create some util functions (not sure if this makes sense now, but just mentioning it).

donations/management/commands/check_missing_donations.py
andrebola: Because they return maximum 100 items. And I think we are not going to have more than 100 donations per day :)

andrebola:
I think we need to add the 'created' value in the current place where we create the donations, do you want me to do it in this PR?

donations/management/commands/check_missing_donations.py
bdejong: @andrebola don't they return the actual number of results? Then you don't have to do this ugly hack of

`if ('L_EMAIL%d' % i) in raw_rsp:`

andrebola: The response has the form: 
`L_TRANSACTIONID0=9XJ88717E86951234&L_TRANSACTIONID1=34A53358N84361234` . You can check an example here:
https://developer.paypal.com/docs/classic/express-checkout/ht_searchRetrieveTransactionData-curl-etc/

Maybe someone has a better way of doing it?

bdejong: Dear god have mercy that's messy. No support for JSON or something more.... sane?! I guess not... I mean I would say that the more sane way is perhaps to filter out the right keys first and then iterate over those? Something like.....

```python
filtered_keys = (key for raw_rsp.keys() if key.starts_with("L_EMAIL"))
for email addres in filtered_keys:
  ...
```

perhaos?

BTW, the docs say: "This integration method is Deprecated as of January 1, 2017"

bdejong: ( https://developer.paypal.com/docs/api/payments/#payment_list perhaps? )

andrebola: yeah, but I couldn't find a way of doing it with the REST API. So I guess they will maintain it for a while. 
We should remove Paypal and leave only Stripe!

andrebola: the payments API returns no results, I think it only works if the payment was created using the API.

ffont: so what's the conclusion about this? we either use a deprecated method or have no other option?

ffont:
> I think we need to add the 'created' value in the current place where we create the donations, do you want me to do it in this PR?

Yes let's do it here

donations/management/commands/check_missing_donations.py
ffont: Can you clarify on this? Don't we have a stable way to get all the payments? Also regarding the maximum of 100, what happens if there are more? I think it would still be nice to retroactively create donation objects since we deployed new donations things. Therefore the first run of this command would be with a high `--days` number. There will certainly be more that 100. What will happen? Also for this first run I'd disable logging to graylog as messages will be misplaced. Once we run this command daily it's not too bad if we have some donation messages that appear in a different hour, but for the first run we would be adding a lot of donations at the time the command is run. Maybe we could add an option to not send to graylog?

ffont: @andrebola just pinging you because this comment is buried in the thread and I'm not sure whether giuthub will notify :)

freesound/local_settings.example.py
```diff
@@ -50,6 +50,10 @@
 STRIPE_KEY = "sk_test_..."
 PAYPAL_EMAIL = "paypal@freesound.org"
 PAYPAL_VALIDATION_URL = "https://www.sandbox.paypal.com/cgi-bin/webscr"
+PAYPAL_PAYMENTS_API_URL = "https://api-3t.sandbox.paypal.com/nvp"
+PAYPAL_PASSWORD = ""
```
ffont: Do we really have to use our paypal username and password? or isn't there any other token that we can generate to access the API?

andrebola: This are not the regular password and username, this are values given by paypal

ffont: ok, where do we get these values?

andrebola: You need to go to the administration page in:

https://developer.paypal.com/developer/accounts/

then, in the left menu click on 'Accounts', after that click on the user and when a menu is displayed click on 'Profile' . Finally in the tab API Credentials you will find this info under 'Classic Sandbox API Credentials'.

Issue #706, opened 2016-02-04
Make links work if user changed name


qubodup:
If a user changed their name by asking an admin to do so and the admin approved, all links break - to my knowledge.

For example if qubodup used to be called qqqubodup, then after the name change, sound id 331381 would be reachable at
http://freesound.org/people/qubodup/sounds/331381/ but not at
http://freesound.org/people/qqqubodup/sounds/331381/

However all the needed information is available at the old link. The sound ID is unique and the user name is not needed to find the respective sound.

I suggest that Freesound's "page not found" page analyzes the URL and if it has a structure like the following pseudeo-regex:

```
http(s)?://(www.)?freesound.org/people/[^/]*?/sounds/NUMBER(/)?
```

Then it redirects\* to the sound that has NUMBER as an ID - if such a sound exists.

*it should redirect, not provide duplicate content at the wrong URL.

alastair:
One problem with allowing any text to be added as an account name means that people could maliciously put someone else's name in place in the url.
A way around this would be to keep a list of old usernames for each user. As username changes are manual and don't change too often, this shouldn't be too much of a problem with regards to speed or storage.

alastair:
Cool URIs don't change: https://www.w3.org/Provider/Style/URI.html

qubodup:
> One problem with allowing any text to be added as an account name means that people could maliciously put someone else's name in place in the url.

Clarification, just in case:

https://freesound.org/people/qubodup/sounds/177182/ would not deliver the content of https://freesound.org/people/johnhanes/sounds/177182/ while https://freesound.org/people/qubodup/sounds/177182/ is in the navbar.

It would **redirect** to https://freesound.org/people/johnhanes/sounds/177182/ . The URL in the navbar would be https://freesound.org/people/johnhanes/sounds/177182/ .

alastair:
Yes, agreed. This is the way I would do it too. However it would still be possible to use an incorrect citation, e.g. 

 > Uses c8philter4.aif by alastairp, from https://freesound.org/people/alastairp/sounds/177182/

(not everyone looks at the URL when they load a page. I would guess that there will also be people who don't notice that the url of the link and the uploader name is different)

which is what I want to reduce the possibility of. By keeping old usernames for a user we can accept any of their usernames, and none others, which I think is safer.

ffont:
[...]

PR #1059, opened 2017-11-29
Make links work if user changed name

andrebola:
issue #706 

Added new model to store the previous usernames of a user, then check on sounds page if the username is an old username and make a redirect. Also added some tests.


accounts/models.py
```diff
@@ -411,3 +411,11 @@ def __unicode__(self):
 class UserEmailSetting(models.Model):
     user = models.ForeignKey(User, related_name="email_settings")
     email_type = models.ForeignKey(EmailPreferenceType)
+
+
+class OldUsername(models.Model):
+    user = models.ForeignKey(User, related_name="old_usernames")
+    username = models.CharField(max_length=255)
```
bdejong: this probably needs an index...

accounts/models.py
```diff
@@ -411,3 +411,11 @@ def __unicode__(self):
 class UserEmailSetting(models.Model):
     user = models.ForeignKey(User, related_name="email_settings")
     email_type = models.ForeignKey(EmailPreferenceType)
+
+
+class OldUsername(models.Model):
```
bdejong: Shouldn't we be auto-filling these if the username gets changed - for users who have sounds this only happens through the admin, btw?

andrebola: yes, we only change this using the admin. I added the auto-filling :+1:

accounts/admin.py
```diff
@@ -188,7 +188,24 @@ def delete_preserve_sounds(self, request, obj):
 
     change_actions = ('full_delete', 'delete_include_sounds', 'delete_preserve_sounds', )
 
+    def save_model(self, request, obj, form, change):
```
ffont: Wouldn't it be better to do this in a `User` `pre_save` signal so that we make sure that it will be ok even if we change a username through the console (which could happen)?

andrebola: I thought that if we only were going to change this in the admin then it didn't make sense every time we do a change in a user object to check if the username changes. But after this a new issue was created regarding allowing the user to change the username so now it might make sense to do it in a pre save.

ffont: Well, given that we almost never change user objects (only profiles) then I think we're fine.

accounts/migrations/0015_auto_20171129_1537.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('accounts', '0014_oldusername'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='oldusername',
+            name='username',
+            field=models.CharField(db_index=True, max_length=255),
```
alastair: because we check usernames with `__iexact` we need a case-insensitive index here: 
```
    "auth_user_username_upper_key" UNIQUE, btree (upper(username::text))
    "auth_user_username_gist_trgm_idx" gist (upper(username::text) gist_trgm_ops)
```
I can't remember why we have both a gist and a btree index, we should look into that and remove the one that we don't need.

sounds/views.py
```diff
@@ -209,6 +209,8 @@ def sound(request, username, sound_id):
     try:
         sound = Sound.objects.select_related("license", "user", "user__profile", "pack").get(id=sound_id)
         if sound.user.username.lower() != username.lower():
```
alastair: This check is done 8 other times in this file for various URLs that contain both the sound id and username. We should do the same old username check in those places (preferably in some helper function so that we don't duplicate code)

ffont:
I'm reopening the original issue as changes are needed.

Issue #1061, opened 2017-11-30
Link to "moderation pending" sounds from user account


bdejong:
As admin it would be super useful to be able to jump to the "moderation pending" sounds of a user straight from the user profile. This would speed up certain tasks like users saying: "hey, where are my sounds?!?"

PR #1063, opened 2017-11-30
Add link to 'moderation pending' sounds in user profile

ffont:
https://github.com/MTG/freesound/issues/1061


bdejong:
I did not expect a one-liner fix :-D 👍

Issue #1066, opened 2017-12-01
Speedup some Django admin pages


ffont:
Django admin pages are super slow, specially for the tables with high number of entries (e.g. Sounds, Users). As an admin, I want a fast server response when browsing Sounds and Users admin pages, including list view and individual views.

The list of pages that should be optimized:

- Sounds list (https://freesound.org/admin/sounds/sound)
- Individual sound page (e.g. https://freesound.org/admin/sounds/sound/412241/change/)
- Users list  (https://freesound.org/admin/auth/user/)
- Individual user page (e.g. https://freesound.org/admin/auth/user/10/change/)

@andrebola already suggested that the individual user page can be improved by removing the permissions section. We can remove that, but then we need to add an action to turn users into moderators (and possibly to remove moderator permissions as well).

The individual sound page is probably already fast so there is not much problem with it. The ones needing more love are probably the list ones.

andrebola:
Regarding users list, the problem is the pagination (the query that counts all the users), this could be changed by using a custom paginator, but also there will be a better solution in newer versions of django:
https://code.djangoproject.com/ticket/8408

About the list of sounds, if I remove the columns user and license then the page loads very fast.

andrebola:
Note that for user form I only removed the individual permissions but the group permissions widget is still there

ffont:
> Note that for user form I only removed the individual permissions but the group permissions widget is still there

Cool, so then we don't need the make moderator buttons. Where was this change made?

Regarding the count problem I think we can't count yet on the future solution because we don't know when it will happen. Let's have a custom paginator for the admin and add a comment which mentions this Django issue and add a note to our "upgrading django version" wiki page to look for this when we upload. Let's use the paginator both in the users list view and the sounds list view. For the sounds case, why removing this columns makes it much faster? Are we missing select related somewhere? Maybe it can be added.

andrebola:
here: https://github.com/MTG/freesound/pull/1069/files

andrebola:
I think the slow part on sounds list is the join in the query with the table of license and users, so it's not a problem of select related because is one single query that takes too much time (maybe a index is missing?)

andrebola:
We decided to remove the permissions widget from the admin, but we have to make sure we have groups for each permission. For this @ffont made this list of action to take:

- Check which special permissions exist and to which users these are assigned (like andres did). write this information in a doc somewhere for the record
- Check if these permission are actually used in code (search for permissions name in code), and remove those which are not used 
[...]

PR #1069, opened 2017-12-01
Speedup admin page

andrebola:
issue #1066


bdejong:
While I think that -in theory- this is a good idea, I don't like us removing the persmissions! It means we disable admins (me!) to edit user permissions, so we reduce a problem (waiting) and introduce a new blocking one. As said in the chat, I think the speed on freesound.org itself is fine...!

andrebola:
I found that the users page in the admin is slow in my local, but it works fine in freesound.org. 
Do we change specific permissions of the users? note that this doesn't affect managing group permission for users.

PR #934, opened 2017-07-18
[WIP] Docker setup

alastair:
Work in progress to get docker working

To do:

 - [x] Web
 - [ ] Processing
 - [ ] Migrate initial database
 - [ ] Search
 - [ ] Similarity
 - [ ] Nginx (zip/packs)
 - [ ] Mail
 - [ ] Logging
 - [ ] Get everything running in docker-compose
 - [ ] Swarm?
 - [ ] HTTPS for development?
 - [ ] Logging

PR #1070, opened 2017-12-03
Initial front-end setup

giuband:
Just setting up a state of art FE environment for our dear Freesound. 😜 
Run `npm install` and `npm run styleguide` to see the docs.


freesound/static/webpack/common.js
```diff
@@ -0,0 +1,56 @@
+const path = require('path');
+
+const rootPath = path.resolve('.');
+
+const SUPPORTED_BROWSERS = [
```
giuband: This is probably the most important line in the entire PR @ffont @alastair

giuband: It sets up both css autoprefixing and js transpiling according to this setting

package.json
```diff
@@ -1,54 +1,54 @@
...
+    "eslint-config-prettier": "^2.9.0",
+    "eslint-plugin-import": "^2.8.0",
+    "extract-text-webpack-plugin": "^3.0.2",
+    "html-webpack-plugin": "^2.30.1",
+    "node-sass": "^4.7.2",
+    "postcss-flexbugs-fixes": "^3.2.0",
+    "postcss-loader": "^2.0.9",
+    "prettier": "^1.8.2",
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0",
```
giuband: `react` and `react-dom` have been added only to have the `styleguide` correctly running

freesound/static/webpack/common.js
ffont: I think we said we can drop IE < 10 (or even < 11) so I think this list is fine. In any case if this thing lets us know when things won't be supported in non-supported browsers (like with a warning or something), then we'll be better able to decide whether we want to have a fallback implementation of something or just leave it totally unsupported. For example, maybe we have a super basic version of the player for old browsers so these can still be used for basic navigation and for playing sounds.

ffont:
The style guide looks awesome :)
@giuband Do you want me merge this PR?
If you plan on continue working on it before "requesting" a merge, you can add the *[WIP]* prefix to the title (see https://github.com/MTG/freesound/pull/934 for example).

giuband:
I would say let's merge it for now. I'll keep opening more and more for each chunk of work.

PR #1071, opened 2017-12-04
Only log download if 'range' not in headers

andrebola:
As discussed in https://github.com/MTG/freesound/issues/840#issuecomment-348940421 
Here I only add a condition to log the Download of Sound and Pack. Now I only log it if the range header is not present in the request.

issue #840


sounds/views.py
```diff
@@ -317,15 +317,16 @@ def sound_download(request, username, sound_id):
...
-        })
+        if 'HTTP_RANGE' not in request.META:
+            downloads_logger.info('Download sound', extra={
+                'user_id': request.user.id,
+                'user_ip': request.META.get('HTTP_X_FORWARDED_FOR'),
+                'protocol': request.META.get('HTTP_X_FORWARDED_PROTOCOL'),
+                'session_id': request.session.session_key,
+                'user_agent': request.META.get('HTTP_USER_AGENT'),
+                'sound_id': sound_id,
+                'range': request.META.get('HTTP_RANGE', None),
```
alastair: It doesn't make sense to record this value if you are checking in the `if` block that `HTTP_RANGE` doesn't exist.

I think this check should also include if the range starts with 0, because we're not sure if it's happening on only GET requests, or if we have a HEAD to get the size of the file too.

andrebola:
I updated this PR to log only downloads without range header or downloads with range=0. After this is merged we should enable logging for some hours to check:
 - If we have duplicate downloads for the same user, sound, session (?)
 - (If the duplicate downloads) then check the relation with the method from the request (GET/HEAD/etc)

ffont:
Nice, I merged the PR and will turn on logging on Monday to collect data until Tuesday. I'll let you know when I do these things.

PR #1072, opened 2017-12-05
Atomic transaction added to multiple views

andrebola:
According to #967 we want to prevent code to run multiple times but also
to prevent inconsistencies because of uncompleted executions.

Here we add transaction.atomic decorator to multiple views, depending on
whether they modify to database or not.

Together with this changes PR #1031 should be enough to solve all the issues related with double-clicks

### DIFF
--- a/accounts/views.py
+++ b/accounts/views.py
@@ -104,6 +104,7 @@ def login(request, template_name, authentication_form):
 
 
 @login_required
+@transaction.atomic()
 def multi_email_cleanup(request):
 
     # If user does not have shared email problems, then it should have not visited this page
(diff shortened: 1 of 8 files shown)


ffont:
Thanks @andrebola for finding all these views ;)
A couple of comments:
- Is there a way in which we only make an atomic transaction if the view does a POST (for views that operate different depending on GET/POST).
- I'll go through all the views and maybe propose to remove the decorator to some of them if I think it is not needed for some reason (I'm not sure if I'll find any).

andrebola:
We could use the syntax: `with transaction.atomic(): ` only when the request is POST. I think the code is more understandable with the decorator, but also might be more efficient in this other way.

ffont:
It's complicated because code for GET and POST is typically mixed. Maybe we'll just keep it as it is know. I'll give it a through look next week.

Issue #1073, opened 2017-12-09
User # of comments mixed up with # ratings


bdejong:
https://freesound.org/forum/bug-reports-errors-and-feature-requests/40685/

the small views are currently using the number of ratings to display the number of comments

ffont:
Related to https://github.com/MTG/freesound/pull/1056
For some reason `Sound.num_comments` is set to `Sound.num_ratings`

PR #1080, opened 2017-12-11
Fix problems with wrong num comments/ratings

ffont:
Issue https://github.com/MTG/freesound/issues/1073

Summary: The `report_count_statuses` command has a problem as using Django's `annotate` functionality when joining over multiple tables returns "wrong" count results. This PR fixes this problem and adds unit tests for the command.

### DIFF
--- a/general/management/commands/report_count_statuses.py
+++ b/general/management/commands/report_count_statuses.py
@@ -21,6 +21,7 @@
 
 import logging
 import pprint
+from collections import defaultdict
 
 from django.contrib.auth.models import User
 from django.core.management.base import BaseCommand
@@ -58,27 +59,8 @@ class Command(BaseCommand):
             if count % 10000 == 0:
                 console_logger.info(message % (total, 100 * float(count + 1) / total))
 
-        # Iterate over all sounds to check: num_comments, num_downloads, avg_rating, num_ratings
-        # While iterating, we keep a list of user ids and pack ids for then iterating over them
-
-        mismatches_report = {
-            'Sound.num_comments': 0,
-            'Sound.num_downloads': 0,
-            'Sound.num_ratings': 0,
-            'Pack.num_sounds': 0,
-            'Pack.num_downloads': 0,
-            'User.num_sounds': 0,
-            'User.num_posts': 0,
-        }
-        mismatches_object_ids = {
-            'Sound.num_comments': list(),
-            'Sound.num_downloads': list(),
-            'Sound.num_ratings': list(),
-            'Pack.num_sounds': list(),
-            'Pack.num_downloads': list(),
-            'User.num_sounds': list(),
-            'User.num_posts': list(),
-        }
+        mismatches_report = defaultdict(int)
+        mismatches_object_ids = defaultdict(list)
 
         # IMPLEMENTATION NOTE: on the code below we iterate multiple times on Sounds, Packs and Users tables.
         # This is done in this way because due to our DB structure and the way that Django ORM works, if we annotate
@@ -205,5 +187,4 @@ class Command(BaseCommand):
 
         console_logger.info("Number of mismatched counts: ")
         console_logger.info('\n' + pprint.pformat(mismatches_report))
-        mismatches_object_ids = {key: value for key, value in mismatches_object_ids.items() if value}
         console_logger.info('\n' + pprint.pformat(mismatches_object_ids))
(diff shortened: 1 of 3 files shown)


general/management/commands/report_count_statuses.py
```diff
@@ -74,153 +80,129 @@ def handle(self,  *args, **options):
...
-                if real_num_downloads != sound.num_downloads:
-                    mismatches_report['Sound.num_downloads'] += 1
-                    mismatches_object_ids['Sound.num_downloads'].append(sound.id)
-                    sound.num_downloads = real_num_downloads
-                    needs_save = True
-
-            # Check num_ratings and avg_rating
+                if not options['no-changes']:
+                    sound.is_index_dirty = True
+                    sound.save()
```
alastair: This is potentially going to be slow (especially the first time you run it again to fix all the counts). Can we do a bulk save (wrap it in a transaction)?

general/tests.py
```diff
@@ -0,0 +1,124 @@
...
+#
+
+from forum.models import Thread, Post, Forum
+from ratings.models import SoundRating
+from sounds.tests import create_user_and_sounds
+from django.core.management import call_command
+from django.test import TestCase
+
+
+class GeneralManagementCommandTestCase(TestCase):
```
alastair: This could be a specific test class for the report count statuses command

general/management/commands/report_count_statuses.py
```diff
@@ -74,153 +80,130 @@ def handle(self,  *args, **options):
...
-            annotations['real_num_downloads'] = Count('downloads')
-
-        for count, sound in enumerate(Sound.objects.all().annotate(**annotations).iterator()):
-
-            needs_save = False
 
-            # Check num_comments
+        # Look at number of comments
+        for count, sound in enumerate(Sound.objects.all().annotate(real_num_comments=Count('comments')).iterator()):
             real_num_comments = sound.real_num_comments
```
bdejong: The only thing which annoys me thoroughly about this whole file is the usage of the "real_" prefix which isn't needed in any of the cases. The whole thing would be so much more readable without that prefix. If you find-replace "real_" with nothing I think this all still works just fine and is a lot more readable.

general/management/commands/report_count_statuses.py
```diff
@@ -47,13 +49,17 @@ def add_arguments(self, parser):
...
+        def report_progress(message, total, count):
+            if count % 10000 == 0:
+                console_logger.info(message % (total, 100 * float(count + 1) / total))
+
         # Iterate over all sounds to check: num_comments, num_downloads, avg_rating, num_ratings
         # While iterating, we keep a list of user ids and pack ids for then iterating over them
-        all_user_ids = set()
 
         mismatches_report = {
             'Sound.num_comments': 0,
```
bdejong: mismatches_report and mismatches_object_ids should be converted to a defaultdict to make this MUCH nicer and less maintainance.

```python
from collections import defaultdict
mismatches_report = defaultdict(int)
mismatches_object_ids = defaultdict(list)
```

general/management/commands/report_count_statuses.py
```diff
@@ -74,153 +80,130 @@ def handle(self,  *args, **options):
...
-        pprint(mismatches_report)
-        mismatches_object_ids = {key:value for key, value in mismatches_object_ids.items() if value}
-        pprint(mismatches_object_ids)
+                if not options['no-changes']:
+                    user_profile.save()
+            report_progress('Checking number of posts in %i users... %.2f%%', total, count)
+
+        console_logger.info("Number of mismatched counts: ")
+        console_logger.info('\n' + pprint.pformat(mismatches_report))
+        mismatches_object_ids = {key: value for key, value in mismatches_object_ids.items() if value}
```
bdejong: If you use a defaultdict this list comprehension is no longer needed!

general/management/commands/report_count_statuses.py
ffont: The objects I'm updating already have a property with the same name without the prefix `real_`. The point here is that I'm fixing those which are wrong, assuming that there will be mismatches in the DB. Therefore to me it makes sense to add the `real_` prefix as these are the counts which have been computed "right now" and that have the correct value. Also to me it makes it readable because "if real number of apples is not equal to number of apples then X". Also I think that removing the `real_` prefix will break the code as then the annotated property will collide with the actual property. 

TL/DR: I won't change it :)

general/management/commands/report_count_statuses.py
ffont: Awesome, fixed it https://github.com/MTG/freesound/pull/1080/commits/e3e54f3b1c4eb3c77c9979266ead374d19a98679
I always forget about these super nice tricks!

Issue #1062, opened 2017-11-30
Gather average moderator stats for last 6 months


bdejong:
As admin I would like to see which moderators moderated in the last 6 months and how many sounds they moderated so I can clean up the pool of moderators.

PR #1084, opened 2017-12-12
Add page with moderation stats for last 6 months

andrebola:
issue #1062

### DIFF
--- a/monitor/urls.py
+++ b/monitor/urls.py
@@ -27,6 +27,7 @@ urlpatterns = [
 
     url(r'^processing/process_sounds/$', monitor.views.process_sounds, name='monitor-processing-process'),
     url(r'^stats/$', monitor.views.monitor_stats, name='monitor-stats'),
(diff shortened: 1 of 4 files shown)


andrebola:
Ideally we should use the modification date from tickets to filter the last 6 months, but this field is automatically updated if you apply any change to the ticket, so (I think) it could happend that in a migration we updated all the tickets so they could have the same modified_date  

https://github.com/MTG/freesound/blob/701bbfe26d4dec98ffbd4c517eef5168d728f193/monitor/views.py#L113-L116

The alternative is to use the created value of the ticket which is not exactly what we want here but probably it will give the same results because we are using a big time slot.

andrebola:
After looking the data in fs I decided to use create date because some users had days where they moderated  thousands of sounds and it didn't look right to me, I think now it is.

The last doubt that I have is to display the moderators that doesn't have any moderation in the period, currently I'm not displaying them.

Issue #829, opened 2017-03-29
Refactor csv_bulk_upload command


ffont:
The command we have to upload sounds in bulk from csv descriptions should me improved in a number of ways:

- Validate data input before starting to upload any sound. Check that all lines in the csv can be read properly, that the sounds exist in the path, that the sounds do no exist in freesound etc.
- Before starting to upload, show the user some stats after the validation process or throw errors if upload can not start. For example: "X sounds will be uploaded, Y will be skipped because of, bla bla. Do you want to proceed?"
- Add some options to the script to set a base bath for the sounds different that that of the csv data, delete sounds if they already exist/update descriptions if sounds already exist/do nothing if sounds already exist, ...

bdejong:
If these improvements are added, could we not add this as a standard FS feature, perhaps through a URL which you only can visit if you have a certain permissions set?

ffont:
Yes, I agree!
But we should report abut upload progress, etc.
Should run as an async job that reports back regularly by storing in db or something. We've been doing similar things recently in related projects and there we use a setup based on django+celery+redis which works quite well, but in Freesound I guess we should stick to the technology stack we have.

ffont:
In addition to what @alastair  did, I'd also add the option to use a `base_dir` for sounds different than that of the csv file. This would come handy in a typical use case in which users upload the sounds using the web interface (so these are stored in `settings.UPLOADS_PATH`) and then send us the csv file separately (or upload them to fs in the future ;)). Then we could run the script, point to the csv file and set the sounds `base_dir` to `settings.UPLOADS_PATH`.

In that case we can skip the first copying of files to `settings.UPLOADS_PATH` (https://github.com/MTG/freesound/blob/f706449072a14099f225ea1e09833daca6596855/sounds/management/commands/csv_bulk_upload.py#L118), or simply check if file exists before copying.

Also what this means is that sounds will be deleted from `settings.UPLOADS_PATH` after the script runs, which in my opinion is ok except for the case in which the script is re-run and then all the sounds that were successfully uploaded in the first run will now fail because "sound does not exist". I don't see an easy way for the script to know that the sound is already in Freesound besides searching for a sound of the same username with the same name etc, although this will still be approximate as names are not unique. Best option would be to use md5 but we can't because file is no longer there. Well, maybe we can simply do a query based on username and soundname and then if it returns something and we see that the corresponding sound is not found, we can suggest the user that this error is probably due to the sound having been already "uploaded".

[...]

PR #1085, opened 2017-12-13
CSV bulk upload also receive the location of the sounds

andrebola:
As described in issue #829 now the command csv_bulk_upload receives the
location of the sounds

When I was testing the changes I couldn't make it work because of the line  https://github.com/MTG/freesound/blob/master/sounds/management/commands/csv_bulk_upload.py#L119

which (I think) is incorrect, I also changed that line removing `self.`


sounds/management/commands/csv_bulk_upload.py
```diff
@@ -116,14 +122,15 @@ def handle(self, *args, **options):
             user = User.objects.get(username=usernamef)
 
             # 1 create dir and move sound to dir
-            directory = self.user.profile.locations()['uploads_dir']
```
ffont: wow this was a bug right? I mean the command does not really work (before applying this PR) right?

andrebola: I think so, it didn't work in my local.

ffont: Oh I saw the comment in description now, nevermind.

ffont:
Great, I'll merge master into it to pass the tests and merge the PR.

Issue #1075, opened 2017-12-11
IndexError: list index out of range


alastair:
https://logserver.mtg.upf.edu/sentry/freesound-web/issues/939/

```
IndexError: list index out of range
  File "search/views.py", line 147, in search
    filter_display = "pack:"+ val.split("_")[1]

IndexError: list index out of range
```

PR #1086, opened 2017-12-14
Fix error when user manually changes pack filter

andrebola:
issue #1075

The pack grouping filter should contain "_", if the user manually changes the parameter then we should not raise an error and skip the filter.

### DIFF
--- a/search/tests.py
+++ b/search/tests.py
@@ -122,3 +122,17 @@ class SearchPageTests(TestCase):
         # Check that we perform one single query to get all sounds' information and don't do one extra query per sound
         with self.assertNumQueries(1):
             self.client.get(reverse('sounds-search'))
+
+    @mock.patch('search.views.perform_solr_query')
+    def test_search_page_with_filters(self, perform_solr_query):
+        perform_solr_query.return_value = self.perform_solr_query_response
+
+        # 200 response on sound search page access
+        resp = self.client.get(reverse('sounds-search'), {"f": 'grouping_pack:"Clutter" tag:"acoustic-guitar"'})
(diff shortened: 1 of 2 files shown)


search/tests.py
```diff
@@ -122,3 +122,17 @@ def test_search_page_num_queries(self, perform_solr_query):
         # Check that we perform one single query to get all sounds' information and don't do one extra query per sound
         with self.assertNumQueries(1):
             self.client.get(reverse('sounds-search'))
+
+    @mock.patch('search.views.perform_solr_query')
+    def test_search_page_with_filters(self, perform_solr_query):
+        perform_solr_query.return_value = self.perform_solr_query_response
+
+        # 200 response on sound search page access
+        resp = self.client.get(reverse('sounds-search'), {"f": 'grouping_pack:"Clutter" tag:"acoustic-guitar"'})
```
ffont: If I understand correctly, this pack filter is wrong because it lacks the `ID_` part right? (it should be `grouping_pack:"1234_Clutter"`). So you're testing that pack filter is not applied right?

If that is the case I think you should add also an example where the filter is correct and check that it is there, and improve the comment so it is clear why the "bad" example is a bad example (why the pack filter is wrong).

search/views.py
```diff
@@ -138,19 +138,25 @@ def search(request):
     filter_query_split = []
     if filter_query != "":
         for filter_str in re.findall(r'[\w-]+:\"[^\"]+', filter_query):
+            display_filter = True
```
ffont: Maybe rename this to `valid_filter`?

Issue #1078, opened 2017-12-11
Allow users to change username


bdejong:
> As a user I want to be known as <newname> so that I can align all my online presence name with my cool new artist/band name

In Zendesk we send this out as a warning:

> Hi <oldusername>,
> 
> You have requested a username change. Please understand some of the disadvantages of changing your username before changing.
> 
> - All the links to your sounds on the internet will be broken.
> - People might have credited you when they used your sound, this credit will now "point to nothing"...
> - As your old username becomes freed up: someone might start using your old username!
> 
> If you are OK with all of these things happening, please reply to this message with the new username you would like to use. Please only use alphanumerical characters. Then make sure that the username you chose is not already taken, you can do this by creating a URL with your new username:
> 
> https://www.freesound.org/people/YOURUSERNAMEHERE
> 
> If this URL tells you there is no such page, you have a good new username.
> 
> greetings,
> 
> Bram

Perhaps it would be easier to automate this. I'm currently getting around 2 requests per week. Doing the username change "by hand" also has the weird side effect that the caches aren't updated until 48-or-so-hours afterwards, so the username is changed, but all the sound displays still point to the old username.

In zendesk we send this out as a warning:

> Hi <old username>,
> 
> We just changed your username to <new username>
> 
> From now on, please log in with that new username.
> 
> It will take a while before all your sounds show up with your new username, give it 48 hours and check again...
> 
> greetings,
> 
> Bram

ffont:
This is related to the recent changes introduced in https://github.com/MTG/freesound/pull/1059, in which we store old usernames in an `OldUsername` table so that we can redirect to the new sound URL when a user changes his username.

When working in this PR, we could also add an extra check in the account view before raising 404 when the username does not match any user, and try in the `OldUsername` table. In here:

https://github.com/MTG/freesound/blob/2b36a25f17720255ad6294e13e91b418b7e876ce/accounts/views.py#L796-L800

PR #1088, opened 2017-12-19
Allow user to change username

andrebola:
issue #1078
Added the username field to ProfileForm, also added the redirect to user page when old_username is stored


accounts/forms.py
```diff
@@ -236,6 +236,13 @@ class UsernameReminderForm(forms.Form):
 
 
 class ProfileForm(forms.ModelForm):
+    username = forms.RegexField(
```
ffont: Is there a way we can "import" the field from the user model form so that we make sure we apply exactly the same restrictions? (the regex, etc?)

andrebola: I guess we could define a form with only that field and then extend it in both forms, but the problem is that the fields are not equal, for example the help text is not the same.

ffont: but we could have them both have the same help text right?
or modify the help text in the `init` of the form. Maybe we could simply assign the field into a variable in the global scope and then add it to both forms? Like:

`
username_field = forms.RegexField(
    label=_("Username"),
    ...
)

class ProfileForm(forms.ModelForm):
    username = username_field

class RegistrationForm(forms.ModelForm):
    username = username_field
`

wouldn't that work? then given that we already overwrite `init` for the `ProfileForm`, we can set the help text to '' there (or keep it as is if useful to users).

andrebola: I think that could work, I will try it

ffont:
In Hipchat we discussed that we should probably only allow a maximum number of username changes. This should be easy to implement by adding a check in `validate_username` (or the equivalent method) and query the `OldUsername` table. I'd set the maximum number in a constant to 3 (so two changes allowed). We should also warn users (probably in the help text of the username field) that username can only be changed X times.

ffont:
I think this PR is now ready to merge. Nevertheless, before deploying we should add the missing redirects as described in https://github.com/MTG/freesound/issues/706. @andrebola did you notice we re-opened the issue? Do you plan to work on this or should I take it?

andrebola:
About the changes suggested in #706 I'm not sure if we should add the redirect since once the user changes the username why someone would access the old link? I understand that it's different for the sound link and maybe the profile page of the user, but for example the link for flagging a user I don't think someone would save it.

This is the list of urls with username:
https://github.com/MTG/freesound/blob/master/freesound/urls.py#L47-L79

ffont:
In some cases I agree it does not make sense, but in others it certainly does. For example packs, bookmarks, etc.. It's the same case as for sounds. After a quick inspection I think we surely need these:

```
url(r'^people/(?P<username>[^//]+)/$', accounts.views.account, name="account"),
    url(r'^people/(?P<username>[^//]+)/sounds/$', sounds.views.for_user, name="sounds-for-user"),
    url(r'^people/(?P<username>[^//]+)/comments/$', comments.views.for_user, name="comments-for-user"),
    url(r'^people/(?P<username>[^//]+)/comments_by/$', comments.views.by_user, name="comments-by-user"),
    url(r'^people/(?P<username>[^//]+)/geotags/$', geotags.views.for_user, name="geotags-for-user"),
    url(r'^people/(?P<username>[^//]+)/sounds/(?P<sound_id>\d+)/$', sounds.views.sound, name="sound"),
    url(r'^people/(?P<username>[^//]+)/sounds/(?P<sound_id>\d+)/download/.*$', sounds.views.sound_download, name="sound-download"),
    url(r'^people/(?P<username>[^//]+)/sounds/(?P<sound_id>\d+)/remixes/$', sounds.views.remixes, name="sound-remixes"),
    url(r'^people/(?P<username>[^//]+)/sounds/(?P<sound_id>\d+)/geotag/$', sounds.views.geotag, name="sound-geotag"),
    url(r'^people/(?P<username>[^//]+)/sounds/(?P<sound_id>\d+)/similar/$', sounds.views.similar, name="sound-similar"),
    url(r'^people/(?P<username>[^//]+)/sounds/(?P<sound_id>\d+)/downloaders/$', sounds.views.downloaders, name="sound-downloaders"),
    url(r'^people/(?P<username>[^//]+)/packs/$', sounds.views.packs_for_user, name="packs-for-user"),
    url(r'^people/(?P<username>[^//]+)/packs/(?P<pack_id>\d+)/$', sounds.views.pack, name="pack"),
    url(r'^people/(?P<username>[^//]+)/packs/(?P<pack_id>\d+)/download/.*$', sounds.views.pack_download, name="pack-download"),
    url(r'^people/(?P<username>[^//]+)/packs/(?P<pack_id>\d+)/downloaders/$', sounds.views.pack_downloaders, name="pack-downloaders"),
    url(r'^people/(?P<username>[^//]+)/sounds/(?P<sound_id>\d+)/display/$', sounds.views.display_sound_wrapper, name="sound-display"),
    url(r'^people/(?P<username>[^//]+)/downloaded_sounds/$', accounts.views.downloaded_sounds, name="user-downloaded-sounds"),
    url(r'^people/(?P<username>[^//]+)/downloaded_packs/$', accounts.views.downloaded_packs, name="user-downloaded-packs"),
    url(r'^people/(?P<username>[^//]+)/bookmarks/$', bookmarks.views.bookmarks, name="bookmarks-for-user"),
    url(r'^people/(?P<username>[^//]+)/bookmarks/category/(?P<category_id>\d+)/$', bookmarks.views.bookmarks, name="bookmarks-for-user-for-category"),
    url(r'^people/(?P<username>[^//]+)/following_users/$', follow.views.following_users, name="user-following-users"),
    url(r'^people/(?P<username>[^//]+)/followers/$', follow.views.followers, name="user-followers"),
    url(r'^people/(?P<username>[^//]+)/following_tags/$', follow.views.following_tags, name="user-following-tags"),
```
These are all the "public" URLs that could be posted somewhere and make sense for sharing. This is almost all URLs that have usernames so I'd consider putting the redirects everywhere.

ffont:
BTW, I'm making changes in this PR. Updated warning messages, fixed a couple of bugs and adding more tests. Will push later.

PR #1089, opened 2017-12-21
Store multiple downloads for the same user/sound

andrebola:
issue #840

After logging downloads for a period we conclude that we should store
the downloads if the Range header is not present. Also we keep the
downloads in the last 5 minutes so we don't store accidental doble
downlodas

### DIFF
--- a/sounds/views.py
+++ b/sounds/views.py
@@ -318,22 +318,13 @@ def sound_download(request, username, sound_id):
     if sound.user.username.lower() != username.lower():
         raise Http404
 
-    if settings.LOG_DOWNLOADS:
-        range_header = request.META.get('HTTP_RANGE', '').replace('bytes=', '').split('-')
-        if 'HTTP_RANGE' not in request.META or range_header[0] == "0":
-            downloads_logger.info('Download sound', extra={
-                'user_id': request.user.id,
-                'user_ip': request.META.get('HTTP_X_FORWARDED_FOR'),
-                'protocol': request.META.get('HTTP_X_FORWARDED_PROTOCOL'),
-                'session_id': request.session.session_key,
-                'user_agent': request.META.get('HTTP_USER_AGENT'),
-                'method': request.method,
-                'sound_id': sound_id,
-                'range': request.META.get('HTTP_RANGE', None),
-            })
+    if 'HTTP_RANGE' not in request.META:
+        cache_key = 'sdwn_%s_%d' % (sound_id, request.user.id)
+        if cache.get(cache_key, None) == None:
+            if not Download.objects.filter(user=request.user, sound=sound).exists():
+                Download.objects.create(user=request.user, sound=sound, license=sound.license)
 
-    if not Download.objects.filter(user=request.user, sound=sound).exists():
-        Download.objects.create(user=request.user, sound=sound, license=sound.license)
+            cache.set(cache_key, True, 60*60*5) # Don't save downloads for the same user/sound in 5 minutes
     return sendfile(sound.locations("path"), sound.friendly_filename(), sound.locations("sendfile_url"))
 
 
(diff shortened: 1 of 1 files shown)


andrebola:
We probably need to solve some issues with the disk on db-server before merging this changes.

ffont:
Now that we have space, I guess this could be released as well? @andrebola @alastair

andrebola:
Yes, I think is time to merge this!

sounds/views.py
```diff
@@ -318,22 +318,13 @@ def sound_download(request, username, sound_id):
...
-                'sound_id': sound_id,
-                'range': request.META.get('HTTP_RANGE', None),
-            })
-
-    if not Download.objects.filter(user=request.user, sound=sound).exists():
-        Download.objects.create(user=request.user, sound=sound, license=sound.license)
+    if 'HTTP_RANGE' not in request.META:
+        cache_key = 'sdwn_%s_%d' % (sound_id, request.user.id)
+        if cache.get(cache_key, None) == None:
+            if not Download.objects.filter(user=request.user, sound=sound).exists():
```
alastair: This only creates a download object if it doesn't already exist. I thought that this patch was to add multiple downloads :confused:

sounds/views.py
```diff
@@ -318,22 +318,13 @@ def sound_download(request, username, sound_id):
...
-                'user_agent': request.META.get('HTTP_USER_AGENT'),
-                'method': request.method,
-                'sound_id': sound_id,
-                'range': request.META.get('HTTP_RANGE', None),
-            })
-
-    if not Download.objects.filter(user=request.user, sound=sound).exists():
-        Download.objects.create(user=request.user, sound=sound, license=sound.license)
+    if 'HTTP_RANGE' not in request.META:
+        cache_key = 'sdwn_%s_%d' % (sound_id, request.user.id)
```
alastair: combining %s and %d?

sounds/views.py
```diff
@@ -318,22 +318,13 @@ def sound_download(request, username, sound_id):
...
-                'session_id': request.session.session_key,
-                'user_agent': request.META.get('HTTP_USER_AGENT'),
-                'method': request.method,
-                'sound_id': sound_id,
-                'range': request.META.get('HTTP_RANGE', None),
-            })
-
-    if not Download.objects.filter(user=request.user, sound=sound).exists():
-        Download.objects.create(user=request.user, sound=sound, license=sound.license)
+    if 'HTTP_RANGE' not in request.META:
```
alastair: Can we have an explanation here:

> Download managers and some browsers use the `range` header to download files in multiple parts. We have observed that all clients first make a GET with no `range` header (to get the file length) and then make multiple other requests. We ignore all requests that have `range` header because we assume that a first query has already been made.
We additionally guard against users clicking on download multiple times by storing a sentinel in the cache for 5 minutes.

sounds/views.py
```diff
@@ -318,22 +318,13 @@ def sound_download(request, username, sound_id):
...
+        cache_key = 'sdwn_%s_%d' % (sound_id, request.user.id)
+        if cache.get(cache_key, None) == None:
```
alastair: Please use `is` to compare against `None`

sounds/views.py
bdejong: on top of that, with the number of downloads we have, won't this flood the memcache?

ffont: You're right @alastair, this `if` should not be here.
And regarding @bdejong comment, we did not think about this when we discussed the best way to implement this feature. Hopefully not, it's not that much data to save and it's only for 5 minutes.

ffont:
I'll add fixes for what's commented here directly in master as it has been already merged and the changes are small.

Issue #1077, opened 2017-12-11
Batch describe sounds


ffont:
> As a user I want to be able to describe sounds in batch so I don't have to go one by one adding tags, description, etc.

It has been a long standing feature request to be able to describe sounds in batch. We have a method to do that using a CSV file which includes metadata for a number of audio files and then creates `Sound` objects. We do this for large audio collections running the management command [`csv_bulk_upload`](https://github.com/MTG/freesound/blob/master/sounds/management/commands/csv_bulk_upload.py). This command has been improved recently to make it more robust (see https://github.com/MTG/freesound/issues/829), but still some minor improvements are needed (see the issue).

It would be easy to enable users to upload a CSV file and do the same thing that the management command does without "our intervention". The workflow would be that users upload files normally, but then instead of clicking in "describe" they click in a new "batch describe" option which allows them to upload a CSV file. We should validate the CSV file and show users any relevant feedback and if all its fine, then we create the `Sound` objects asynchronously in a worker. It would be nice to still show the user feedback about the progress in real-time, or at least in his home page.

The "batch describe" should only be enabled to users that have already uploaded more than X sounds.

andrebola:
The required changes on command csv_bulk_upload are already in a PR, so we can start discussing how this should work. 

This are the steps that the user will follow to describe multiple sounds:

 - In sounds upload page the user has a list of uploaded sounds, here we will add a button to upload a csv file and submit it.
 - After the user submits the CSV with the data of the sounds, we first check that everything is fine and if there is any error we should display it, otherwise we will start the async task of creating the sounds and processing them. 
 - If there are no errors we will display all the sounds in a table and update every X seconds the status. 

In the second bullet we need to figure out how we will validate if everything is fine without creating the actual sounds. Finally in the third bullet we need to see how to update the status of the async task with the user, this might be by using the gearman info together with the data stored in each sound, or we might need to create a new model for this tasks.

andrebola:
I created 3 mockups with the screens for the this bullets and uploaded them to invision.

ffont:
Nice @andrebola !

[...]

PR #1093, opened 2018-01-03
Batch describe sounds

andrebola:
issue #1077

Support for batch describe sounds by uploading a CSV file. The information of a new CSV uploaded is stored in the model  BulkUploadProgress, in which we store the status that could be: 
 - Not Started (initial state)
 - Validated
 - Start
 - Finished

Once the file is uploaded is validated, then the user will check the output of the validation and eventually continue with the process (this will change the state to 'Start'), finally the sounds will be processed and it will change the state to 'Finished'.

The responsable of changing the states is a new management command:  csv_bulk_describe


andrebola:
Currently the uploaded csv files are stored in a folder named with the id of user inside another folder with the name 'csv' - @bdejong proposed to put all the csv files from the different users in the same folder, since it would simplify the code. 

I will leave this PR for review so you can make all the comments and propose other corrections, then I will continue with the changes that we decide.

ffont:
I've been testing this a little bit, just some thoughts for improvements:

- Add `BulkUploadProgress` to admin
- Add sample.csv file for download so users can use it as template (in the describe page)
- Do not allow to set username in the csv for bulk uploads from users, otherwise users can assign sounds to other users ;) 
- Validation step (bulk describe command), move it to an async task
- I got some errors that were not caught on validation step, we need better input validation in `csv_bulk_upload`. the error I got was because of malformed geotag
- Highlight the wrong field in the table when showing validation info to users
- Sound creation step (bulk describe command), move it to an async task
- Show status of `BulkUploadProgress` in user home, so the process can be retaken at any time
- Redirect to home after hitting red button (show a message above with info)
- If a sound is already present in freesound we will only know when sound objects are being created, not in the validation phase. I guess this can't really be helped without big changes.
- Add unit tests for the whole process
- Improvements in UI and the way things are explained to users. This can wait until the rest is ok.

andrebola:
I will start to work in the first 4 items in the list, so if @ffont you want to work on this at the same time we can coordinate.

andrebola:
I also added the list of uploaded csv files on users home, according to item 7 of the list.

ffont:
I did another test and it all seems to work really nice. Thanks for your work @andrebola, this is going to be great! I'll take it from here and add the missing things and some improvements. Hopefully this will be our new cool feature of march :)

ffont:
Functionality is now complete and deployed in Freesound test. Nevertheless we still have no unit tests and a number of things should be refactored (particularly in relation to the validation of metadata fields). Our next planned steps are to implement the unit tests and then do the internal refactoring. Meanwhile, other Freesound devs can test the functionality in test env and suggest improvements.

ffont:
To refactor the validation of sound fields we will reuse the forms defined in `sounds.forms`. We can pass input data to the form object in a dictionary, then run `form.is_valid()` and get errors (if any) in a dictionary with `form.errors.as_dict()`.

We'll still need to manually validate that the geotag information is in the form `float,float,int`, and then pass the individual values to the form.

Sound fields validation that happens in the API replicates some of this code, therefore we should update the functions there to reuse these forms as well.

ffont:
After some more refactoring, improvements and after adding unit tests, this PR is now ready for review. Current version is deployed in test so you can give it a try. 

There are many changes in this PR, so I'm not sure what would be the best way to go through the review process. Any suggestions @alastair @andrebola @bdejong @philtgun ?

philtgun:
- A bit confusing what is the relation of bulk describe to uploading sounds, but that's just maybe me that is not familiar with the flow of first uploading sound and then describing them. Not sure if the interface should be available if you have no uploaded sounds to be described. Maybe redirect user or suggest to go to upload sounds interface with appropriate notification?

- Error message for geotag maybe can have some improvement
image

- Possible bug of having an error without explanation? Geotag is `0, 0.1, 14` here
image

Otherwise the process looks good

bdejong:
Initial comments:

- I would heavily reduce the amount of text on the "pick your sounds to describe" page
- I would reorganize the selection of sounds and then the choice of description (1. choose the sounds... 2. pick the description)

ffont:
> I would reorganize the selection of sounds and then the choice of description (1. choose the sounds... 2. pick the description)

Maybe it is not clear enough but you don't select sounds if you go the "batch describe" way. Selection is only for the manual description. Maybe the two ways of describing should be further separated.

bdejong:
# Looks awesome.
👍  I went through it with a fine comb, ... <pedantic mode enabled>

Instructions say I should use ";", but in the error on top says I should use ",":
- Invalid header. Header should have the following 7 columns: audio_filename, name, tags, geotag, description, license, pack_name

No rows in CSV:
- No error, but only option is to delete the file and restart. That's a bit weird

Red fields:
- I have to wait each time for the hover to show me something, I'd rather click, which is faster

Line numbers:
- The line numbers start at "2", that's a bit weird. Maybe start them at "1"?

Regarding errors:
- In general: Instead of showing rerrors I would show hints and explain the fields. So, for example for an error in "Audio filename" I would say:
  - Problem: the audio file "what you wrote" can't be found.
     Hint: this should be the name of oneof the files you uploaded.

If the sound file is not found and the "name" field is blank it gives an additional error:
- I.e. "this fields needs to be 5 characters"

License:
- Problem: can't find a license called "what you wrote
   Hint: this field should be either X, Y or Z
    
Tags:
- Problem: you wrote "..."
  Hint: this field should contain at least 3 distinct tags, separated by commas, e.g. "bang,crash,boom"

If you add the same tag it produces an error:
- A hint would be nice: use 3 *different* tags

Geotags:
- "You should zoom in until you reach at least zoom 11"
   Is funny, but not so helpful :-D

License:
- Case insensitive match would be nice...
   I wrote "CREATIVE COMMONS 0" and it was unhappy

There is no way to flag sounds as explicit through bulk upload.

"Click on the button below to close this bulk description process and stop showing it in your home screen."
- Saying this is a bit weird if you don't first say it's on the home screen.

- "Bulk Describe Sound processes" should probably be at the top of the profile, not the bottom.

You can delete the word "sounds" from all of these lines.
I would also remove all lines that show zeros. What's the point?
- Sounds successfully described: 3
- Sounds published: 1
- Sounds in moderation: 0
- Sounds currently processing: 0
- Sounds pending processing: 0
- Sounds failed processing: 2
- Sounds that failed description: 0
- Sounds remaining to be described: 0

ffont:
Thanks for your comments @bdejong, I'll work on addressing them. One thing It's not clear to be, in the errors table, do you propose to add the "problem: xxxx hint: yyyy" text in each cell that has errors? or are you proposing this to be the contents of the hover/onclick? If we put it in the cell then there is no need for hover (which I think is in fact better), but then the problem is that it can take a lot of space. One option is to make the table scrollable not only in height but also in width and have a minimum column width so that error messages won't look too bad. What do you think?

Btw regarding line numbers, they start at 2 because I assume header is number 1 and then the line number seen in Freesound interface should match the line number in Excel or whatever editor people is using. Do you think I should still change it to start at 1 (0=header)?

bdejong:
@ffont 
- problem/hint: in the popup, not in the cell... btw, the table is already scrolling in width for me...
- Re. line number, yeah, ... true. Leave it.

ffont:
This PR has not been updated incorporating feedback from @bdejong and @philtgun. We could now enter a testing phase with Freesound moderators. Once deployed, we should add the following text to the FAQ wiki page, under the sounds section:

```
### I have many sounds to upload, is there a way to describe many sounds in bulk?

You can describe your uploaded audio files in batches of 10 sounds using the Freesound [describe sounds](/home/describe/) page. However, if you have many sounds to describe this can become a tedious process. Luckily, if you have demonstrated being a faithful Freesound user, when you go to the [describe sounds](/home/describe/) page you'll see an extra option for **describing sounds in bulk using a CSV file**.

Using that option you can provide descriptions for several files at once by **uploading a CSV file** which contains all the necessary metadata information. The uploaded CSV file must be created following the instructions below. 

<i>Batch describe CSV file format instructions:</i><a name="csv-file-instructions"></a>
                
The uploaded CSV file must start with a header row and then have one row per audio file to be described. It must use semi-colon **;** as the delimiter character. These are the columns that should be included for each row (those marked with ***** are required, the others can be left blank):

<ul>
    <li><b>audio_filename*</b>: filename of the sound to describe (must be one of the filenames listed above in "Your uploaded files" section).</li>
    <li><b>name</b>: name to be given to the sound. If left blank, <i>audio_filename</i> will be taken as a default.</li>
    <li><b>tags*</b>: tags for the sound, separated with spaces like in the standard audio description forms.</li>
    <li><b>geotag</b>: geotag information for the sound. The geotag should be indicated as
"<i>latitude,longitude,zoom</i>" (separated by spaces). Latitude must be in the range [-90, 90]. Longitude must be in the range [-180, 180]. Zoom must be in the range [11, 21]. If this field is left blank, no geotag will be added to the sound. For example: "<i>41.40348, 2.189420, 18</i>" is a valid geotag.</li>
    <li><b>description*</b>: textual description given to the sound. Can be multiline like sound descriptions introduced using the standard describe interface.</li>
    <li><b>license*</b>: name of the license to be given to the sound. Must be one of "<i>Creative Commons 0</i>", "<i>Attribution</i>", or "<i>Attribution Noncommercial</i>".</li>
    <li><b>pack_name</b>: name of the pack in which the sound should be included. If there exists no pack with this name, a new pack will be created. If this field is left blank, the sound won't be included in any pack.</li>
    <li><b>is_explicit*</b>: must be "<i>1</i>" if sound contains explicit content, or "<i>0</i>" if it does not.</li>
</ul>

For your convenience, here we provide an <a href="/media/sample.csv">example CSV file</a> that you can use as a template.
```

bdejong:
Weirdly enough I can't see the bulk upload option anymore in test.freesound.org ... Help?

ffont:
>  Weirdly enough I can't see the bulk upload option anymore in test.freesound.org ... Help?

You don't see it because now we have maps deployed in test ;)
I'll deploy it again today to share it with moderators, then you can test again. I addressed almost all of your comments, there are only a couple of things that I left unmodified.

ffont:
Thanks for your feedback @qubodup and the evil CSV file. After considering all feedback received I decided to work on the following improvements:

- [ ] Properly handle unexpected errors during CSV validation.
- [ ] Re-organize describe page so that  the "normal" way and the "csv" way appear one in each column instead of one before the other (which could lead to confusion).
- [ ] Add example filled in CSV file instead of a template, and link to it more prominently.
- [ ] Make clear that CSV files can easily be created with software like Excel and Google Spreadsheets.

ffont:
Some more comments coming from moderator's feedback:

- In the template, it would be great to have the "Attribution Noncommercial" license as one of the 3 examples given (replacing one of the two "Attribution" examples).
- The geotag is a little fiddly to figure out in this process - is it possible to add the geotag functionality that is currently possible when editing sound information (in that we can select a point on the map and it displays the longitude/latitude info to the side - it is possible to have a page to aid with that? Otherwise, I suspect most people will add that information afterwards)
- A brief note on case sensitivity might help (e.g. pack names with the same words but different capitalized letters will create two different packs)
- Some short words are removed from tags during the process (granted, irrelevant words like "to" and "are", but just flagging in case this is a bug)

qubodup:
> Bulk description of files
> Validation results of the data file "k][213 asdasda]2[{}';;;2.csv"
> We have encountered some major issues when validating your data file:
> 
> An unexpected error occurred while validating your data file
> Please make sure you are following these guidelines to create your data file.

[k][213 asdasda]2[{}';;;2.zip](https://github.com/MTG/freesound/files/2044477/k.213.asdasda.2.2.zip)

ffont:
The error caused by the "bad" data file you created is known, but the system now informs the user that there is an unexpected error instead of hanging forever in "validation" so we consider this to be the "fix" for now. We'll collect validation errors caused by bad files like the one you created and add fixes to remove the number of unexpected errors, but it could be that at release time your file is not yet supported.

PR #1094, opened 2018-01-05
Check oldusernames before  404 redirect on views

andrebola:
As discussed in issue #706 and in #1088 we now check the old usernames before displaying a 404 error. For this I added a method in utils app that checks the OldUsername table.


accounts/views.py
```diff
@@ -689,7 +685,9 @@ def attribution(request):
 
 
 def downloaded_sounds(request, username):
-    user = get_object_or_404(User, username__iexact=username)
+    user = get_user_from_oldusername(username)
+    if user == None:
+        raise Http404
```
ffont: I like the idea of the util function that gets the user, but this pattern:
```
user = get_user_from_oldusername(username)
    if user == None:
         raise Http404
```
I see everywhere. Why not making `get_user_from_oldusername` raise 404 insetad? In fact, it could be renamed to `get_user_or_404` and I think it would look more elegant everywhere.

bookmarks/tests.py
```diff
@@ -49,6 +49,20 @@ def test_no_user(self):
 
         self.assertEqual(404, resp.status_code)
 
+    def test_bookmarks_oldusername(self):
+        user = User.objects.get(username='Anton')
```
ffont: An alternative way of testing the redirects would be to:
- get user
- rename
- save (just as you do)
- call page with original username
- call page with new username
- check that content of page with both usernames is the same

Wouldn't that be easier and cleaner?
In the current tests you only check that it works with the old username, but not with the new. If we break the `get_user...` util function for "new" usernames the test would not catch that.

sounds/views.py
```diff
@@ -316,7 +318,9 @@ def sound_download(request, username, sound_id):
                                                     reverse("sound", args=[username, sound_id])))
     sound = get_object_or_404(Sound, id=sound_id, moderation_state="OK", processing_state="OK")
     if sound.user.username.lower() != username.lower():
-        raise Http404
+        user = get_user_from_oldusername(username)
+        if user == None or user.id != sound.user_id:
+            raise Http404
```
ffont: I think we also need a util function here. Wouldn't it make sense to have a `get_sound_or_404`  function that checks for old usernames as well? This possibly applies to other places as well.

bookmarks/tests.py
andrebola: Yes, it's true. This is because in general there is another test for the regular case, where the username is not in OldUsername table. But you're right is not exactly the same case.

andrebola:
I implemented a decorator that checks the parameter 'username' from the view and makes a redirect if the username is in the oldusername table. 

Something that we should take into account is that previously for the sound page I checked if the sound_id correspond to the username in the table `oldusername` before making the redirect, now the redirect is done and then it checks if the sound_id is valid for that user, if it's not valid shows a 404.

ffont:
All right, I think it looks good. I'd like to propose renaming the util functions in this way:

`redirect_old_username` -> `redirect_if_old_username`
`get_user_from_oldusername` -> `get_user_from_username_or_oldusername`

> Something that we should take into account is that previously for the sound page I checked if the sound_id correspond to the username in the table oldusername before making the redirect, now the redirect is done and then it checks if the sound_id is valid for that user, if it's not valid shows a 404.

I think that is fine.

utils/username.py
```diff
@@ -0,0 +1,38 @@
...
+from django.urls import reverse
+
+def get_user_from_oldusername(username):
+    # Helper to get the user from an username that could have changed
+    user = None
+    try:
+        user = User.objects.select_related('profile').get(username__iexact=username)
+    except User.DoesNotExist:
+        try:
+            old = OldUsername.objects.get(username__iexact=username)
```
bdejong: you don't need the `old = ` part, you can just use

```user = OldUsername.objects.get(username__iexact=username).user```

bdejong:
I LOVE commits where the code gets cleaner all over the place. ❤️ 

I would only change ONE thing now... I would remove the comment `# We don't use @redirect_if_old_username_or_404 here because we don't want to redirect for internal links` which you put in many places and just add this as documentation to the redirect_if_old_username_or_404 function! I.e. don't use this function for internal views

ffont:
Your comment makes a lot of sense, I'll do that!

PR #1106, opened 2018-01-12
Remove files already in .gitignore

philtgun:
These files are listed in .gitignore, but still present in repository. Not sure if they are necessary or not, please check. If they are necessary, maybe .gitignore should be updated


alastair:
Thanks!

PR #1114, opened 2018-01-23
Pack downloads final updates

andrebola:
Following with #914, now PackDownload and PackDownloadSound models are filled and used to display pack download information. 

Also note for the attribution page it was a bit obscure how I merge the results from Downloads and PackDownloads table, the reason to do it in that way is because of the pagination and the sort by 'created' field which is common for both tables.

IMPORTANT: after applying the migrations the items in Downloads table referring packs will be deleted, so the command that copies the downloads to the PackDownload table must be executed before.


alastair:
> IMPORTANT: after applying the migrations the items in Downloads table referring packs will be deleted, so the command that copies the downloads to the PackDownload table must be executed before.

Can you make this 2 pull requests? Our deploy script runs migrates during deploy, and I don't trust us to disable this before we deploy it.

ffont:
Hmm looking at this problem it's making me think (don't hate me too much): is there a reason why we split pack downloads and sound downloads? We could simply add the `PackDownloadSounds` table and make it refer to `Download` right? Maybe I missed something at some point and this was already discussed. I mean, the goal of the issue is to store the sounds that are downloaded within a pack, not necessarily to split pack and sound download tables. Does that make sense? (maybe it doesn't just commenting) @alastair @andrebola

alastair:
The reason for splitting into a second packs downloads table is to reduce the number of tables where we store two types of data and filter based on the value of one of the columns of the table. We've been doing this with generic relations, Downloads is one of the last ones. It simplifies our indexes because we don't have to have a bunch of indexes where particular columns are null/not null.

ffont:
Agreed via Hipchat discussion that the approach here is correct. I'm wondering if we should rename `Download` to `SoundDownload` but anyway hehe

andrebola:
I just moved the migration files to #1115

ffont:
Instructions for merging and deploying this PR:
0. Read the **important note** below and make necessary changes in this PR and in #1115.
1. Update `python manage.py copy_downloads` to add the bit of code that fills in `PackDownloadSounds` as well (in **master** branch).
2. Run `python manage.py copy_downloads` (will take considerable time)
3. Stop freesound
4. Run again `python manage.py copy_downloads` to have it perfectly in sync (should be fast). Check that once finished all content has been properly copied.
5. Deploy #1114 
6. Start freesound
7. Make sure that new downloads are properly stored in new models (maybe during a couple of days)
8. Deploy #1115 (which deletes the no longer needed rows in `Download` model and the no longer needed 'pack' column in `Download` model)
9. Open a beer and have fun

**IMPORTANT NOTE:** Before doing any of this #1114 and #1115 should be updated so that they can work *separately*. #1115 should only include the delete things, meaning the migration for deleting the no longer needed rows and field AND the corresponding changes in the model (which are now in #1114). #1114 should only have the code for the new logic of storing downloads in new models AND [the related name thing](https://github.com/MTG/freesound/pull/1115/files#diff-5483a865c596c87fdeb29a3cc43b3b87R19) which is currently in #1115.

ffont:
Update: we've completed steps from 0 to 2

ffont:
This PR has been merged and deployed. Therefore in the instructions above we finished step 6. Remaining steps will be copied in https://github.com/MTG/freesound/pull/1126 so we continue from there.

PR #1115, opened 2018-01-24
Add migration files with changes in download tables

andrebola:
This PR only contains the migrations files for the PR #1114. So, it should be merge after.


ffont:
Closing this PR without merging as https://github.com/MTG/freesound/pull/1126 will be used instead.

PR #1116, opened 2018-01-24
Add setting to disable graylog

philtgun:
Connection routine to graylog was not guarded, added LOG_TO_GRAYLOG setting in `similarity_settings.py`


alastair:
nice :+1:

Issue #1054, opened 2017-11-27
Signal non-https image tags as an error


bdejong:
If the user right now types

`<img src="http://blah.jpg" />` this will be replaced with `<a href="http://blah.jpg">blah.jpg</a>` because otherwise we get non-secure images in secure pages, which most browsers flag as non-secure.

However, this is done without alerting the user (and also to support the things which are already in the DB).

For new messages we should show an error to the user:

> You are trying to embed an image which is hosted on a non-secure website (http). For security reasons we don't support this in Freesound. Please provide an image link to a secure (https) website

alastair:
I think that we shouldn't disallow this - it's possible that images are only available in http, and the user may be OK with a link. We could have a real-time notification below the text box saying this (or something similar), but don't prevent the user from saving it.

ffont:
I thought we had some sort of notification for this. Anyway, as alastair says we should not throw an error but I agree in showing a warning.

alastair:
@ffont tested this on the forum reply page and it works.
We should verify that it works on all forms that use an `HtmlCleaningCharField`

PR #1120, opened 2018-02-02
Add more checks for unsafe img src in forms

philtgun:
Went through instances of `HtmlCleaningCharField` in forms and made sure that templates have `unsecureImageCheck` JS calls on proper fields (#1054)

### DIFF
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -334,6 +334,8 @@ class ProfileForm(forms.ModelForm):
         model = Profile
         fields = ('home_page', 'about', 'signature', 'sound_signature', 'is_adult', 'not_shown_in_online_users_list', )
 
+    def get_img_fields(self):
+        return [self['about'], self['signature'], self['sound_signature']]
 
 class EmailResetForm(forms.Form):
     email = forms.EmailField(label=_("New e-mail address"), max_length=254)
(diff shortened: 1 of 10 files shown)


accounts/forms.py
```diff
@@ -334,6 +334,8 @@ class Meta:
         model = Profile
         fields = ('home_page', 'about', 'signature', 'sound_signature', 'is_adult', 'not_shown_in_online_users_list', )
 
+    def get_img_fields(self):
```
ffont: I'd be nice to document this as this function as just by looking at the form class it won't be obvious to understand why this is here. If I'm not wrong the fields returned are the `HtmlCleaningCharField` fields, therefore you might want to rename the method to something like `get_html_cleaning_fields`.

ffont: In fact, maybe you can refactor the function so that it dynamically checks the type of the fields and returns them? (so that we don't have to update it if we ever add a new `HtmlCleaningCharField`.

philtgun: The thing is that field being of `HtmlCleaningCharFields` class doesn't mean it needs img notifications (there are examples of title fields in other modules)

templates/accounts/edit.html
```diff
@@ -66,8 +66,9 @@ <h3>Delete account</h3>
 
 <script>
   $(function() {
-    unsecureImageCheck($('#id_profile-about'));
-    unsecureImageCheck($('#id_profile-signature'));
+      {% for field in profile_form.get_img_fields %}
```
ffont: I guess this should be renamed to `profile_form.get_img_check_fields` now?

philtgun: Updated, forgot to change name before

Issue #1103, opened 2018-01-11
Disallow using email-like usernames


bdejong:
Users sometimes sign up with their email address as username. We should really disallow people to sign up with a "^[^@]+@[^@]+$" username ;)

ffont:
Allowing usernames of the form of an can generate problems like the one see in this issue: https://github.com/mtg/freesound/issues/1102

alastair:
I recommend just blocking "@", any sort of magic to try and discover if something is a valid email address will just cause endless problems.

ffont:
> I recommend just blocking "@", any sort of magic to try and discover if something is a valid email address will just cause endless problems.

But Django already has that implemented right? Can't we reuse the validation from email field and "negate" it?

alastair:
Oh yeah, if we're using that in some places then it seems reasonable to use the reverse too

bdejong:
The fix for this:

https://github.com/MTG/freesound/blob/4da3894e812e38bc4a6f9a3e0617dd0fe43d6c6a/accounts/forms.py#L130

to

`regex=r'^[\w.-]+$',`

I almost did it via in-line editing in github, but then there are no tests for this.....

ffont:
> I almost did it via in-line editing in github, but then there are no tests for this.....

Love it! You need to add a test inline :)

philtgun:
- [ ] Check if django changed validation regex for username field

alastair:
https://github.com/django/django/blob/master/django/contrib/auth/validators.py#L20

Issue #1102, opened 2018-01-11
MultipleObjectsReturned: get() returned more than one User -- it returned 2!


ffont:
https://logserver.mtg.upf.edu/sentry/freesound-web/issues/990/

```
MultipleObjectsReturned: get() returned more than one User -- it returned 2!
(2 additional frame(s) were not displayed)
...
  File "utils/session_checks.py", line 36, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "accounts/views.py", line 237, in resend_activation
    & Q(is_active=False))

MultipleObjectsReturned: get() returned more than one User -- it returned 2!
```

bdejong:
![screen shot 2018-01-11 at 14 25 44](https://user-images.githubusercontent.com/805269/34827708-5b4cd238-f6db-11e7-92de-0d2f0db4926e.png)

ffont:
In this case this happened because there is a pair of users A and B where A.username == B.email.
We should not allow emails as usernames, currently people can login with email.

philtgun:
Now, even if we don't allow email-like usernames, this scenario still can occur if the username of existing user is an `email1` (with his actual email being different) and there will be a new user using `email1` as his e-mail. 

We need to add the check against usernames to email field for registration form to prevent it as well as to clean up the data.

Also it might be duplicated in 'forgot my password/username' form

philtgun:
We can leave the existing users that have `@` symbol in the username be if it is not a duplicate of the email. However, we need to check the number of users that fall into the latter case and based on it make the decision on how to deal with them: 
- either consider it a special case and add safeguard in the code
- just remove this scenario from occurring if there are just a few

philtgun:
This query will find users that can cause the scenario to happen
```sql
select u1.username, u1.email, u2.username, u2.email from auth_user u1, auth_user u2 where lower(u1.email) = lower(u2.username);
```
Drop `lower` to increase performance but ignore case insensitivity

alastair:
42000 rows from this query, but that includes the same account where the email is the same as the usename.
664 rows where the username of a user is the same as an email from a different user.

The next thing we should decide is
 * Can we delete the users? (if they're old and haven't logged in in a long time)
 * Should we make SameUser objects (for the 664 case)
 * Can we merge them?
 * Can we change the login to only select the account based on username in this case

ffont:
Oh no, this gives me a headache. I'll try to read this very slowly and see if I understand...

PR #1125, opened 2018-02-09
Update username field (disallow @), add validation tests

philtgun:
- Replace `username_field` with a class, changed from `RegexField` to `CharField` with `Validator`
- Remove `@` from validation regex (#1103)
- Add some tests for validation
- Updated `help_text` to match regex
- Fixed some typos

### DIFF
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -35,6 +35,7 @@ from django.template import loader
 from django.urls import reverse
 from django.core.mail import EmailMultiAlternatives
 from django.core.exceptions import PermissionDenied
+from django.core.validators import RegexValidator
 from multiupload.fields import MultiFileField
 from django.conf import settings
 from accounts.models import Profile, EmailPreferenceType, OldUsername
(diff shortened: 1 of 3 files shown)


accounts/forms.py
```diff
@@ -285,10 +293,15 @@ def __init__(self, request, *args, **kwargs):
...
             self.fields['username'].disabled = True
-        self.fields['username'].help_text = help_text
+        self.fields['username'].help_text += " " + help_text
 
     def clean_username(self):
         username = self.cleaned_data["username"]
+
+        # If user has accidentally cleared the field, treat it as unchanged
+        if not username:
+            username = self.request.user.username
```
ffont: nice catch!

accounts/tests.py
```diff
@@ -1028,12 +1029,11 @@ def test_resend_activation_code_from_username(self):
         self.assertEqual(resp.status_code, 200)
         self.assertEqual(len(mail.outbox), 1)  # Check no new email was sent (len() is same as before)
 
-    def test_resend_activation_code_from_username(self):
+    def test_resend_activation_code_from_long_username(self):
```
ffont: What does "long username" mean here?

accounts/tests.py
```diff
@@ -1205,9 +1205,35 @@ def test_change_username_form_admin(self):
...
+        form = self.TestForm(data={'username': 'email@username'})
+        self.assertFalse(form.is_valid())
+
+    def test_short_invalid(self):
+        """ Should be longer than 3 characters """
+        form = self.TestForm(data={'username': 'a'})
+        self.assertFalse(form.is_valid())
+
+    def test_long_invalid(self):
+        """ Should be shorter than 3 characters """
```
ffont: Instead of 3 it should say 30 here right?

ffont:
After we deploy this, will we have to do something with existing usernames with @ characters in them?
Should we email them and ask for change? Or change directly and send an email to notify? How many of these users we have?

accounts/tests.py
philtgun: the name of the function was the same as the previous one defined at line 1014 I changed it because essentially it was redefined, while the test is different

accounts/tests.py
philtgun: updated

philtgun:
That's a good question. I have made a related comment in issue #1102. I think it would be nice to do it, but it probably should be part of the major version upgrade.

I guess replacing `@` in usernames with `_` and notifying them via e-mail would be best from dev point of view, but what about the user experience? Maybe we should do it for less active users and ask more active users to do it directly. And in this case should we keep `OldUsername`s with `@`?

But I think dealing with this problem is separate issue

PR #1126, opened 2018-02-09
Remove `pack` field from `Download` table

ffont:
This PR re-implements what was previously in https://github.com/MTG/freesound/pull/1115/ after the comments in https://github.com/MTG/freesound/pull/1114#issuecomment-364424777. Here we simply delete the `pack` field of the `Download` table and remove all entries in the `Download` which have `sound_id=NULL` (i.e. entries which corresponded to pack downloads information that now is stored in `PackDownload` table).


ffont:
After having merged and deployed https://github.com/MTG/freesound/pull/1114, the remaining steps for this PR are:

1. Make sure that new downloads are properly stored in new models (maybe during a couple of days)
2. Modify this PR to remove all filters of `sound_id__isnull=False` in queries to the `Download` table. This is because after applying this PR the `Download` table will only include rows corresponding to sound downloads therefore `sound_id` will never be null.
3. Do what @alastair mentions in the [comment above](https://github.com/MTG/freesound/pull/1126#issuecomment-365614164) and edit this PR to it deletes the no longer needed indexes.
4. Deploy this PR (which deletes the no longer needed rows in `Download` model and the no longer needed 'pack' column in `Download` model) - the migration might take time, we should consider moving to maintenance mode or testing before in test.
5. Close issue https://github.com/MTG/freesound/issues/914 
6. Open a beer and have fun (finally)

ffont:
Finally this PR is merged with master after successful testing in local and fs test. Will deploy tomorrow morning.

Issue #1015, opened 2017-11-07
MultipleObjectsReturned: get() returned more than one User -- it returned 2!


alastair:
When we created our unique constraint on the email field we didn't consider that we allow logins using case insensitive usernames and email addresses. We should verify how many case insensitive email address duplicates are in the database and see if we need to create rows in the `accounts.SameUser` table for these users.

https://logserver.mtg.upf.edu/sentry/freesound-web/issues/820/

```
MultipleObjectsReturned: get() returned more than one User -- it returned 2!
(22 additional frame(s) were not displayed)
...
  File "utils/session_checks.py", line 36, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "accounts/views.py", line 89, in login
    response = LoginView.as_view(template_name=template_name, authentication_form=authentication_form)(request)
  File "accounts/modelbackend.py", line 43, in authenticate
    user = User.objects.get(email__iexact=username)

MultipleObjectsReturned: get() returned more than one User -- it returned 2!
```

philtgun:
Create new migration, similar to [this](https://github.com/MTG/freesound/commit/78189137a12f18892331fd51f53c3c8a73326585#diff-8bb6e460bac7a55a676b489191eed711R11), but for case-insensitively-equal emails

alastair:
This query gets the accounts that fall into this case:

    select upper(email), count(*) from auth_user group by upper(email) having count(*) > 1

PR #1136, opened 2018-02-21
Create migration to deal with case-insensitive duplicate emails

philtgun:
#1015 
As the case insensitive email lookup is already part of registration form, this should eliminate all scenarios for the error from the issue


ffont:
Deployed in test and it works well, will merge with master. 9 new `SameUser` objects are created. After adding the unique case-insensitive index, new users with same case-insensitive email can't be created as `duplicate key value violates unique constraint...` error is raised.

PR #1143, opened 2018-03-11
Progress with new FE

giuband:
- Progress on the homepage layout
- Add a bunch of useful classes for styles (`modifiers`)
- Add custom onscroll behavior for navbar on homepage
- Add search fields (we can already start plugging the BE to them, just to define how to use them)
- Add veeery basic audio player


giuband:
Wow, I honestly had no clues about this dependency warning. Which dependency is it? I can't see it from here

I need to check styleguide, it was working for me yesterday but then in the last few hours I stopped using it to focus on the page itself.

I'll check the sound images and try to reuse them!

ffont:
The dependency in question is `ssri`. In particular Github says that 'ssri' should be updated from `5.1.0` to at least `5.2.2`.

ffont:
Merging this for now, @giuband will include the fix for the styleguide and the dependency update in the next bulk of commits.

Issue #1142, opened 2018-03-09
Moderation ticket page improvements


philtgun:
- [x] Editing ticket status in the ticket page should not be possible
- [x] Changing sound moderation status in the ticket page to OK should close the ticket
- [x] Changing sound moderation status in the ticket page to Delete should delete the sound and close the ticket (not sure if this happens like that)
- [x] Changing sound moderation status in the ticket page to Pending should: 1) set ticket to accepted if there is an assignee; or 2) set ticket to new if there is no assignee
- [x] In ticket view there should be an option to return ticket
- [x] In ticket view the form currently named "sound state" should probably be renamed into something like "ticket actions" more similar to the "moderation assigned" tickets interface

Issue #1092, opened 2018-01-03
New "assign to me" buttons in moderation


ffont:
> As a moderator I want to be able to easily assign all currently open tickets from a user even if they have already been assigned to other moderators

> As a moderator I want to be able to assign a moderation ticket from the ticket page even if it is already assigned to another moderator

Currently moderation tickets can be assigned in 2 ways:

- By clicking on the "Assign to me" link below the user avatar in the "Users with new sounds section" (https://freesound.org/tickets/moderation/). This will assign ALL sounds pending moderation by user X.
- By clicking on the "Assign to me" link below each sound in the tardy users or tardy moderators section (https://freesound.org/tickets/moderation/). This will assign a single sound.

These two methods are insufficient because once tickets have been assigned to a moderator, there is no easy way in which another moderator can re-assign them to himself. We can improve that situation by implementing 2 new "assign to me" buttons:

- Add "assign to me" link in the ticket page (e.g. https://freesound.org/tickets/e775c5f550264ec5b59f92d1b04c6f9d/), next to the "assigned to" field.
- Add "assign all user X sounds pending moderation to me" link in the "uploaded sounds pending moderation" page for user X (e.g. https://freesound.org/tickets/moderation/pending/TechnoScore/)

philtgun:
For the first scenario (https://freesound.org/tickets/moderation/) the "Users with new sounds" section contains the _sounds that are pending moderation with no assignee_. For example, in case if there are 2 sounds that are new, but one is already assigned to mod, this page will show only the sound that is not assigned. 

If you click on "1 new sound" near user avatar, you will go to "uploaded sounds pending moderation" page (https://freesound.org/tickets/moderation/pending/xxxx) that contains _all pending sounds with or without assignee_ (2 in the example). Is it intentional? Because it is a bit confusing to see label that says "1 new sound" that goes to list of 2 sounds.

philtgun:
Yes, two sections for pending sounds should clear the confusion. Also it would be useful to add the info about the moderator for the pending assigned sounds

PR #1151, opened 2018-03-20
Moderation UI update

philtgun:
Addresses issues #1092 and #1142 
- In the ticket view page:
  - the sidebar with ticket and sound state change has been changed to be similar to ticket actions in the moderation queue page
  - added "assign to me" button that shows up if the assignee is different from you
- In the pending sounds by user page:
  - sounds with no mod assigned to the ticket are shown on top and have indication of it in the info box
  - sounds with mod assigned have the assignee shown in the info box to the right 
  - if there are some sounds that are not assigned to you, the button 'assign all to me' is shown that assigns all sounds (new and already assigned)
- Fixed some display inconsistencies, typos and errors
- Added tests to verify that ticket and sound state change are working as intended

### DIFF
--- a/templates/accounts/pending.html
+++ b/templates/accounts/pending.html
@@ -43,6 +43,13 @@
                     </div>
                 </div>
                 <div class="sample_information" style="width: 250px;">
+                    <span>
+                        {% if ticket.assignee %}
+                            Assigned to {{ ticket.assignee.username }}
+                        {% else %}
+                            Not assigned yet
+                        {% endif %}
+                    </span><br />
 
                     <span class="date">{{sound.created|date:"F jS, Y"}}</span><br />
                     {% if not moderators_version %}
--- a/templates/accounts/pending.html
+++ b/templates/accounts/pending.html
@@ -81,6 +81,13 @@
         have any uploaded sounds awaiting moderation.</p>
     {% endfor %}
 
+    {# TODO: provide info on how many sounds by this user are assigned to you, and if all - don't show 'assign to me' #}
+    <div class="moderation-user-list"><p>
+        <a href="{% url "tickets-moderation-assign-user" user.id %}">
+        <img src="{{media_url}}images/moderation_plus.png"/>
+            Assign all to me</a>
+    </p></div>
+
     {% if show_pagination %}
         {% show_paginator paginator page current_page request "moderation ticket" %}
     {% endif %}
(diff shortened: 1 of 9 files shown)


templates/accounts/pending.html
```diff
@@ -43,16 +48,30 @@ <h2>Uploaded sounds awaiting moderation {% if moderators_version %} by  <a class
                     </div>
                 </div>
                 <div class="sample_information" style="width: 250px;">
+                    <span>
+                        {% if ticket.assignee %}
+                            Assigned to <a href="{% url "account" ticket.assignee.username %}">{{ ticket.assignee.username }}</a>
```
ffont: Maybe say "Assigned to moderator {{uname}}" instead of "Assigned to {{uname}}". This small change might make things a bit more clear for some users.

templates/tickets/ticket.html
```diff
@@ -21,10 +21,16 @@ <h3>Ticket details</h3>
...
-                     {{ ticket.assignee.username }}
+                     <a href="{% url "account" ticket.assignee.username %}">{{ ticket.assignee.username }}</a>
                  {% else %}
                      no one
-                 {% endif %}</li>
+                 {% endif %}
+
+                 {# Show 'assign to me' link if the assigned moderator is not you #}
+                 {% if ticket.assignee.id != user.id %}
+                 (<a href="{% url 'tickets-moderation-assign-single-ticket' ticket.sender.id ticket.id %}?next=ticket">assign to me</a>)
```
ffont: Maybe we could also add the green cross here as it is always used in "assign to me" actions.

templates/tickets/ticket.html
```diff
@@ -60,10 +66,9 @@ <h3>Add new message to the ticket</h3>
         </div> <!-- /#ticket-left -->
 
         <div id="ticket-right">
-            {% if perms.tickets.can_moderate %}
+            {% if perms.tickets.can_moderate and ticket.sound %}
```
ffont: That's ok, but lets rename the header below which says "Ticket linked content" and the text "No linked content" displayed when there is no sound. I'd say "Ticket's sound" instead of "Ticket linked content"m and "This ticket has no assigned sound or the sound assigned to this ticket was deleted." instead of "No linked content".

(for some reason Github does not allow me to add this comment to the specific lines in code)

templates/tickets/ticket.html
philtgun: It makes the UI look ugly without further adjustments, I would prefer not putting it there. Anyway as I understand it will be redesigned with the new frontend.
screenshot_20180323_152817

ffont: ok, fair enough

Issue #1119, opened 2018-02-01
Reactivate success page has wrong title


bdejong:
When a user reactivates, this is the "success" page:

![image002](https://user-images.githubusercontent.com/805269/35704539-b511f260-0754-11e8-86b6-96aba626c483.png)

This is especially annoying as people who don't have an account (they failed to register properly) go to the reactivate, and they get this result page.

PR #1154, opened 2018-03-21
Fix result page after resending activation email

philtgun:
Adresses #1119


ffont:
Looks good to me, maybe @alastair want's to have a look at the wording?

Issue #1135, opened 2018-02-21
Problems with "about" display field: when to show


bdejong:
- Admins should always see the About: etc fields
- Users should always see their own About: etc fields
- The fields of users who didn't download but did upload sounds should be shown to other users

philtgun:
Should we consider uploaded sounds that are pending moderation good enough to enable about field, or should we only check for accepted ones?

ffont:
I'd say accepted ones only, but is not really a very big deal.

ffont:
This was implemented, but it won't be active until we fix https://github.com/MTG/freesound/issues/1156 and uncomment line https://github.com/MTG/freesound/pull/1155/files#diff-f0dc1751b547d90c03894819dd68ea0dR843

philtgun:
Also enable relevant tests by proper inheritance [here](https://github.com/MTG/freesound/pull/1155/files#diff-1ec00e92826603d6c4ec5e5c0c8f2f77R1242)

PR #1155, opened 2018-03-21
Expand condition for about field visibility

philtgun:
Adresses #1135
- expand condition for about field visibility
- add relevant tests


philtgun:
Adresses #1135
- expand condition for about field visibility 
- add relevant tests

PR #1158, opened 2018-03-31
Usable players + almost completed new homepage

giuband:


ffont:
Thanks @giuband!
That's getting better and better.
I have some comments but I'll wait until you consider you have the first "final" version of the homepage. Merging this PR now.
BTW, the demo sound is great ;)

Issue #1101, opened 2018-01-11
Make sure `Sound.invalidate_template_cache` works properly and add it where needed


ffont:
We observed that `Sound.invalidate_template_cache` has been broken sometimes without us noticing. We need to add a unit test for this function which renders a sound, runs invalidate template cache and makes sure the proper keys are removed from the cache. The function can be easily broken if we add new elements in the cache key in the template and don't update `Sound.invalidate_template_cache` accordingly: https://github.com/MTG/freesound/blob/8e53ca9237740a185ae8d837f4e8fe86dc6c3f87/sounds/models.py#L766-L772
To properly test that we probably need to mock the cache backend or make sure there is one that works in unit tests.

Furthermore, we also observed that there are some situations in which we should call `Sound.invalidate_template_cache` but we don't (for example, after adding/deleting a sound comment). We should revise all elements in the cache key and make sure we are calling `Sound.invalidate_template_cache` when appropriate.

PR #1168, opened 2018-04-20
Sound template cache 

philtgun:
Adresses issue #1101
- Update the template cache key computation function with one provided by django
- For scenarios that might change the data in cached template blocks:
  - Add missing cache invalidation calls
  - Write unit tests that verify the proper invalidation as well as proper sound UI updates
- Reduced number of cache indexes (sound moderation and  state) and add related invalidation calls

### DIFF
--- a/templates/sounds/display_sound.html
+++ b/templates/sounds/display_sound.html
@@ -7,7 +7,7 @@
     If you change this cache index, be sure to change the invalidation in
     ratings/views.py as well
 {% endcomment %}
-{% cache 43200 display_sound sound_id request.user.is_authenticated sound.processing_state sound.moderation_state request.user.profile.is_adult  %}
+{% cache 43200 display_sound sound_id request.user.is_authenticated is_explicit sound.processing_state sound.moderation_state sound.similarity_state %}
 
 <div class="sample_player_small" id="{{ sound_id }}">
 
--- a/templates/sounds/display_sound.html
+++ b/templates/sounds/display_sound.html
@@ -7,7 +7,7 @@
     If you change this cache index, be sure to change the invalidation in
     ratings/views.py as well
 {% endcomment %}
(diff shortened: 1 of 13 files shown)


utils/cache.py
```diff
@@ -19,13 +19,14 @@
...
-from hashlib import md5
-from django.utils.http import urlquote
-from functional import compose
 
 def invalidate_template_cache(fragment_name, *variables):
-    args = md5(u':'.join(map(compose(urlquote, unicode), variables)))
-    cache_key = 'template.cache.%s.%s' % (fragment_name, args.hexdigest())
-    cache.delete(cache_key) 
+    cache_key = get_template_cache_key(fragment_name, *variables)
+    # print('Invalidating: ', cache_key)
```
ffont: oups ;)

templates/sounds/sound.html
```diff
@@ -41,7 +41,7 @@
 
 {% block section_content %}
 
-{% cache 3600 sound_header sound.id request.user.is_authenticated request.user.profile.is_adult %}
+{% cache 3600 sound_header sound.id is_explicit %}  {# cache both blurred and normal version for explicit sounds #}
```
ffont: Did you make sure that `request.user.is_authenticated` is not used for other things that might affect the template?

templates/sounds/display_sound.html
```diff
@@ -7,7 +7,7 @@
     If you change this cache index, be sure to change the invalidation in
     ratings/views.py as well
 {% endcomment %}
-{% cache 43200 display_sound sound_id request.user.is_authenticated sound.processing_state sound.moderation_state request.user.profile.is_adult  %}
+{% cache 43200 display_sound sound_id is_authenticated is_explicit %}
```
ffont: Hmm, I guess having the `moderation_state` and `processing_state` in the key does not make sense at all because sounds have the same state for all users, but did you make sure that this is the case?

Issue #1156, opened 2018-03-22
Profile.num_sound_downloads and Profile.num_pack_downloads not updating


ffont:
We recently added two new fields in the `Profile` model, namely `num_sound_downloads` and `num_pack_downloads`. Originally, these fields were updated using Django signals after download events (see  [this code](https://github.com/MTG/freesound/commit/938a669904dfddbeb5d181370972f8fa8e27def1), changes in `sounds/models.py` signals). However, after a bad merge (conflicts not properly fixed), the lines of code which update `Profile.num_sound_downloads` and `Profile.num_pack_downloads` were removed. Because we have no test to check `Profile.num_sound_downloads` and `Profile.num_pack_downloads` are ok and we don't (yet) make an extensive use of these fields, we did not realize until now about the bad merge.

We should re-implement functionality (including tests!) so that `Profile.num_sound_downloads` and `Profile.num_pack_downloads` are updated when users download sounds and packs. This should happen for sound downloads ([here](https://github.com/MTG/freesound/blob/4786f2b62f19ad680609d6554ca9ab192830f7cd/sounds/models.py#L1064-L1076)) and also for pack downloads ([here](https://github.com/MTG/freesound/blob/4786f2b62f19ad680609d6554ca9ab192830f7cd/sounds/models.py#L1091-L1101)).

Once this is done, `Profile.num_sound_downloads` and `Profile.num_pack_downloads` fields will be reliable again, and we can revert that [temporary change with the "show about"](https://github.com/MTG/freesound/pull/1155/commits/06c3fa9adb9532068d9ebfcc16fa8027c0cff9ff) condition.

Also, after this is done an optimization could be implemented in the [`downloaded_sounds`](https://github.com/MTG/freesound/blob/master/accounts/views.py#L703) and the [`downloaded_packs`](https://github.com/MTG/freesound/blob/master/accounts/views.py#L718) views as `user.profile.num_sound_downloads` and `user.profile.num_pack_downloads` can be passed to the paginator to avoid one extra query to the DB to get the total count of objects (similarly as is done [here](https://github.com/MTG/freesound/blob/master/sounds/views.py#L924)).

[...]

PR #1170, opened 2018-04-24
Fix broken num_sound_downloads and num_pack_downloads

philtgun:
Adresses #1156 
- Re-implement broken update routines that are bound to django signals
- Add relevant test assertions
- Add recalculation to `report_count_statuses`
- Optimize `downloaded_sounds` and `downloaded_packs` views by passing these fields to paginator to reduce db queries

Would be nice to run updated `report_count_statuses` on test environment just to make sure it is safe to update.


sounds/models.py
```diff
@@ -1049,7 +1050,7 @@ class Meta:
 
 
 class Download(models.Model):
-    user = models.ForeignKey(User)
+    user = models.ForeignKey(User, related_name='sound_downloads')
```
ffont: Did you run `makemigrations? I think this changes generate migrations in Django that you should add in the PR.

philtgun: Good point! Included in next commit

ffont:
Next steps:

- Deploy this code in test env
- Measure how long it takes to run `report_count_statuses` with and without downloads. Compare with the time it takes in production db and report results here.

ffont:
I did run `report_count_statuses` in test env before and after deploying this PR and these are the results.

- Before deploy (run1): 21m, 550 objects updated
- Before deploy (run2): 14m, 0 objects updated
- After deploy (run1): 286m, 275000 objects updated
- After deploy (run2): 221m, 0 objects updated

The query for getting number of downloads per user (https://github.com/MTG/freesound/blob/profile-num-sound-and-pack-downloads/general/management/commands/report_count_statuses.py#L196) takes 3 hours to compute!

I'm not sure how this affects the DB, in test I was able to navigate it while this query was running, but I think it was a bit slower.

The idea is that this command we should just run it from time to time (or ideally never), so maybe it is not a problem if we run it once a week or once every two weeks for example and we get a bit of a slowdown. We can also skip the downloads part of the command (the option is already implemented) and run the other parts once a week and the downloads thing once a month (it's less important). I think we should just deploy to production and run the command and see how Freesound is affected.

philtgun:
Technically after we run this command to fix the number of downloads, if there are no bugs, we don't need to run it again in a while. The cronjob for it is weekly and it skips recalculation of the downloads.

ffont:
Yeah, we just need to run it from time to time so I'm not much worried about it taking 3 hours. But still we should evaluate how much it impacts performance. I'll merge for now and test when deployed to have a better understanding.

Issue #884, opened 2017-06-08
Disable email sending to users whose addresses bounce, and inform them


alastair:
We can now receive notifications from our mail provider telling us if an email address that we are delivering to has bounced. We should use these notifications to make sure that we don't keep sending emails to accounts that can't receive email.

* If the email was an activation email, the account can never be used, delete the account.
* If it was any other email, mark the user who has this email as "email invalid"
* When sending emails, check if the user's email is marked as invalid, and if so, skip it.
* If a user logs in and has an invalid email, require them to change their email address to reactivate their account

alastair:
https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html#bounce-object

philtgun:
- [ ] db change: add invalid email field 
- [ ] django management command to use AWS API to get bounced emails and set flags (will be run periodically)
- [ ] cleanup script that removes unactivated users with invalid emails
- [ ] code checks for invalid email before sending and notify user with the middleware once he logins

philtgun:
Related to removing unactivated users with invalid emails - do we remove unactivated users after certain timeout? If not, should we? @alastair @ffont

alastair:
No, we don't automatically remove unactivated users, but we can add this.  We mark spammy users as inactive while they are in moderation, and so we don't want to delete all of them.

philtgun:
Two things to adress:
- [ ] Amazon returns punycode email domains, they should be processed properly
- [ ] Need to add potion to save bounce data for possible processing and analytics

ffont:
Closing this issue as it has been merged (https://github.com/MTG/freesound/pull/1189) and deployed

PR #1172, opened 2018-05-02
Email bounce integration

philtgun:
First pull request to address issue #884
- Add `EmailBounce` model that will be populated from AWS queue
- Extend `send_email_template` to support user objects in addition to email strings
- Add check for email being valid before sending
- Add relevant tests

Should we extend `EmailBounce` model to have more information from [AWS SQS](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html#bounce-object)?


ffont:
This PR looks good to me. Maybe after experimenting with getting queue info from AWS we can decide whether more info should be added to the `EmailBounce` model, but for now it might be just fine. I'd suggest that @alastair takes a look as this is related with the tickets he was assigned to and might have some extra ideas to bring in.

utils/mail.py
```diff
@@ -105,7 +120,7 @@ def send_mail(subject, email_body, email_from=None, email_to=list(), reply_to=No
         return False
 
 
-def send_mail_template(subject, template, context, email_from=None, email_to=[], reply_to=None):
+def send_mail_template(subject, template, context, email_from=None, email_to=list(), reply_to=None):
```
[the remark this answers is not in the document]
philtgun: I made this change due to warning from PyCharm, and you are right in your guess. Good explanation is [here](http://docs.python-guide.org/en/latest/writing/gotchas/)

accounts/models.py
```diff
@@ -443,3 +447,18 @@ class OldUsername(models.Model):
...
+
+
+class EmailBounce(models.Model):
+    user = models.ForeignKey(User, related_name="email_bounces")
+
+    # Bounce types
+    UNDETERMINED = 'UD'
+    PERMANENT = 'PE'
+    TRANSIENT = 'TR'
+    type = models.CharField(db_index=True, max_length=2, default=UNDETERMINED)
```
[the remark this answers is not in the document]
philtgun: Updated in commit 4ab7115

utils/mail.py
```diff
@@ -55,8 +55,8 @@ def wrapper(subject, email_body, email_from=None, email_to=list(), reply_to=None
             email_to = [email_to]
 
         from accounts.models import SameUser
-        emails_mapping = {su.secondary_trans_email: su.orig_email for su in SameUser.objects.all()}
-        email_to = list(set([emails_mapping.get(email, email) for email in email_to]))
+        user_mapping = {su.secondary_user: su.main_user for su in SameUser.objects.all()}
+        email_to = list(set([user_mapping.get(user, user) for user in email_to]))  # strings are going to be left as is
```
alastair: We originally made this decorator because we needed a way to go from a user -> modified email address -> user -> real email address.
Now that we have a user object passed in as the `to` field, we could do this in a nicer way. We can use `user.profile.get_sameuser_object().orig_mail` (catch exception) to get the correct email to send to if the user is one of these merged email users.
If this is the case, then we can remove the decorator and put this logic inside `send_mail` where you check `email_is_valid`.

After discussing with @ffont we decided that it's not a great idea to make `email_to` accept either users or email addresses. We think that there is only one place where we have to use an email address (in the email change form). In this case, we could have a separate parameter (`raw_email_to`?) which contains these email addresses. Allow this function to only be called with email_to or raw_email_to.

philtgun: For the record there are several more places where the `send_email` function is used with the admin/support email addresses from `config.py`.

alastair: I wonder if we could split this into 2 methods, send_email_to_user and send_email_to_admin (which either reads the admin email addresses from settings, or read email addresses from the user profile and then calls a generic send_email method.

philtgun:
Due to unnecessary workarounds that would have been implemented to keep AWS bounce info integration (#884) completely separate from email refactor (#1179), some parts of the latter will be addressed in this pull request - specifically Falsiness of `email_to` to send emails to support.

freesound/local_settings.example.py
```diff
@@ -104,3 +105,5 @@
 RAVEN_CONFIG = {
     'dsn': 'DSN_PROVIDED_BY_SENTRY',
 }
+
+AWS_SQS_QUEUE_URL = ''  # also proper AWS credentials should be configured in ~/.aws
```
alastair: Make sure that you also open a pull request on the freesound-deploy repository to include the actual queue url.

Is `~/.aws` the only way of doing credentials? I'm not sure how well this will work for us in production, it would be great if there was an alternative way of setting the credentials (via environment variables?)

accounts/management/commands/process_email_bounces.py
```diff
@@ -0,0 +1,79 @@
...
+                VisibilityTimeout=0,
+                WaitTimeSeconds=0
+            )
+            has_messages = len(response['Messages']) > 0
+
+            for message in response['Messages']:
+                receipt_handle = message['ReceiptHandle']
+                body = loads(message['Body'])
+
+                data = loads(body['Message'])
```
alastair: I prefer to `import json` and use `json.loads` here to be more explicit as to where this method comes from

accounts/management/commands/process_email_bounces.py
```diff
@@ -0,0 +1,79 @@
...
+        sqs = client('sqs')
+        queue_url = settings.AWS_SQS_QUEUE_URL
+
+        total = 0
+        has_messages = True
+        while has_messages:
+            # Receive message from SQS queue
+            response = sqs.receive_message(
+                QueueUrl=queue_url,
+                MaxNumberOfMessages=1,
```
alastair: Does this mean that you process 1 message at a time? What is the alternative - can you read more? Why would we want to read only 1 at a time instead of 10 or 50?

accounts/tests.py
```diff
@@ -1282,3 +1283,48 @@ def test_admin(self):
...
+
+    def test_request_email_change(self):
+        pass  # TODO
+
+    @mock.patch('accounts.management.commands.process_email_bounces.client')
+    def test_populate_bounce(self, client):
+        message_body = '{"Type": "Notification", "Message": "{\\"notificationType\\":\\"Bounce\\",' \
+                       '\\"bounce\\":{\\"bounceType\\":\\"Permanent\\",\\"bounceSubType\\":\\"Suppressed\\",' \
+                       '\\"bouncedRecipients\\":[{\\"emailAddress\\":\\"user@freesound.org\\"}],' \
+                       '\\"timestamp\\":\\"2018-05-20T13:54:37.821Z\\"}}"}'
```
alastair: To make this easier to read you could do `json.dumps({some dict})`

alastair:
This looks really good now. I'll work with @ffont next week to deploy it on test so that we can check email sending, and then do the deploy.

freesound/local_settings.example.py
philtgun: No, it is possible to specify credentials in code. I will update the config to reflect that, I was not sure what would be the preferred way of doing this.

accounts/management/commands/process_email_bounces.py
philtgun: Maximum number is 10, but from my testing there are usually duplicates present in them, that usually results in ~4 unique messages to process. I pondered making it a config variable because there are pros and cons for 1 and 10.

accounts/management/commands/process_email_bounces.py
```diff
@@ -0,0 +1,80 @@
...
+import logging
+
+logger = logging.getLogger('console')
+
+
+class Command(BaseCommand):
+    help = 'Retrieves email bounce info from AWS SQS and updates db'
+
+    def handle(self, *args, **options):
+        sqs = client('sqs', aws_access_key_id=settings.AWS_ACCESS_KEY_ID,
```
alastair: One final thing - if these settings are empty we could print an error and exit, instead of trying and failing to run

accounts/tests.py
```diff
@@ -942,31 +944,31 @@ def test_fix_email_issues_with_both_users_email_change(self):
         self.assertRedirects(resp, reverse('messages'))
 
     def test_replace_when_sending_email(self):
+        return True  # TODO: implement new way of testing this
```
ffont: What should we do with this? If we want to skip it we can use the `skipIf` decorator as well.

freesound/local_settings.example.py
```diff
@@ -104,3 +105,10 @@
 RAVEN_CONFIG = {
     'dsn': 'DSN_PROVIDED_BY_SENTRY',
 }
+
+# AWS
+
+AWS_ACCESS_KEY_ID = ''
+AWS_SECRET_ACCESS_KEY = ''
+AWS_SQS_QUEUE_URL = ''
+AWS_SQS_MESSAGES_PER_CALL = 1
```
ffont: These parameters are not used if set to empty right?
Why not moving them to `settings.py` so that these are not needed to be manually copied to `local_settings.py` (and in production we override them normally)?

ffont: done here https://github.com/MTG/freesound/pull/1172/commits/2a1b5b03cc9087c1db58caddc2eaabba44ed3099

accounts/tests.py
ffont: Fixed here https://github.com/MTG/freesound/pull/1172/commits/40c503d742541d85fa047798b76b79a3fbbe3b2d

Issue #1175, opened 2018-05-07
Implement map features with Mapbox


ffont:
Because of new new pricing policy with Google Maps API, we'll have to switch to a new maps provider if we want to continue offering same map features. Mapbox seems a great (and cheaper) alternative. We should implement map features with Mapbox and evaluate/compare them with GMaps.

ffont:
This has now been deployed to production, but I'm not closing this issue yet as we should still get rid of remaining google maps code after some days of having mapbox in prod.

ffont:
Closing as google maps code has been removed now!

PR #1176, opened 2018-05-08
Implementation of maps using Mapbox API

ffont:
This PR includes the implementation of all map features in Freesound using Mapbox API.
Google Maps version is still available in this PR, and cat be loaded by adding `gmaps=1` to the query parameters of any page using maps.

It also includes the following changes:

- Simplified map sharing: now the URL of the map page can be used for sharing (no need to click on the "share this map" and copy from the "link" section).
- Improved map embeds:  embed code now has options to chose map size and turning clustering on/off.
- Added map search in the main geotags page and in the maps for sound edit and sound describe. This is a new feature 🎉! In fact this is quite useful for adding geotags to sounds and for browsing the sounds of the world.
- Increased a bit the map size in main geotags page so that it better fits the page.

You can test the new maps in Freesound test: https://test.freesound.org/

Issue: https://github.com/MTG/freesound/issues/1175


philtgun:
I think the 'zoom in' link on marker popup is broken. Would be nice if aggregation bubbles would be clickable with zooming like in gmaps. Otherwise looks good.

ffont:
Fixed issues reported by @philtgun and added a couple more improvements.

Issue #1171, opened 2018-04-25
Bug/feature: simplify register page


bdejong:
Our current register page is too complex and users keep thinking they signed up, wait for the activation email (which they won;t get because the signup failed) and then contacting support to ask for help. Searching for their email address in the DB yields no results.

Potentially we could simplify the register page dramatically:

1. username
2. email (no repeat!)
3. password (no repeat, with "reveal" button!)
4. CAPTCHA
5. [button, Create account] ("By creating an account you agree with our <link>TOS</link>)

This should reduce all errors...

bdejong:
By password reveal I mean something like this:

![eye](https://i.ytimg.com/vi/IW8NHJarjl0/hqdefault.jpg)

bdejong:
( I've heard rumors that "By creating an account you agree with our TOS" is no longer valid under GDPR. )

ffont:
> ( I've heard rumors that "By creating an account you agree with our TOS" is no longer valid under GDPR. )

We know that. We are working with our lawyer so that he tells us what we have to do in our case.

PR #1177, opened 2018-05-09
Simpler registration page

ffont:
This PR simplifies the registration page as described in issue https://github.com/MTG/freesound/issues/1171. Here is a screen shot of how it looks like now:

sshot_registration_form

Visible changes:

- Removed duplicated fields
- Add option to show password
- Remove first name/last name fields
- Captcha errors appear now next to captcha widget
- TOS acceptance checkbox appears to the left of its label instead of to the right

Also as part of this PR I added unit tests and fixed wrong captcha behavior when no  captcha API key is present (it should be disabled).


ffont:
BTW, after deploying that (or before) we should set all first_name/last_name to '' for existing users.

templates/accounts/registration.html
```diff
@@ -1,47 +1,76 @@
...
+                        $('#check_success').toggle(data.result);
+                }, "json" );
+            }
+        });
+    });
+
+    function togglePasswordInputType() {
+        var passwordInput = document.getElementById("id_{{ form.password1.name }}");
+        var eyeImg = document.getElementById("eye-img");
+        if (passwordInput.type === 'password'){
```
alastair: We should probably make the submit form change this back to type=password if a user has set it, so that we can hint to the browser that it's a login and that it should offer to save it

accounts/forms.py
```diff
@@ -127,15 +127,13 @@ def get_user_by_email(email):
...
-        required_str = _("Required. ") if required else ""
-
         super(UsernameField, self).__init__(
             label=_("Username"),
             min_length=3,
             max_length=30,
             validators=[RegexValidator(r'^[\w.+-]+$')],  # is the same as Django UsernameValidator except for '@' symbol
-            help_text=_(required_str + _("30 characters or fewer. Can contain: letters, digits, underscores, dots, "
-                                         "dashes and plus signs.")),
+            help_text=_(_("30 characters or fewer. Can contain: letters, digits, underscores, dots, "
```
bdejong: the double underscore here seems weird...

accounts/forms.py
```diff
@@ -146,13 +144,9 @@ class RegistrationForm(forms.Form):
     recaptcha_response = forms.CharField(widget=CaptchaWidget, required=False)
     username = UsernameField()
 
-    first_name = forms.CharField(help_text=_("Optional."), max_length=30, required=False)
-    last_name = forms.CharField(help_text=_("Optional."), max_length=30, required=False)
     email1 = forms.EmailField(label=_("Email"), help_text=_("We will send you a confirmation/activation email, so make "
```
bdejong: does it make sense to have the "1" postfix? same for password...

bdejong:
Ah, wanted to try in test, but it's not deployed yet... Will wait until deployed for checking it out.

ffont:
I want to deploy the batch upload PR in test, so this won't be available. Nevertheless I could deploy this one temporarily to test when you need it.  Just let me know.

bdejong:
Could we switch to a git (work)flow where we can merge multiple things to a
"dev" (staging, ...) branch and deploy them all to test, then merge to
master and deploy to production?

accounts/forms.py
ffont: I would prefer not to do that now.

accounts/forms.py
ffont: Fixed in https://github.com/MTG/freesound/pull/1177/commits/b53628dc0101b4718be271805e9133151242844c

accounts/tests.py
```diff
@@ -332,6 +332,85 @@ def test_user_save(self):
         self.assertEqual(Profile.objects.filter(user=u).exists(), True)
         u.save()  # Check saving user again (with existing profile) does not fail
 
+    def test_user_registration(self):
+        RegistrationForm.captcha_key = ''  # Override this property so that captcha is not enabled in the tests
+        # NOTE: using override_settings and setting RECAPTCHA_PUBLIC_KEY to '' won't work because RegistrationForm is
```
alastair: You could fix this by setting `RegistrationForm.captcha_key` in the form's `__init__`, or by reading from settings directly where you need to use it.
Ideally you would revert this setting back at the end of the test

sounds/forms.py
```diff
@@ -241,8 +241,9 @@ class FlagForm(forms.Form):
 
     def clean_recaptcha_response(self):
         captcha_response = self.cleaned_data.get("recaptcha_response")
-        if not captcha_response:
-            raise forms.ValidationError(_("Captcha is not correct"))
+        if self.captcha_key:
+            if not captcha_response:
+                raise forms.ValidationError(_("Captcha is not correct"))
```
alastair: How often are we using translatable strings?

templates/api/minimal_registration.html
```diff
@@ -73,43 +73,57 @@
         }
 
     </style>
+
+    <script type="text/javascript">
+        function setPasswordType() {
+            // Set type to password so that browser offers to store it
```
alastair: It would be nice to share this with the main login page (even though we're doing the interface change, etc)

alastair:
I see that there are some fixtures that still have first_name, last_name in them. I guess this doesn't completely matter because we're not going to delete these columns (as it's a django table), but do we want to clear them anyway?

general/templatetags/util.py
```diff
@@ -46,4 +47,8 @@ def duration(value):
 
 @register.filter
 def in_list(value,arg):
-    return value in arg
\ No newline at end of file
+    return value in arg
+
+@register.filter(name='is_checkbox')
+def is_checkbox(field):
```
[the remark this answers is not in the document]
ffont: Probably part of a solution for rendering the registration form that was not finally implemented. Will remove.

PR #1185, opened 2018-05-30
New ToS and license upgrade workflow

alastair:
Add a middleware to check if the user has agreed to the new terms of service, and if not send them to a page where they can read them and agree.
This replaces the old terms of service middleware. We keep the old `profile.accepted_tos` flag so we still know which users never agreed to the original ToS, but it is not used anymore.

Added as a separate field to save space - adding new columns to the Profile model will increase its size significantly, and if they're mostly null then it doesn't make sense to waste the space.

This PR also adds support for CC 4.0 licenses (co-existing with 3.0) and provides users methods to bulk upgrade their old 3.0 licenses to 4.0. This PR has **migrations** to add new tos consent table and to add the new licenses.

This PR should be merged on the 4th of April as this is when the new ToS be effective. Also during that day, the contents of the wiki page in help/tos_web/ should be replaced by those in help/tos_web_2022 and the link from the blog post to tos_web_2022 updated. We can also decide to keep a link to the told ToS.


alastair:
Tests that broke due to newly created users not having agreed to the terms have been fixed

accounts/models.py
```diff
@@ -343,6 +347,14 @@ class Meta(SocialModel.Meta):
         ordering = ('-user__date_joined', )
 
 
+class GdprAcceptance(models.Model):
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+    # Automatically add the date because the presence of this field means that
+    # the user accepted the terms
+    date_accepted = models.DateTimeField(auto_now_add=True)
+    # TODO: IP Address & Browser?
```
alastair: remove

ffont:
I added some stuff to this PR and I think it is ready to deploy once we have the final text for the privacy policy and update the summary for the `gdpr_consent.html`.

PR #1189, opened 2018-06-15
Email bounce integration part 2

philtgun:
Addresses issue #884, should be merged after #1172
- Added middleware to check if email is invalid and require the user to change email address in this case
- Added management command to clean up unactivated users with invalid emails


accounts/management/commands/cleanup_unactivated_users.py
```diff
@@ -0,0 +1,37 @@
...
+import logging
+
+logger = logging.getLogger('console')
+
+
+class Command(BaseCommand):
+    help = 'Removes unactivated users with invalid emails'
+
+    def handle(self, *args, **options):
+        bounces = EmailBounce.objects.filter(type__in=EmailBounce.TYPES_INVALID, user__is_active=False)
```
ffont: We should add the `last_login__isnull=True` condition here to avoid matching users that were de-activated (e.g. after spam reports).

accounts/management/commands/cleanup_unactivated_users.py
```diff
@@ -0,0 +1,37 @@
...
+from django.core.management.base import BaseCommand
+from django.contrib.auth.models import User
+from accounts.models import EmailBounce
+import logging
+
+logger = logging.getLogger('console')
+
+
+class Command(BaseCommand):
+    help = 'Removes unactivated users with invalid emails'
```
ffont: If I understand correctly, the idea here is to remove users that were never activated and never logged in to Freesound. That is the case for users that registered with an invalid email address. Lets make this more clear in the help text.

freesound/middleware.py
```diff
@@ -119,3 +119,28 @@ def __call__(self, request):
...
         response = self.get_response(request)
         return response
+
+
+class UpdateEmailHandler(object):
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if request.user.is_authenticated \
```
ffont: In gdpr branch we added a util function to handle the "don't redirect" URLs. When eventually merging gdpr to master and master to this branch we should update that function to include `resetemail` and use it here.

templates/accounts/email_reset_form.html
```diff
@@ -4,7 +4,13 @@
 
 {% block content %}
 
-<h1>E-mail reset</h1>
+{% if required %}
```
ffont: Maybe we can use `django.contrib.messages` here so the formatting is the "standard" in Freesound.
It's a matter of importing `from django.contrib import messages` in `middleware` and adding a message before redirecting to email reset page in `UpdateEmailHandler`: 
```
messages.add_message(request, messages.INFO, "We have identified that some emails that we have sent to you didn't go through, thus it appears that your email address is not valid. Please update your email address to a working one to continue using Freesound")
```

freesound/middleware.py
philtgun: Yes, as I mentioned in the comment below :)

accounts/models.py
```diff
@@ -358,6 +358,12 @@ def delete_user(self, remove_sounds=False):
         else:
             Sound.objects.filter(user=self.user).update(is_index_dirty=True)
 
+    def has_content(self):
+        return Sound.objects.filter(user=self.user) or \
+               Post.objects.filter(author=self.user) or \
+               Comment.objects.filter(user=self.user)
+        # TODO: add more checks?
```
ffont: Yes we should add more checks:

- Packs
- Ratings
- Downloads
- PackDownloads
- Bookmarks
- APIV2 clients

Can we list all tables which have foreign keys to user_id to make sure we don't miss any important thing?
Then we can choose which things are "important" and should be included and which not. Or maybe we decide to include them all. In any case this should be properly documented in `has_content`.

philtgun: Current list of tables that are referencing user:
- `accounts_emailbounce`
- `accounts_oldusername`
- `accounts_profile`
- `accounts_resetemailrequest`
- `accounts_sameuser`
- `accounts_sameuser`
- `accounts_useremailsetting`
- `apiv2_extendedapplication`
- `auth_message`
- `comments_comment`
- `django_admin_log`
- `donations_donation`
- `favorites_favorite`
- `forum_post`
- `forum_subscription`
- `forum_thread`
- `fsmessages_message`
- `fsmessages_message`
- `general_akismetspam`
- `geotags_geotag`
- `oauth2_provider_accesstoken`
- `oauth2_provider_application`
- `oauth2_provider_grant`
- `oauth2_provider_refreshtoken`

ffont: I updated the list of "important" things I made with info from here. You removed some things from this list right? There is no `sounds_sound` table but I assume you removed it because it was already being considered right?

philtgun: My bad, I haven't included all of them. Now updated the list

ffont: Thanks, I edited again my list, I think checking these objects should be enough to consider user has "important" content.

PR #1193, opened 2018-06-19
Remove locals() from code

alastair:
Remove last set of `locals()` from template context parameters and replace with explicit dictionaries.

This round of commits affected some mails:
 * spam reports (manually tested, seems to be OK)
 * stream emails (not tested, but stepped through with @ffont)
 * Moderation emails (not tested, but emails are very simple)

Perhaps we can deploy and test this functionality in test before merging.


ffont:
I double checked stream emails and moderation emails in test and it works fine. Merging this PR...

Issue #1209, opened 2018-07-02
Apple Touch Icon


Superanton:
Could you add a "Apple Touch Icon" to Freesound?

You just have to add a file named "apple-touch-icon.png" to the root (152x152 pixel). 

I already created an icon:
![apple-touch-icon](https://user-images.githubusercontent.com/803655/42175318-51a53eec-7e25-11e8-8d64-8c0828e43797.png)

I could also create a pull request but I first want to get sure that there is not a reason against a touch icon. You could also add icons for Android.

bdejong:
This is the full list of all ios and android required resolutions, actually:

```<!-- iPhone(first generation or 2G), iPhone 3G, iPhone 3GS -->
<link rel="apple-touch-icon" sizes="57x57" href="touch-icon-iphone.png">
<!-- iPad and iPad mini @1x -->
<link rel="apple-touch-icon" sizes="76x76" href="touch-icon-ipad.png">
<!-- iPhone 4, iPhone 4s, iPhone 5, iPhone 5c, iPhone 5s, iPhone 6, iPhone 6s, iPhone 7, iPhone 7s, iPhone8 -->
<link rel="apple-touch-icon" sizes="120x120" href="touch-icon-iphone-retina.png">
<!-- iPad and iPad mini @2x -->
<link rel="apple-touch-icon" sizes="152x152" href="touch-icon-ipad-retina.png">
<!-- iPad Pro -->
<link rel="apple-touch-icon" sizes="167x167" href="touch-icon-ipad-pro.png">
<!-- iPhone X, iPhone 8 Plus, iPhone 7 Plus, iPhone 6s Plus, iPhone 6 Plus -->
<link rel="apple-touch-icon" sizes="180x180" href="touch-icon-iphone-6-plus.png">
<!-- Android Devices High Resolution -->
<link rel="icon" sizes="192x192" href="icon-hd.png">
<!-- Android Devices Normal Resolution -->
<link rel="icon" sizes="128x128" href="icon.png">
```

ffont:
To have icons for all sizes/os etc you need to have quite a number. I used in the past some online tools to do that and it works perfect. There is no specific reason why we have not done that for Freesound, we should definitely do it. I used a tool like this https://realfavicongenerator.net/
I'd by happy if @Superanton opens a pull request :)

ffont:
Hi @Superanton, I just merged and deployed your PR.
[...]

PR #1213, opened 2018-07-08
Add touch icons

Superanton:
I added icons for iOS/Android/Windows. See issue #1209.


ffont:
Hi, thanks for your contribution!
We internally discussed with the team that if we add favicon (which is a great idea) we should do it properly so there is support for all browsers/OS. As I mentioned in my [comment](https://github.com/MTG/freesound/issues/1209#issuecomment-402054800) in the issue, you could use [this website](https://realfavicongenerator.net/) to generate all needed favicon files and code. Would you mind updating the pull request with this?

Superanton:
Hey @ffont 
I used this website. :) So my commit doesn't only include icons for iOS and Mac. The icon is also supported on Android and Windows. I think all common browsers are supported.

ffont:
Thanks @Superanton I'll merge and see how it looks like :)

Issue #1200, opened 2018-06-21
Compute CRC using python library instead of external program


alastair:
When we compute sound CRC (for pack downloads) we use an external `crc32`. This catches all exceptions, and so isn't great if there is some kind of error: https://github.com/MTG/freesound/blob/e7670fc4483988124fe63edc41cf0c7cedd2f383/utils/sound_upload.py#L201

We should use the crc32 function available in the zlib module so that we don't have to rely on an external program: https://docs.python.org/2/library/zlib.html#zlib.crc32

Issue #1201, opened 2018-06-21
Set an upper limit of number of sounds in a pack


alastair:
We could add an upper limit of the number of sounds that can appear in a pack, because
 - uncompressing a folder with many files can make some operating systems/filebrowsers slow
 - such large packs consisting of wav files would be really large - we probably don't want to let people download so much

If we do this, ensure also in bulk upload that we inform users during validation if this limit is reached

bdejong:
Actually zip32 has a certain limit of file-size:

> The original .ZIP format had a 4 GiB limit on various things (uncompressed size of a file, compressed size of a file and total size of the archive), as well as a limit of 65535 entries in a ZIP archive.

alastair:
That's a good point that I didn't think about. Do we think that it's worth trying to support zip64, or should we just make sure we stay within the limits for zip32 for biggest compatibility?

bdejong:
> v1.1.6: Features: **Zip64 for large archives**, serial subrequests, UTF-8 filenames.

Looks like mod_zip 1.1.6 supports zip64. But, for example Windows can't deal with zip64 files by default. You need to use an unzipper like 7zip in order to be able to open the files. And I think the default unzip in macOS also does not deal with them...

So, I would say that having an upper limit and limit to zip32 would make sense...

PR #1221, opened 2018-07-18
Replace external crc32 with zlib crc call

philtgun:
Addresses #1200
Removed broad except because we want to know if something goes wrong

### DIFF
--- a/sounds/models.py
+++ b/sounds/models.py
@@ -60,6 +60,7 @@ import gearman
 import subprocess
 import datetime
 import json
+import zlib
 
 
 search_logger = logging.getLogger('search')
@@ -729,8 +730,11 @@ class Sound(SocialModel):
             self.save()
 
     def compute_crc(self, commit=True):
-        p = subprocess.Popen(["crc32", self.locations('path')], stdout=subprocess.PIPE)
-        self.crc = p.communicate()[0].split(" ")[0][:-1]
+        with open(self.locations('path'), "rb") as fp:
+            data = fp.read()
+        crc = zlib.crc32(data)
+        self.crc = '{:0>8x}'.format(crc & 0xffffffff)  # right aligned with zero-padding, width of 8 chars
+
         if commit:
             self.save()
 
--- a/utils/sound_upload.py
+++ b/utils/sound_upload.py
@@ -196,11 +196,8 @@ def create_sound(user, sound_fields, apiv2_client=None, process=True, remove_exi
         for moderator in moderators:
             invalidate_template_cache("user_header", moderator.id)
 
-    # 9 proces sound and packs
-    try:
-        sound.compute_crc()
-    except:
-        pass
+    # 9 process sound and packs
+    sound.compute_crc()
 
     if process:
         try:


ffont:
I did some testing and this seems to work great. When the file does not exist (it should always exist), the new `compute_crc` raises an Exception whereas the previous one set a "wrong" crc. I think it is fine to leave it with the exception so we will get notified if this happens and can take actions (it should never happen). I seem to remember there were some other concerts using `zlib` but I might be wrong. Perhaps something related to file size @alastair @bdejong ?

alastair:
The other zip issue was with the size of pack downloads - the basic zip standard only supports up to ~4gb archives (https://github.com/MTG/freesound/issues/1201), using zlib here for crc computation is not related.

sounds/models.py
```diff
@@ -729,8 +730,11 @@ def post_delete_comment(self, commit=True):
             self.save()
 
     def compute_crc(self, commit=True):
-        p = subprocess.Popen(["crc32", self.locations('path')], stdout=subprocess.PIPE)
-        self.crc = p.communicate()[0].split(" ")[0][:-1]
+        with open(self.locations('path'), "rb") as fp:
```
alastair: Mixing quotes? :/

sounds/models.py
```diff
@@ -729,8 +730,11 @@ def post_delete_comment(self, commit=True):
...
+            data = fp.read()
+        crc = zlib.crc32(data)
```
bdejong: this is not how you're supposed to use crc32 though...

the idea is:

```
...
with open(self.locations("path"), "rb") as f:
    for chunk in iter(lambda: f.read(4096), b''):
        crc = zlib.crc32(chunk, crc)
self.crc = '{:0>8x}'.format(crc & 0xffffffff)
...
```

bdejong: Otherwise you area loading the full file into memory...!

ffont: Nice catch, can you update this @philtgun ?

freesound/settings.py
```diff
@@ -199,6 +199,9 @@
 # Number of ratings of a sound to start showing average
 MIN_NUMBER_RATINGS = 3
 
+# Buffer size for CRC computation
+CRC_BUFFER_SIZE = 4096
```
bdejong: Does this really need to be a settings.py wide setting? This seems a bit weird.

sounds/models.py
```diff
@@ -730,9 +730,12 @@ def post_delete_comment(self, commit=True):
...
-        with open(self.locations('path'), "rb") as fp:
-            data = fp.read()
-        crc = zlib.crc32(data)
+        crc = 0
+        with open(self.locations('path'), 'rb') as fp:
+            data = fp.read(settings.CRC_BUFFER_SIZE)
+            while len(data) > 0:
+                crc = zlib.crc32(data, crc)
+                data = fp.read(settings.CRC_BUFFER_SIZE)
         self.crc = '{:0>8x}'.format(crc & 0xffffffff)  # right aligned with zero-padding, width of 8 chars
```
bdejong: I find this breaks the DRY principle... If you don't want to use the function iteration example I gave, then perhaps:

```python
while True:
        chunk = f.read(4096)
        if not chunk:
            break
        dostuff(chunk)
```

Makes more sense

freesound/settings.py
philtgun: I agree that maybe it is too global to put in into the settings.py, but don't we want to have it configurable? If not, is it ok to have 'magic' numbers in the code? I can always leave it as it was in your code example

ffont: I think it's fine to have it in `settings.py`. We'll have organize it a bit a some point but I'd argue that's fine.

sounds/models.py
philtgun: I wanted to make code a bit more understandable, but I can see your point. And `iter` is much better than `break` from `while True`.

Issue #1218, opened 2018-07-11
Add mroe information in spam email reports


bdejong:
The current comment spam email reads:

> Report for spammer/offensive user XXXXXX
> ----------------------------------------
> Hi there super admin!
> User pachok89 has been reported as a spammer or offensive at least three times by different users.
> Here you have a list of the items that have been marked as bad content:
> 
>     * comment: https://freesound.org/admin/comments/comment/XXXXXX/change/
>     * comment: https://freesound.org/admin/comments/comment/XXXXXX/change/
>     * comment: https://freesound.org/admin/comments/comment/XXXXXX/change/
> 
> If you can confirm this user is a spammer or offensive, please delete him following https://freesound.org/admin/auth/user/XXXXXX/delete/
> If the user is not a spammer or offensive, you can clean the reports by following https://freesound.org/people/XXXXXX/clear_flags/

What would be more useful is:

> Comment spam report for user: https://freesound.org/people/bram/
> Comments by user: https://freesound.org/people/bram/comments_by/
> Clear flags: https://freesound.org/people/bram/clear_flags/
>
> DANGER - Delete user: https://freesound.org/admin/auth/user/XXXXXX/actions/full_delete/

ffont:
Spam emails are not only for comments but can include other types of content (always for the same user). I think what would be even more useful is to include an excerpt of the reported comments/messages/posts in the email itself so in most cases there will be no need to even go to Freesound to make a decision.

Eg:

```
* comment: https://freesound.org/admin/comments/comment/XXXXXX/change/
    "bla bla bla bla ...."
* comment: https://freesound.org/admin/comments/comment/XXXXXX/change/
    "bla bla bla bla ...."
* forum post: https://freesound.org/xxx/
    "bla bla bla bla ...."
```

PR #1233, opened 2018-08-27
Add object previews in spam report emails

ffont:
Addresses https://github.com/MTG/freesound/issues/1218

* Spam report emails now include object preview in the body of the email 
* Small refactorings in user flag code
* Add unit tests

### DIFF
--- a/accounts/tests.py
+++ b/accounts/tests.py
@@ -30,7 +30,7 @@ from django.core.files.uploadedfile import InMemoryUploadedFile, SimpleUploadedF
 from django.core.management import call_command
 from django.core import mail
 from django.conf import settings
-from accounts.models import Profile, EmailPreferenceType, SameUser, ResetEmailRequest, OldUsername, EmailBounce
+from accounts.models import Profile, EmailPreferenceType, SameUser, ResetEmailRequest, OldUsername, EmailBounce, UserFlag
 from accounts.views import handle_uploaded_image
 from accounts.forms import FsPasswordResetForm, DeleteUserForm, UsernameField, RegistrationForm
 from accounts.management.commands.process_email_bounces import decode_idna_email
@@ -40,6 +40,7 @@ from utils.filesystem import File
 from tags.models import Tag
 from comments.models import Comment
 from forum.models import Thread, Post, Forum
+from messages.models import Message, MessageBody
 from tickets.models import Ticket
 from utils.mail import transform_unique_email, send_mail
 import accounts.models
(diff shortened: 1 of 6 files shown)


accounts/tests.py
```diff
@@ -1575,3 +1576,166 @@ def test_bulk_describe_state_closed(self):
...
+            resp = self.client.post(reverse('flag-user', kwargs={'username': self.spammer.username}), data={
+                'object_id': object.id,
+                'flag_type': flag_type,
+            })
+            self.assertEqual(resp.status_code, 200)
+            self.assertEqual(UserFlag.objects.count(), i + 1)  # Now we have more flags
+            self.assertEqual(UserFlag.objects.all().order_by('id')[i].reporting_user,
+                             reporter)  # Flag object created by reporter
+
+            if i == settings.USERFLAG_THRESHOLD_FOR_NOTIFICATION - 1:  # Last iteration
```
alastair: Instead of having all of these asserts and then the conditional, I'd be a little more manual about the checking.
for example, send 1 report, check that there is a UserFlag, check that the outbox is empty, then send all the other reports and check that the email has been triggered.

accounts/tests.py
```diff
@@ -1575,3 +1576,166 @@ def test_bulk_describe_state_closed(self):
...
+                self.assertEqual(len(mail.outbox), 1)  # Notification email sent
+                self.assertTrue("[freesound] Spam/offensive report for user" in mail.outbox[0].subject)
+                self.assertTrue("has been reported" in mail.outbox[0].body)
```
alastair: If this patch was to include the objects in the email, should we check for them here?

accounts/tests.py
```diff
@@ -1575,3 +1576,166 @@ def test_bulk_describe_state_closed(self):
...
+
+            if i == settings.USERFLAG_THRESHOLD_FOR_AUTOMATIC_BLOCKING - 1:  # Last iteration
+                self.assertEqual(len(mail.outbox), 2)  # New notification email sent
+                self.assertTrue("[freesound] Spam/offensive report for user" in mail.outbox[1].subject)
+                self.assertTrue("has been blocked" in mail.outbox[1].body)
+
+            else:
+                self.assertEqual(len(mail.outbox), 1)  # No new wmail sent
+
+        # Flag the object again and no new notification emails are sent
```
alastair: I would have this as a separate test, and only do it for one object type (if the logic is the same for all of them)

accounts/tests.py
```diff
@@ -1575,3 +1576,166 @@ def test_bulk_describe_state_closed(self):
...
+            self.assertEqual(UserFlag.objects.all().order_by('id')[i].reporting_user,
+                             reporter)  # Flag object created by reporter
+
+            if i == settings.USERFLAG_THRESHOLD_FOR_NOTIFICATION - 1:  # Last iteration
+                self.assertEqual(len(mail.outbox), 1)  # Notification email sent
+                self.assertTrue("[freesound] Spam/offensive report for user" in mail.outbox[0].subject)
+                self.assertTrue("has been reported" in mail.outbox[0].body)
+
+        # Continue flagging object until it reaches blocked state
+        for i in range(settings.USERFLAG_THRESHOLD_FOR_NOTIFICATION,
```
alastair: consider doing this as a separate test

accounts/tests.py
ffont: That's in fact what I do, but I agree the last check could be outside the loop and this might make things more clear. I'll update that!

accounts/tests.py
ffont: Yes we could but it would introduce some "if else" to get the content from the object and check if it's in the email body which would be a bit ugly. I think it is not really needed as if there is a problem retrieving the content the test will fail, and in any case when these emails are sent to the admins we will quickly identify the potential issue. I'd argue that for the sake of simplicity there is no need to add this checks.

accounts/tests.py
ffont: Hmm, what I'm testing here is that a first email is sent when a user has object(s) flagged by at least 3 users, and that a second email is sent when flags come from 6 distinct users (and the reported user is blocked). Then no more emails should be sent. That last bit of the test is to check that no more emails are sent so to me it is part of the same thing right? I keep on flagging user's content and checking if an email was sent or not.

accounts/tests.py
ffont: Ok, now I see the point. If this part is separated the previous one (your previous comment) could be separated as well. To me it makes sense that both kinds of reports are tested in the same test because it is part of the same logic. I'll keep it like that for now because I think it's fine as it is, but if you have some stronger argument which I did not think about I'll be happy to change it.

PR #1246, opened 2018-10-11
labels in advanced search allow clicking text, not just checkboxes

qubodup:
**Issue(s)**
none

**Description**
This allows to click text and not just the checkboxes in advanced search. A Usability improvement.
Video demo: https://youtu.be/fQ02W3SmBBo

**Deployment steps**:
I don't think so. hard reload in browser was enough for it to work on my end.

If it is possible to force that browsers get `all.css` anew, that would be good, as without the change in `all.css`, the advanced search form puts every option in its own block.

### DIFF
--- a/media/css/all.css
+++ b/media/css/all.css
@@ -241,17 +241,9 @@ form {
     margin-bottom: 10px;
 }
 
-form label {
-    display: block;
-}
-
 form input,form textarea {
     padding: 3px;
     font-size: 13px;
-}
-
-form#search_form  label {
-    display: inline;
 }
 
 #content textarea {
--- a/media/css/all.css
+++ b/media/css/all.css
@@ -249,6 +249,11 @@ form input,form textarea {
     padding: 3px;
     font-size: 13px;
 }
+
+form#search_form  label {
+    display: inline;
+}
+
 #content textarea {
     width: 530px;
 }
--- a/templates/accounts/registration.html
+++ b/templates/accounts/registration.html
@@ -3,7 +3,7 @@
 {% block title %}registration{% endblock %}
 
 {% block section_content %}
-<style> label[for=id_accepted_tos] {display: inline} #username_check, #check_success, #spinner { display:none; color: #bd2d33; font-style: italic; } #check_success {color:green;}</style>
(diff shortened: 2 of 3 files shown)


media/css/all.css
```diff
@@ -249,6 +249,11 @@ form input,form textarea {
     padding: 3px;
     font-size: 13px;
 }
+
+form#search_form  label {
```
alastair: I'm looking into other places that we have a label inside a form. There's a rule earlier in this file which make this a `display: block`, however I've found a few places where this is overridden _again_ to make it inline. Looking at the list of template files using `<label>`:

* templates/follow/stream.html: overridden to inline-block here: https://github.com/MTG/freesound/blob/032cd6f03b1e500e28899d1d6b9a2e35a7d19d3d/media/css/all.css#L4106
* templates/api/minimal_registration.html: doesn't use all.css
* templates/accounts/recursive_file.html: doesn't use `form label` selector
* templates/accounts/registration.html: Uses display: block
* templates/search/advanced_search_form.html: your change, uses inline
* templates/forum/_section.html: Uses display: block

I wonder if it makes more sense to explicitly only use display: block on the pages where it's needed (only registration page, I think that forum advanced search can be inline or block-inline). As the default display for a label is inline anyway, we wouldn't need this explicit inline for the search form.

Does this make sense? Are you happy to try and make this change, or would you prefer that I do it?

qubodup:
Hi, thanks for responding. Sorry I didn't check whether this affects other sites. Is the #search_form not enough to avoid side effects? I probably won't have time to add anything to this before next year I'm afraid. If you have a solution, it'd be great if you could use it, thanks!

alastair:
OK, no problem!
You're right that #search_form is enough to avoid other side-effects, but it has the result of adding _another_ method of doing the same thing that is done in a different way on other pages.
I'll make these changes myself, thanks for the initial work.

ffont:
Thanks @qubodup for the comments and PR. @alastair did these changes ever happen?

alastair:
I'm going to work on it this week, I was busy with project meeting

alastair:
I've made some changes to this to simplify our rules. @ffont, can you merge it if you're happy?

Issue #1263, opened 2018-12-12
Admin-only view on sound flags


bdejong:
It would be great to have an admin-view for sound flags as administrating them through emails is rather tedious work that requires a lot of clicks. A very simple page like this would help a lot in speeding up the process:

# Offending

Table with these cells, per row, ordered oldest first

- Date
- Date
- Reporting user (either link to username or email)
- Report (text)
- Link to sound
- Link to uploader of sound
- X (delete, via JS, no confirmation)

# Illegal

Table with these cells, per row, ordered oldest first

- Date
- Reporting user (either link to username or email)
- Report (text)
- Link to sound
- Link to uploader of sound
- X (delete, via JS, no confirmation)

# Other

Table with these cells, per row, ordered oldest first

- Date
- Reporting user (either link to username or email)
- Report (text)
- Link to sound
- Link to uploader of sound
- X (delete, via JS, no confirmation)

ffont:
We could implement this easily using the Django admin I guess?

bdejong:
> We could implement this easily using the Django admin I guess?

I have no idea :-D I would guess that things like javascript-to-delete-things might be more complicated to do in the admin than in just a plain old HTML page

ffont:
It would not be using javascript to delete, but Django admin allows you to select a number of items and then delete them in bulk. I guess that workflow would already be a significant improvement? And adding this tables using the Django admin might be ~20 lines of code (maybe).

bdejong:
True... Perhaps not a bad idea...!

bdejong:
I would say:
– list the ID in the list and make that clickable, clicking on the username is weird
- make the reporting username (if it’s there) clickable to go to the user, make the email address clickable (if it’s there)
- add a clickable "uploader" username
- add a clickable link to the sound
- add the reason in the list, replace newlines with " | " and perhaps trim it to a certain length
- provide one filter on the right-hand side to filter either “offending”, “illegal”, “other”

ffont:
Pending to do:

- Double length of reason summary
- Display sound name instead of sound ID (also truncated)

If we truncate in two places we might need to find the Django util function that does it or implement one ourselves. Hmm wait a sec, maybe we can import that https://github.com/MTG/freesound/blob/fb09da014c3aa30b617ed8fd5dbe92961cb3f23e/general/templatetags/util.py#L35 ?

bdejong:
[...]

PR #1267, opened 2018-12-20
Improve flag management in admin

ffont:
**Issue(s)**
https://github.com/MTG/freesound/issues/1263

**Description**
Added more capabilities to Django admin page for managing sound flags as described in the issue. now this will hopefully be good enough so that no extra pages are needed to manage flags.

Here's a screenshot of how it looks like:
screenshot 2018-12-20 at 15 46 41

### DIFF
--- a/sounds/admin.py
+++ b/sounds/admin.py
@@ -21,6 +21,7 @@
 #
 
 from django.contrib import admin
+from django.urls import reverse
 from sounds.models import License, Sound, Pack, Flag, DeletedSound, SoundOfTheDay, BulkUploadProgress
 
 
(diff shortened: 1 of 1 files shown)


sounds/admin.py
```diff
@@ -61,7 +62,36 @@ class PackAdmin(admin.ModelAdmin):
...
+    sound_uploader_link.allow_tags = True
+
+    def sound_link(self, obj):
+        return '<a href="{0}" target="_blank">{1}</a>'.format(reverse('short-sound-link', args=[obj.sound_id]),
+                                                              obj.sound_id)
+    sound_link.allow_tags = True
+
+    def reason_summary(self, obj):
+        reason_no_newlines = obj.reason.replace('\n', '|')
+        return reason_no_newlines if len(reason_no_newlines) < 50 else reason_no_newlines[:50] + '...'
```
alastair: isn't there a django function to do this?

ffont: There is a templatetag and I guess there must be a function somewhere but I did not find it after a 2 minute search and it was so easy to implement.

PR #1271, opened 2019-01-27
Progress on browse sounds page

giuband:
**Description**
Updated the layout of the browse sounds page. In general it's missing a bit of interactivity, but you might want to start adapting it already.

**Deployment steps**:
Choose the option "Browse" when asked which page to start at running `npm start`.


ffont:
Hi @giuband,
Thanks for the PR! Here are some comments. Take into account that this is using Safari (just in case). What I did is to compare the design renders given by the designer with what I see here. There are probably many things that you have not worked on yet and I'm listing them anyway. I hope the feedback is useful:

- Pages are not prepared for mobile, should we set a minimum width for now so the layout is more under control?
- Username link should have hand icon
- Similar sounds icon should have hand icon
- Bookmark sound icon should have hand icon
- Empty checkboxes do not appear
- Checkbox labels too separated from checkbox
- "Sort by" options missing
- Sound display textual description color should be darker
- Star icons too much separated (stars should be closer)
- Tag border a bit too dark?
- Missing paginator [example here](https://projects.invisionapp.com/d/main#/console/11781900/257443463/preview)
- Too much spacing between user icon, username and date? (check spacing generally as I see more spacing than in the designs)
- Waveform looks smaller compared to the design?
- In filters section you'll need to add the "related tags" type of filter (note the fading that the designer added) [example here](https://projects.invisionapp.com/d/main#/console/11781900/257443463/preview)
- Missing loop button in sound player
- Play button in sound player has some gray background that should be transparent
- Start browsing text should have more weight
- The cross icon (to remove query) should only appear when something has been typed and should be gray. Also it should appear next to the text. Also hand icon should be shown on howvering and clicking should clear input contents (btw: is this kind of behaviour supposed to be implemented in the template?) [example here](https://projects.invisionapp.com/d/main#/console/11781900/255631838/preview)
- The cursor for typing search terms should appear in the left of the placeholder instead of the middle. See [example here](https://projects.invisionapp.com/d/main#/console/11781900/257441522/preview).
- Hovering tags should change their background to black (invert colors) instead of have the shadow. See [example here](https://projects.invisionapp.com/d/main#/console/11781900/255631838/preview).
- The page navbar (is it called like that?) is supposed to change when you scroll down. The version with the full logo is only shown in the front page when you are at the top ([this one](https://projects.invisionapp.com/d/main#/console/11781900/254102154/preview)). In other pages the navbar shown is the one with the reduced logo (only waveform) and the search bar, and the "donate" and "upload sound" buttons are inverted (no background) ([like here](https://projects.invisionapp.com/d/main#/console/11781900/257443464/preview)). In the search page, the search bar from the navbar is not shown ([like in here](https://projects.invisionapp.com/d/main#/console/11781900/255631838/preview)).

You'll need to check the different renders created by the designer as, for example, the "related tags" filter appears in some renders but not in others.

giuband:
@ffont I added a build script (`npm run build`). Should we merge the PR and tackle the feedback later? Otherwise it might stay open for a while (actually I would prefer to tackle the feedback at the very end and first develop the big blocks). We could also set up a trello board where to store all your feedback to not lose track of it.

Issue #654, opened 2015-02-12
Better log of sounds that fail processing


ffont:
We should have a better logging of sounds that fail processing, e.g. showing the extension of the file (which in most cases explains why the sound failed). Complementarily, we could periodically reprocess those sounds that failed in the past. In some cases processing fails because of memory problems in the processing machines, this would be fixed also by restarting the workers periodically (related with Issue #651).

ffont:
Closing as this has been addressed in https://github.com/MTG/freesound/pull/1289

Issue #638, opened 2014-11-04
Improve processing and analysis function


ffont:
Processing and analysis functions should be improved in several ways:
- In some case, temporary files created for processing and analysing sounds are not properly cleaned up
- The limit of 100 mb for analysing sounds is checked on the filesize of the original files, which might be compressed formats like mp3. This check should be performed either on the duration field of the sound or on the size computed after ffmpeg converts to wav
- Sounds should be sent to process and analyze with high priority by default. Then we should define commands to reprocess and reanalyze in bulk, setting low priority
- Sounds that can con be analysed because the file size is to big should be set to a new analysis state such as "SK" - "skipped" rather than "FA" - "failed"

ffont:
Closing as this has been addressed in https://github.com/MTG/freesound/pull/1289

Issue #595, opened 2014-03-13
Fix status update SQL


alastair:
In https://github.com/MTG/freesound/blob/master/sounds/models.py#L337 we perform immediate updates of sound status values so that other workers don't clobber status.
This SQL needs to be constructed using proper attribute positioning, not by manually constructing strings.

alastair:
This was fixed in #1289, thanks @ffont for finally addressing it!

PR #1289, opened 2019-04-08
Processing improvements

ffont:
**Issue(s)**
https://github.com/MTG/freesound/issues/654
https://github.com/MTG/freesound/issues/638
https://github.com/MTG/freesound/issues/595

**Description**
This is a big PR (sorry) with a number of improvements in processing. This is a list of improvements:

- Better logging of processing/analysis output, including storing individual sound processing logs in `Sound.processing_log` field (see `NOTE` at the end).
- Re-structured analysis and processing code into classes which both inherit from `FreesoundAudioProcessorBase`. In this way less code is duplicated.
- Better handling of errors in processing/analysis functions, including a general timeout for processing/analysis workers.
- Better handling of cleaning up tmp files after processing/analysis. Now tmp directories are cleaned in processing/analysis functions as well as in the worker. Also, instead of cleaning files individually these are all added to a directory and then the whole directory is cleaned.
- Workers check if there is free space in disk before running processing/analysis. Otherwise they log the issue and cancel the job.
- Simplified `gm_worker_processing.py` code by having separate functions for processing and analysis.
- Simplified `gm_worker_processing.py` code by retrieving sound objects using standard `Sound.objects.get(id=X)` and without retrials if DB connection fails. Current code was prepared to handle DB connection problems because in a previous infrastructure we had this was a common problem. With current infrastructure this should not be an issue so I decided to simplify the code.
- Added minimal support for newer Essentia `FreesoundExtractor` which uses different naming conventions for statistics and frames files. Current Freesound analysis still used old extractor, but now the code is better prepared for the new ones.
- Removed the methods in `Sound` model for setting single fields using raw SQL. Now we use `Sound.save(update_fields=X)` instead (and only when necessary).

**Deployment steps**:
Standard deploy to website and also to processing machines.
After deployment, we should try analyzing/processing large number of sounds (~10000) and inspect the results, make sure that sounds don't fail processing/analysis more than before and everything is logged/works as expected. After that we can re-process all sounds to create new display images for BW frontend. 

`NOTE`: Analysis output logs are not stored in DB as the way analysis is handled is still pending a deeper refactoring re how we store the results. Some months ago we added the `SoundAnalysis` object to store AudioCommons analysis results, but this was designed to be able to store other analysis results in the future. At some point we should start using those for the "standard analysis" and then we could store analysis out log in that object.


sounds/models.py
```diff
@@ -589,56 +591,34 @@ def set_license(self, new_license):
...
+        self.save(update_fields=['similarity_state'])
 
     def set_audio_info_fields(self, info):
-        field_names = ['samplerate', 'bitrate', 'bitdepth', 'channels', 'duration']
-        field_values = [[field, info[field], False] for field in field_names]
-        self.set_fields(field_values)
+        allowed_fields = ['samplerate', 'bitrate', 'bitdepth', 'channels', 'duration']
+        for field_name, value in info.items():
+            if field_name in allowed_fields:
+                self.__setattr__(field_name, value)
```
alastair: try `setattr(self, field_name, value)` - I guess this will work in django

sounds/models.py
```diff
@@ -589,56 +591,34 @@ def set_license(self, new_license):
...
+        self.save(update_fields=list(info.keys()))
```
alastair: this should be the intersection of info.keys and allowed_fields

sounds/models.py
```diff
@@ -589,56 +591,34 @@ def set_license(self, new_license):
...
-
-    def set_moderation_date(self, date):
-        self.set_single_field('moderation_date', str(date))  # Date should be datetime object
-
-    def set_original_path(self, path):
-        self.set_single_field('original_path', path)
+        self.similarity_state = state
+        self.save(update_fields=['similarity_state'])
 
     def set_audio_info_fields(self, info):
```
alastair: this method is only used once, so consider refactoring it to pass all fields in as arguments to make it more explicit

accounts/tests.py
```diff
@@ -638,10 +639,7 @@ def test_select_uploaded_files_to_describe(self):
         self.client.login(username='testuser', password='testpass')
         user_upload_path = settings.UPLOADS_PATH + '/%i/' % user.id
```
alastair: I note that this will create a temp folder and not clean up after the test. Is this something that we could clean up in a teardown method, or at the end of the test?

freesound/settings.py
```diff
@@ -281,9 +281,12 @@
 
 SESSION_COOKIE_DOMAIN = None # leave this until you know what you are doing
 
-# Sound analysis properties
+# Sound analysis/processing worker properties
+WORKER_TIMEOUT = 5 * 60  # General timeout for processing/analysis workers to avoid to hang forever
```
alastair: small change, it'd be nice to see comments before the line that they affect. For this one, indicate that the time is in seconds

requirements.txt
```diff
@@ -44,3 +40,9 @@ ConcurrentLogHandler==0.9.1
 freezegun==0.3.9
 boto3==1.7.10
 xlrd==1.1.0
+numpy==1.9.0
+scikits.audiolab==0.11.0
+#apt-get install libyaml-dev (to allow the cloader)
```
alastair: This makes sense to have in a separate INSTALL or README file instead of in requirements

utils/filesystem.py
```diff
@@ -92,9 +93,16 @@ def remove_directory_if_empty(path):
 
 
 def create_directories(path):
```
alastair: just a thought here - python 3 has `os.makedirs(path, exist_ok=True)`. If we made this method with the same signature then it'd be really easy to switch to it when we change

utils/filesystem.py
```diff
@@ -92,9 +93,16 @@ def remove_directory_if_empty(path):
...
-            if exc.errno != errno.EEXIST:
-                raise
+    try:
+        os.makedirs(path)
+    except OSError as exc:
+        if exc.errno != errno.EEXIST:
+            raise  # Ignore exception if directory already existing
+
+
+def remove_directory(path):
```
alastair: Perhaps we can make this name a bit clearer - it's not clear that it will recursively remove a directory, or ignore errors.
Can we add simple docstrings to the two directory methods?

sounds/management/commands/gm_worker_processing.py
```diff
@@ -49,93 +90,74 @@ def add_arguments(self, parser):
...
+                log_error("Failed analyizing sound %s" % sound_id)
+
+        except WorkerException as e:
+            log_error('%s' % e)
+
+        except Exception as e:
+            log_error('Unexpected error while analyzing sound %s' % e)
 
-    def task_process_x(self, gearman_worker, gearman_job, func):
+        cancel_timeout_alarm()
```
alastair: In order to avoid passing `None` to `cleanup_tmp_directory`, perhaps we could use try/finally here?
It might involve making the out of disk space exception its own exception instead of `WorkerException`. Alternatively, something like a temp dir context manager might work: https://stackoverflow.com/questions/6884991/how-to-delete-dir-created-by-python-tempfile-mkdtemp
What I"m not sure about here is if mkdtemp can return some kind of exception (it's likely), in which case we probably still need to catch it.

utils/audioprocessing/freesound_audio_analysis.py
```diff
@@ -18,143 +18,78 @@
...
-
 logger = logging.getLogger("processing")
 
 
-def analyze(sound):
+class FreesoundAudioAnalyzer(FreesoundAudioProcessorBase):
 
-    def alarm_handler(signum, frame):
-        raise AudioProcessingException("timeout while waiting for Essentia")
+    def failure(self, message, error=None, failure_state="FA"):
```
alastair: can the name of this be more like a verb? "set failure" or "set processing failed"?

utils/audioprocessing/freesound_audio_analysis.py
```diff
@@ -18,143 +18,78 @@
...
+            create_directories(os.path.dirname(frames_path))
+
+            # Run Essentia's FreesoundExtractor analsyis
+            audioprocessing.analyze_using_essentia(
+                settings.ESSENTIA_EXECUTABLE, tmp_wavefile, os.path.join(self.tmp_directory, 'ess_%i' % self.sound.id),
+                essentia_profile_path=settings.ESSENTIA_PROFILE_FILE_PATH)
+
+            # Move essentia output files to analysis data directory
+            if settings.ESSENTIA_PROFILE_FILE_PATH:
+                # Never versions of FreesoundExtractor using profile file use a different naming convention
```
alastair: newer?

utils/audioprocessing/freesound_audio_analysis.py
```diff
@@ -18,143 +18,78 @@
...
-            failure('could not find file with path %s' % sound_path)
-            return False
+                shutil.move(os.path.join(self.tmp_directory, 'ess_%i_statistics.yaml' % self.sound.id), statistics_path)
+                shutil.move(os.path.join(self.tmp_directory, 'ess_%i_frames.json' % self.sound.id), frames_path)
+
+            self.log_info("created analysis files with FreesoundExtractor: %s, %s" % (statistics_path, frames_path))
+
+            # Change sound analysis and similarity states
+            self.sound.set_analysis_state('OK')
+            self.sound.set_similarity_state('PE')  # Set similarity to PE so sound will get indexed to Gaia
```
alastair: sent to gaia?
indexed by gaia?
Can we move this comment to the line above?

utils/audioprocessing/freesound_audio_analysis.py
```diff
@@ -18,143 +18,78 @@
...
-            audioprocessing.convert_using_ffmpeg(sound_path, tmp_wavefile, mono_out=True)
-            to_cleanup.append(tmp_wavefile)
-            success("converted to PCM: " + tmp_wavefile)
         except AudioProcessingException as e:
-            failure("conversion to PCM failed", e)
-            return False
-        except IOError as e:
-            # Could not create tmp file
-            failure("could not create tmp_wavefile file", e)
+            self.failure(e)
```
alastair: the signature of failure is (message, error). Is this parameter the message or the error (if the message, perhaps it should be `str(e)`?)

utils/audioprocessing/freesound_audio_analysis.py
```diff
@@ -18,143 +18,78 @@
...
             return False
-        except Exception as e:
-            failure("unhandled exception while converting to PCM", e)
+        except (Exception, OSError) as e:
```
alastair: why do you need OSError? Isn't it a subclass of exception?

utils/audioprocessing/processing.py
```diff
@@ -589,3 +589,19 @@ def alarm_handler(signum, frame):
...
     if process.returncode != 0 or not os.path.exists(output_filename):
         raise AudioProcessingException(stdout)
+
+
+def analyze_using_essentia(essentia_executable_path, input_filename, output_filename_base, essentia_profile_path=None):
+    """
+    runs Essentia's FreesoundExtractor analsyis
+    """
+    essentia_dir = os.path.dirname(os.path.abspath(essentia_executable_path))
+    os.chdir(essentia_dir)
```
alastair: I'm not sure this is necessary - why can't you run it from any directory? as long as the binary path and profile path are absolute paths (which I thinkthey should be)

utils/test.py
```diff
@@ -0,0 +1,99 @@
+#
```
alastair: I don't like having both `utils/test.py` and `utils/tests.py`

utils/tests.py
```diff
@@ -565,3 +560,370 @@ def test_bulk_describe_from_csv(self):
...
+def convert_to_pcm_mock_create_file(input_filename, output_filename):
+    create_test_files(paths=[output_filename], n_bytes=2048)
+    return True
+
+
+def convert_to_pcm_mock_fail(input_filename, output_filename):
+    raise AudioProcessingException("failed converting to pcm")
+
+
+def stereofy_mock(stereofy_executble_path, input_filename, output_filename):
```
alastair: executable

utils/tests.py
```diff
@@ -565,3 +560,370 @@ def test_bulk_describe_from_csv(self):
...
+    return dict(
+        duration=123.5,
+        channels=2,
+        samplerate=44100,
+        bitrate=128,
+        bitdepth=16)
+
+
+def stereofy_mock_fail(stereofy_executble_path, input_filename, output_filename):
```
alastair: executable

utils/tests.py
```diff
@@ -565,3 +560,370 @@ def test_bulk_describe_from_csv(self):
...
+def analyze_using_essentia_mock_fail(essentia_executable_path, input_filename, output_filename_base, **kwargs):
+    raise AudioProcessingException("essentia extractor returned an error")
+
+
+class AudioProcessingTestCase(TestCase):
+
+    fixtures = ['initial_data']
+
+    def pre_test(self, create_sound_file=True):
+        # Do some stuff which needs to be carried out right before each test
```
alastair: does this have to be done in all tests? why is it not part of setUp?

utils/tests.py
```diff
@@ -565,3 +560,370 @@ def test_bulk_describe_from_csv(self):
...
+        with self.assertRaises(AudioProcessingException) as cm:
+            FreesoundAudioAnalyzer(sound_id=999)
+        exc = cm.exception
+        self.assertIn('did not find Sound object', exc.message)
+
+    @override_settings(USE_PREVIEWS_WHEN_ORIGINAL_FILES_MISSING=False)
+    def test_sound_path_does_not_exist(self):
+        tmp_directory = self.pre_test(create_sound_file=False)
+        result = FreesoundAudioAnalyzer(sound_id=Sound.objects.first().id, tmp_directory=tmp_directory).analyze()
+        # analysis will fail because sound does not exist
```
alastair: clarify comment. Sound object exists, but the audio file doesn't exist?

utils/tests.py
```diff
@@ -565,3 +560,370 @@ def test_bulk_describe_from_csv(self):
...
+        self.assertFalse(result)  # Analysis failed, returning False
```
alastair: I don't think that this `# Analysis failed, returning False` is useful any time that you have it

utils/tests.py
```diff
@@ -565,3 +560,370 @@ def test_bulk_describe_from_csv(self):
...
+        self.user = user
+
+    def test_sound_object_does_not_exist(self):
+        with self.assertRaises(AudioProcessingException) as cm:
+            FreesoundAudioAnalyzer(sound_id=999)
+        exc = cm.exception
+        self.assertIn('did not find Sound object', exc.message)
+
+    @override_settings(USE_PREVIEWS_WHEN_ORIGINAL_FILES_MISSING=False)
+    def test_sound_path_does_not_exist(self):
```
alastair: It'd be good to have basic docstrings for all these tests describing the thing being tested in a bit more detail than the function name

utils/tests.py
```diff
@@ -565,3 +560,370 @@ def test_bulk_describe_from_csv(self):
...
+        self.sound = sounds[0]
+        self.user = user
+
+    def test_sound_object_does_not_exist(self):
+        with self.assertRaises(AudioProcessingException) as cm:
+            FreesoundAudioAnalyzer(sound_id=999)
+        exc = cm.exception
+        self.assertIn('did not find Sound object', exc.message)
+
+    @override_settings(USE_PREVIEWS_WHEN_ORIGINAL_FILES_MISSING=False)
```
alastair: Did we decide to make this setting default to True? (only in test?). It feels to me like it should be always False and only True in development if we want it?

utils/tests.py
```diff
@@ -565,3 +560,370 @@ def test_bulk_describe_from_csv(self):
...
+        result = FreesoundAudioAnalyzer(sound_id=Sound.objects.first().id, tmp_directory=tmp_directory).analyze()
+        # analysis will fail because sound does not exist
+        self.assertFalse(result)  # Analysis failed, returning False
+        self.sound.refresh_from_db()
+        self.assertEqual(self.sound.analysis_state, "FA")
+        self.assertFalse(os.path.exists(tmp_directory))
+
+    @mock.patch('utils.audioprocessing.processing.convert_to_pcm', side_effect=convert_to_pcm_mock_fail)
+    @override_settings(USE_PREVIEWS_WHEN_ORIGINAL_FILES_MISSING=False)
+    @override_settings(SOUNDS_PATH=tempfile.mkdtemp())
```
alastair: how is this directory cleaned up?

utils/tests.py
```diff
@@ -565,3 +560,370 @@ def test_bulk_describe_from_csv(self):
...
+        self.assertFalse(os.path.exists(tmp_directory))
+
+    @mock.patch('utils.audioprocessing.processing.convert_to_pcm', side_effect=convert_to_pcm_mock_create_file)
+    @override_settings(USE_PREVIEWS_WHEN_ORIGINAL_FILES_MISSING=False)
+    @override_settings(SOUNDS_PATH=tempfile.mkdtemp())
+    @override_settings(MAX_FILESIZE_FOR_ANALYSIS=1024)
+    def test_big_pcm_file_is_not_analyzed(self, *args):
+        tmp_directory = self.pre_test()
+        result = FreesoundAudioAnalyzer(sound_id=Sound.objects.first().id, tmp_directory=tmp_directory).analyze()
+        # analysis will skip as generated PCM filesize is biggen than MAX_FILESIZE_FOR_ANALYSIS
```
alastair: bigger

utils/test.py
ffont: Rename `utils/test.py` to `utils/test_helpers.py`.
Moved `utils/tests.py` to `tests` module inside `utils` module.
bc130f45c713bf47468afc54be2bab16c392b88a

accounts/tests.py
ffont: Addressed in 1a0e0b5642771df0cc1c05aef62c300d1b15d068

freesound/settings.py
ffont: Addressed in 5b9d860ce7beaed0ffa21f493d76239c47905013

requirements.txt
ffont: I agree, it is here for reference and because these should be installed before some of the python requirements. This will become clear when we move to Docker, and we currently have no up to date place where to put that information so I prefer to keep it like that for now. Nevertheless I added a small comment in the file for clarification 446b98f2536fc0d42d19b9d24bd6e611c2b41423.

sounds/management/commands/gm_worker_processing.py
```diff
@@ -18,24 +18,65 @@
...
-from utils.audioprocessing.freesound_audio_processing import process
+from utils.audioprocessing.freesound_audio_analysis import FreesoundAudioAnalyzer
+from utils.audioprocessing.freesound_audio_processing import FreesoundAudioProcessor
+from utils.filesystem import remove_directory
+
+logger = logging.getLogger("processing")
+logger_error = logging.getLogger("processing_errors")
+
+
+class WorkerException(Exception):
```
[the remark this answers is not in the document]
ffont: addressed in 99f859e2999fd9e15b993bec72283541d28c2a34, 3f7aa00e6971df5b0b74c71c6cc289628ce85151

sounds/management/commands/gm_worker_processing.py
```diff
@@ -18,24 +18,65 @@
...
+
+def log_error(message):
+    logger.error(message)
+    logger_error.info(message)
+
 
-logger = logging.getLogger("console")
+def check_if_free_space(directory='/tmp/', min_disk_space_percentage=0.05):
+    """
+    Raises an Exception if the perfectage of free disk space in the volume of the given 'directory' is lower
```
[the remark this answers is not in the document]
ffont: Addressed in 3f7aa00

sounds/management/commands/gm_worker_processing.py
```diff
@@ -49,93 +90,74 @@ def add_arguments(self, parser):
...
+        job_data = json.loads(gearman_job.data)
+        sound_id = job_data['sound_id']
+        set_timeout_alarm(settings.WORKER_TIMEOUT, 'Analysis of sound %s timed out' % sound_id)
 
-    def task_process_sound(self, gearman_worker, gearman_job):
-        return self.task_process_x(gearman_worker, gearman_job, process)
+        logger.info("---- Starting analysis of sound with id %s ----" % sound_id)
+        try:
+            check_if_free_space()
+            tmp_directory = tempfile.mkdtemp(prefix='analysis_%s_' % sound_id)
```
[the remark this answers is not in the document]
ffont: fe8d87b8ff84937f209f4eb04e720caa8e3ba474

sounds/management/commands/gm_worker_processing.py
```diff
@@ -18,24 +18,65 @@
...
+
 
-logger = logging.getLogger("console")
+def check_if_free_space(directory='/tmp/', min_disk_space_percentage=0.05):
+    """
+    Raises an Exception if the perfectage of free disk space in the volume of the given 'directory' is lower
+    than 'min_disk_space_percentage'.
+    """
+    stats = os.statvfs(directory)
+    percetage_free = stats.f_bfree * 1.0 / stats.f_blocks
```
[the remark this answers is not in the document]
ffont: fe8d87b8ff84937f209f4eb04e720caa8e3ba474

sounds/management/commands/gm_worker_processing.py
```diff
@@ -49,93 +90,74 @@ def add_arguments(self, parser):
...
+            check_if_free_space()
+            tmp_directory = tempfile.mkdtemp(prefix='analysis_%s_' % sound_id)
+            result = FreesoundAudioAnalyzer(sound_id=sound_id, tmp_directory=tmp_directory).analyze()
+            if result:
+                logger.info("Successfully analyzed sound %s" % sound_id)
+            else:
+                log_error("Failed analyizing sound %s" % sound_id)
+
+        except WorkerException as e:
+            log_error('%s' % e)
```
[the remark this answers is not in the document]
ffont: 00aedbe2e75e81ab9f8d95ffb3c792e2b3912497

sounds/management/commands/gm_worker_processing.py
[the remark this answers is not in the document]
ffont: 613a4c9b304d1d4a810854e58711d4854a90afe1

sounds/management/commands/gm_worker_processing.py
```diff
@@ -18,24 +18,65 @@
...
+    signal.signal(signal.SIGALRM, alarm_handler)
+    signal.alarm(time)
+
+
+def cancel_timeout_alarm():
+    signal.alarm(0)
+
+
+def cleanup_tmp_directory(tmp_directory):
+    if tmp_directory is not None:
```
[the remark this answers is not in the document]
ffont: function removed in 613a4c9b304d1d4a810854e58711d4854a90afe1

sounds/management/commands/gm_worker_processing.py
ffont: After 613a4c9b304d1d4a810854e58711d4854a90afe1 we use a context manager which simplified the cleaning significantly

sounds/models.py
```diff
@@ -665,25 +657,24 @@ def change_moderation_state(self, new_state, commit=True, do_not_update_related_
 
         self.invalidate_template_caches()
 
-    def change_processing_state(self, new_state, commit=True, use_set_instead_of_save=False):
+    def change_processing_state(self, new_state, processing_log=None, commit=True):
```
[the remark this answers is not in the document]
ffont: Addressed in ca52aefdd68d2205a9c327524b7407a5f7a8e4c4
Unneeded argument removed

sounds/models.py
```diff
@@ -665,25 +657,24 @@ def change_moderation_state(self, new_state, commit=True, do_not_update_related_
...
         """
         Change the processing state of a sound and perform related tasks such as set the sound as index dirty if
-        required. The 'use_set_instead_of_save' can be used to directly set the change of state in the db as an
-        update command without affecting other fields of the Sound model. This is needed when the processing tasks
-        change the processing state of the sound to avoid potential collisions when saving the whole object.
+        required. Only the fields that are changed are saved to the object. This is needed when the processing tasks
+        change the processing state of the sound to avoid potential collisions when saving the whole object. We do not
```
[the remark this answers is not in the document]
ffont: Addressed in ca52aefdd68d2205a9c327524b7407a5f7a8e4c4

sounds/models.py
```diff
@@ -589,56 +591,46 @@ def set_license(self, new_license):
...
-                             (field[0], (field[1] if not field[2] else ("'%s'" % field[1])))) for field in fields])
-        query += " WHERE id = %s" % self.id
-        with transaction.atomic():
-            cursor = connection.cursor()
-            cursor.execute(query)
-
-    def set_processing_state(self, state):
-        self.set_single_field('processing_state', state)
+    # N.B. The set_xxx functions below are used in the distributed processing and other parts of the app where we only
+    # want to save an individual field of the model to prevent overwritting other model fields.
```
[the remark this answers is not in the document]
ffont: Addressed in 12bf9486d495a1099bbf97ee52e74c993d4a951c

sounds/models.py
```diff
@@ -665,25 +657,24 @@ def change_moderation_state(self, new_state, commit=True, do_not_update_related_
...
-                self.set_processing_state(new_state)
-                self.set_processing_date(datetime.datetime.now())
-            else:
-                self.processing_state = new_state
-                self.processing_date = datetime.datetime.now()
-                if commit:
-                    self.save()
+            self.processing_state = new_state
+            self.processing_date = datetime.datetime.now()
+            self.processing_log = processing_log
```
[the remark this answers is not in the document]
ffont: Yes, this is all expected behaviour

utils/audioprocessing/freesound_audio_analysis.py
ffont: Addressed in 89e30d00ad951a2dfd2a69cb07fe65e3d6d2d748

utils/audioprocessing/processing.py
ffont: addressed in a4d1d259d28a6f0355a913472059970101cfc9df

PR #1305, opened 2019-05-08
Extra improvements in processing

ffont:
This should be the last PR with processing-related improvements. In this PR we optimize the logic for checking error messages in `convert_to_pcm` and re-introduce a processing step that solves `stereofy` crashes if the PCM file has no correct format. The step consists in trying conversion to PCM suing `ffmpeg` in case `stereofy` fails because existing PCM file has bad format. This was removed in a recent refactoring because it was not clear why it was needed. Now we've seen this is useful for a (very low**) fraction of sounds and we re introduce it. Typical case is when a sound is already PCM format so `convert_to_pcm` is not run even though the sound might be corrupted.

** This only affects 19 sounds out of 410000+ so the impact is rather little.

### DIFF
--- a/_sandbox/stereofy/main.cpp
+++ b/_sandbox/stereofy/main.cpp
@@ -81,7 +81,7 @@ int main(int argc, char* argv[])
 	// open the soundfile
 	if(!(fileIn = sf_open (inputFilename.c_str(), SFM_READ, &sfinfoIn)))
 	{
-		std::cout << "stereofy: failed to open file" << sf_strerror(fileIn) << std::endl;
+		std::cout << "stereofy: failed to open file: " << sf_strerror(fileIn) << std::endl;
 		return  1;
 	}
 	else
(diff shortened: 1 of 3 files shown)


_sandbox/stereofy/main.cpp
```diff
@@ -81,7 +81,7 @@ int main(int argc, char* argv[])
 	// open the soundfile
 	if(!(fileIn = sf_open (inputFilename.c_str(), SFM_READ, &sfinfoIn)))
 	{
-		std::cout << "stereofy: failed to open file" << sf_strerror(fileIn) << std::endl;
+		std::cout << "stereofy: failed to open file: " << sf_strerror(fileIn) << std::endl;
```
alastair: will we recompile?

ffont: yes we will. better to do it manually in the processing machines because the deploy scripts skips recompilation when binary existing

PR #1309, opened 2019-05-09
Forum improvements

### LINKED ISSUE #1287: Admins can't post on forum more than twice in 5 minutes
We limit the number of times that a user can post on the forum to once every 5 minutes, but it doesn't make sense to have this limit for admins (or perhaps moderators too). We could check the user permissions to see if we enforce this wait or not.

### LINKED ISSUE #1302: Incorrect values stored in solr for number of posts per thread
Some documents in the forum solr index have the following values:
`"num_posts": "F(num_posts) + Value(1)",`
This looks like an error where we save a post/thread, update the counts of the thread and forum in a post-save hook, and then index the item in solr. e.g. 
https://github.com/MTG/freesound/blob/a888d08fe9e28c32f12465358b3c4ed2b002fadc/forum/views.py#L266-L267
https://github.com/MTG/freesound/blob/a888d08fe9e28c32f12465358b3c4ed2b002fadc/forum/views.py#L383-L384
https://github.com/MTG/freesound/blob/a888d08fe9e28c32f12465358b3c4ed2b002fadc/forum/views.py#L193-L194

We should use `post.reload()` and `thread.reload()`, or pass a post id to the add methods, and re-get the object from the database before using it.

alastair:
**Issue(s)**
Fixes #1302
Fixes #1287

**Description**
This started as a fix of #1302 but got a bit bigger once we found some further issues in our forum hooks.
* Doesn't send F-values to solr
* Updates posts when a thread is updated
* Fix some bugs in setting first_post and last_post in a thread
* Clean up some imports, documentation, and code layout
* Don't send control characters to solr, and add tests for remove_control_chars
* Allow admins to make as many forum posts as they want (perhaps this only affected me, because I have no uploaded sounds, which also removes the limit)
* Change the number posts per day check to be able to reason about it better in tests

**Deployment steps**:

There are some maintenance tasks that we should also run:

**Update posts that have control characters**
There should be 2

```py
from utils import text
posts = Post.objects.all()
for p in posts:
     newd = text.remove_control_chars(p.body)
     if newd != p.body:
         print("%d - not the same" % p.id)
         # p.body = newd
         # p.save()
```

**Remove Threads that have no Posts**

    Thread.objects.annotate(post_count=Count('post')).filter(post_count=0).count()
    Thread.objects.annotate(post_count=Count('post')).filter(post_count=0).delete()

**Update Threads who have no last_post set**

```py
for thread in thread_qs:
    try:
        last_post = thread.post_set.filter(moderation_state="OK").reverse().all()[0]
    except IndexError:
        last_post = None
    if thread.last_post != last_post:
        print("bad thread %d" % thread.id)
        thread.last_post = last_post
        thread.save()
```

**Update Threads whose first_post isn't the earliest Post in the thread**

```py
for thread in thread_qs:
    try:
        first_post = thread.post_set.all()[0]
    except IndexError:
        first_post = None
    if thread.first_post != first_post:
        print("bad thread %d" % thread.id)
        # thread.first_post = first_post
        # thread.save()
```

**Re-index all forum posts**

    python manage.py reindex_solr_forum

### DIFF
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -278,7 +278,7 @@ class Profile(SocialModel):
 
             # Do not allow posts if last post is not older than 5 minutes
             seconds_per_post = settings.LAST_FORUM_POST_MINIMUM_TIME
-            if (today - self.user.posts.all().reverse()[0].created).seconds < seconds_per_post:
+            if (today - self.user.posts.all().reverse()[0].created).total_seconds() < seconds_per_post:
                 return False, "We're sorry but you can't post to the forum because your last post was less than 5 " \
                               "minutes ago"
 
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -267,6 +267,11 @@ class Profile(SocialModel):
                  ORDER BY tags_tag.name;""" % self.user_id)
 
     def can_post_in_forum(self):
+
+        # A forum moderator can always post
+        if self.user.has_perm('forum.can_moderate_forum'):
+            return True, ""
+
         user_has_posts_pending_to_moderate = self.user.posts.filter(moderation_state="NM").count() > 0
         if user_has_posts_pending_to_moderate:
             return False, "We're sorry but you can't post to the forum because you have previous posts still " \
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -272,7 +272,7 @@ class Profile(SocialModel):
             return False, "We're sorry but you can't post to the forum because you have previous posts still " \
                           "pending to moderate"
 
-        if self.user.posts.all().count() >= 1 and self.user.sounds.all().count() == 0:
+        if self.num_posts >= 1 and self.num_sounds == 0:
             today = datetime.datetime.today()
             reference_date = self.user.posts.all()[0].created
 
@@ -282,9 +282,9 @@ class Profile(SocialModel):
                 return False, "We're sorry but you can't post to the forum because your last post was less than 5 " \
                               "minutes ago"
 
-            # Do not allow posts if user has already posyted N posts that day
+            # Do not allow posts if user has already posted N posts that day
             max_posts_per_day = settings.BASE_MAX_POSTS_PER_DAY + pow((today - reference_date).days, 2)
-            if self.user.posts.filter(created__range=(today-datetime.timedelta(days=1), today)).count() > \
+            if self.user.posts.filter(created__range=(today-datetime.timedelta(days=1), today)).count() >= \
                     max_posts_per_day:
                 return False, "We're sorry but you can't post to the forum because you exceeded your maximum number " \
                               "of posts per day"
--- a/accounts/tests/test_profile.py
+++ b/accounts/tests/test_profile.py
@@ -96,7 +96,7 @@ class UserEditProfile(TestCase):
 
     @override_avatars_path_with_temp_directory
     def test_handle_uploaded_image(self):
-        user = User.objects.create_user("testuser", password="testpass")
+        user = User.objects.create_user("testuser")
         f = InMemoryUploadedFile(open(settings.MEDIA_ROOT + '/images/70x70_avatar.png'), None, None, None, None, None)
         handle_uploaded_image(user.profile, f)
 
@@ -106,8 +106,8 @@ class UserEditProfile(TestCase):
         self.assertEqual(os.path.exists(user.profile.locations("avatar.L.path")), True)
 
     def test_edit_user_profile(self):
-        User.objects.create_user("testuser", password="testpass")
-        self.client.login(username='testuser', password='testpass')
+        user = User.objects.create_user("testuser")
+        self.client.force_login(user)
         self.client.post("/home/edit/", {
             'profile-home_page': 'http://www.example.com/',
             'profile-username': 'testuser',
@@ -148,8 +148,8 @@ class UserEditProfile(TestCase):
 
     def test_edit_user_email_settings(self):
         EmailPreferenceType.objects.create(name="email", display_name="email")
-        User.objects.create_user("testuser", password="testpass")
-        self.client.login(username='testuser', password='testpass')
+        user = User.objects.create_user("testuser")
+        self.client.force_login(user)
         response = self.client.post("/home/email-settings/", {
             'email_types': 1,
         })
@@ -160,8 +160,8 @@ class UserEditProfile(TestCase):
 
     @override_avatars_path_with_temp_directory
     def test_edit_user_avatar(self):
-        User.objects.create_user("testuser", password="testpass")
-        self.client.login(username='testuser', password='testpass')
+        user = User.objects.create_user("testuser")
+        self.client.force_login(user)
         self.client.post("/home/edit/", {
             'image-file': open(settings.MEDIA_ROOT + '/images/70x70_avatar.png'),
             'image-remove': False,
@@ -185,12 +185,10 @@ class AboutFieldVisibilityTest(TestCase):
     """Verifies visibility of about field"""
 
     def setUp(self):
-        self.spammer = User.objects.create_user(username='spammer', email='spammer@example.com', password='testpass')
-        self.downloader = User.objects.create_user(username='downloader', email='downloader@example.com',
-                                                   password='testpass')
-        self.uploader = User.objects.create_user(username='uploader', email='uploader@example.com', password='testpass')
-        self.admin = User.objects.create_user(username='admin', email='admin@example.com', password='testpass',
-                                              is_superuser=True)
+        self.spammer = User.objects.create_user(username='spammer', email='spammer@example.com')
+        self.downloader = User.objects.create_user(username='downloader', email='downloader@example.com')
+        self.uploader = User.objects.create_user(username='uploader', email='uploader@example.com')
+        self.admin = User.objects.create_user(username='admin', email='admin@example.com', is_superuser=True)
 
         self.downloader.profile.num_sound_downloads = 1
         self.uploader.profile.num_sounds = 1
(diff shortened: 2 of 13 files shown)


alastair:
@ffont this is finished and ready for review now

accounts/models.py
```diff
@@ -267,24 +267,29 @@ def get_user_tags(self, use_solr=True):
...
-            if (today - self.user.posts.all().reverse()[0].created).seconds < seconds_per_post:
+            if (today - self.user.posts.all().reverse()[0].created).total_seconds() < seconds_per_post:
                 return False, "We're sorry but you can't post to the forum because your last post was less than 5 " \
                               "minutes ago"
 
-            # Do not allow posts if user has already posyted N posts that day
+            # Do not allow posts if user has already posted N posts that day
             max_posts_per_day = settings.BASE_MAX_POSTS_PER_DAY + pow((today - reference_date).days, 2)
-            if self.user.posts.filter(created__range=(today-datetime.timedelta(days=1), today)).count() > \
+            if self.user.posts.filter(created__range=(today-datetime.timedelta(days=1), today)).count() >= \
```
alastair: I made this change because it makes more sense in tests - if the limit is 9 posts, then I want the user to be able to make 9 and then fail on the 10th. With only `>` it would allow 10 and then fail on 11

accounts/models.py
```diff
@@ -267,24 +267,29 @@ def get_user_tags(self, use_solr=True):
...
+        if self.user.has_perm('forum.can_moderate_forum'):
+            return True, ""
+
         user_has_posts_pending_to_moderate = self.user.posts.filter(moderation_state="NM").count() > 0
         if user_has_posts_pending_to_moderate:
             return False, "We're sorry but you can't post to the forum because you have previous posts still " \
                           "pending to moderate"
 
-        if self.user.posts.all().count() >= 1 and self.user.sounds.all().count() == 0:
+        if self.num_posts >= 1 and self.num_sounds == 0:
```
alastair: saves some queries!

forum/tests.py
```diff
@@ -276,7 +315,7 @@ class ForumPageResponses(TestCase):
     def setUp(self):
         self.N_FORUMS = 1
         self.N_THREADS = 1
-        self.N_POSTS = 5
+        self.N_POSTS = 4
```
alastair: changed because of the > to >= change in can_post_in_forum

forum/views.py
```diff
@@ -378,10 +383,9 @@ def post_edit(request, post_id):
         if request.method == 'POST':
             form = PostReplyForm(request, '', request.POST)
             if form.is_valid():
-                delete_post_from_solr(post)
```
alastair: don't need to delete something to update it

utils/sound_upload.py
```diff
@@ -173,7 +174,7 @@ def create_sound(user, sound_fields, apiv2_client=None, process=True, remove_exi
             sound.geotag = geotag
 
     # 6 set description, tags
-    sound.description = sound_fields['description']
+    sound.description = remove_control_chars(sound_fields['description'])
```
alastair: this was removed when we merged these 2 methods, so I'm restoring it (we discussed that perhaps it wasn't needed because we also call it when adding sound to solr, but I think that it's also good to have this fixed in the database)

forum/models.py
```diff
@@ -214,27 +229,38 @@ def update_num_posts_on_post_insert(sender, instance, created, **kwargs):
...
 
 @receiver(post_delete, sender=Post)
-def update_last_post_on_post_delete(sender, instance, **kwargs):
+def update_thread_on_post_delete(sender, instance, **kwargs):
     """Update num_posts counts and last_post pointers when a Post is deleted.
 
     Reduce num_posts by 1 for the author, forum, and thread
     Set last_post for the forum
     Set last_post for the thread. If this was the only remaining post
     in the thread and it was deleted, also delete the thread.
```
ffont: What's the policy for ending sentences with/without a dot? I always struggle with this in code comments. Looks like if there's a single sentence in the line, then no need for dot. If multiple sentence in same line put dots at the end. But title is single sentence and has dot... Is this your recommendation?

forum/views.py
```diff
@@ -174,13 +174,13 @@ def reply(request, forum_name_slug, thread_id, post_id=None):
...
     latest_posts = Post.objects.select_related('author', 'author__profile', 'thread', 'thread__forum')\
                        .order_by('-created').filter(thread=thread, moderation_state="OK")[0:15]
-    user_can_post_in_forum = request.user.profile.can_post_in_forum()
+    user_can_post_in_forum, user_can_post_message = request.user.profile.can_post_in_forum()
     user_is_blocked_for_spam_reports = request.user.profile.is_blocked_for_spam_reports()
 
     if request.method == 'POST':
         form = PostReplyForm(request, quote, request.POST)
 
-        if user_can_post_in_forum[0] and not user_is_blocked_for_spam_reports:
```
ffont: hehe, much better with the unpacked variables.

forum/views.py
```diff
@@ -246,29 +246,34 @@ def reply(request, forum_name_slug, thread_id, post_id=None):
...
                     set_to_moderation = True
                 else:
-                    post = Post.objects.create(author=request.user, body=form.cleaned_data['body'], thread=thread)
-                    add_post_to_solr(post)
+                    post = Post.objects.create(author=request.user, body=post_body, thread=thread)
+                    add_post_to_solr(post.id)
                     set_to_moderation = False
 
                 # Add first post to thread (first post will always be the same)
                 # We need to reload thread object from DB, not so overwrite the object we created before when saving
```
ffont: typo: `not to overwrite`

utils/search/search_forum.py
```diff
@@ -17,73 +17,84 @@
...
+    logger.info("done")
 
 
-def add_posts_to_solr(posts):
-    logger.info("adding multiple forum posts to solr index")
-    solr = Solr(settings.SOLR_FORUM_URL, auto_commit=False)
+def add_post_to_solr(post_id):
+    """Add a forum post to solr
+    Arguments:
+        post_id (int): ID of a post object"""
```
ffont: wrong docstring format, although I'm starting to like this one better because the types are more readable... Maybe we should switch.

utils/search/search_forum.py
```diff
@@ -17,73 +17,84 @@
...
-    solr.add(documents)
 
-    solr.commit()
-    logger.info("optimizing solr index")
-    #solr.optimize()
-    logger.info("done")
+def add_all_posts_to_solr(slice_size=4000):
+    """Add all forum posts to solr
+    Arguments:
+        slice_size (int): The number of posts to send to solr at a time"""
```
ffont: Same docstring comment as above

utils/sound_upload.py
ffont: I'm still not sure about this one. Why would be want to have this fixed in DB for sound descriptions and not for, e.g., comments or any other text field? If we only use `remove_control_chars` for this field, I think we should add a comment about why here and not in other places.

accounts/tests/test_profile.py
```diff
@@ -345,3 +347,83 @@ def test_email_is_valid(self):
         # Test email is still invalid after user is deleted and bounces happened
         EmailBounce.objects.create(user=user, type=EmailBounce.PERMANENT)
         self.assertEqual(user.profile.email_is_valid(), False)
+
+
+class ProfilePostInForumTest(TestCase):
+    def setUp(self):
+        self.user = User.objects.create_user("testuser", password="testpass", email='email@freesound.org')
```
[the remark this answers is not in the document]
alastair: `create_user` requires a password, but by using force_login, now it doesn't matter what it is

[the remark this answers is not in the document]
alastair: ah, I didn't realise that there was a default of None. You're right, this could probably be removed as long as force_login still works

accounts/tests/test_profile.py
```diff
@@ -345,3 +347,83 @@ def test_email_is_valid(self):
...
+        self.assertFalse(can_post)
+        self.assertIn("you have previous posts", reason)
+
+    def test_can_post_in_forum_time(self):
+        """If you have no sounds, you can't post within 5 minutes of the last one"""
+        created = datetime.datetime(2019, 2, 3, 10, 50, 00)
+        post = Post.objects.create(thread=self.thread, body="", author=self.user, moderation_state="OK")
+        post.created = created
+        post.save()
+        with freezegun.freeze_time("2019-02-03 10:52:30"):
```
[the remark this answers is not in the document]
alastair: good question. The examples don't show this, but the documentation seems to show that it's possible. What would you prefer, that we use the datetime.datetime constructor, or actually use this iso format? In my opinion it's actually easier to understand as a string. We could use dateutil to turn them into date objects. It's installed as a dependency of freezegun anyway.

forum/models.py
alastair: the policy is unclear. I like using . after the first sentence because it's normally used as a short description in html docs, and so is a good policy to get into. This specific example is a bit confusing because these appear to be list items (which I wouldn't normally put a dot after). I notice a few lines down I skip the . in the last sentence of a paragraph. It's all a mess, let's make some guidelines

utils/search/search_forum.py
alastair: yeah, we had this conversation. Tell me which one you prefer and I'll add it to the programming guidelines.

accounts/tests/test_profile.py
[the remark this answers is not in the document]
ffont: I like strings more as well, so if possible I'd do it that way

utils/search/search_forum.py
ffont: we decided to use the reStructuredText/Sphinx, but I like how parameter types are shown here, it is more readable. maybe we should use this (google) one.

utils/sound_upload.py
alastair: I raised this as #1322

alastair:
I addressed some comments, opened some tickets for some future work, and also opened #1320 for guidelines for documentation

ffont:
In the instructions, should `thread_qs`be `Thread.objects.all()`?

PR #1317, opened 2019-05-15
Warn if tags include characters outside of a-zA-Z0-9

### LINKED ISSUE #1210: UnicodeEncodeError: 'ascii' codec can't encode character u'\xe1' in position 1: ordinal not in range(128)
https://logserver.mtg.upf.edu/sentry/freesound-web/issues/1133/

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe1' in position 1: ordinal not in range(128)
(5 additional frame(s) were not displayed)
...
  File "accounts/views.py", line 600, in describe_sounds
    if not forms[i][f].is_valid():
  File "utils/forms.py", line 51, in clean
    tags = clean_and_split_tags(str(value))

UnicodeEncodeError: 'ascii' codec can't encode character u'\xe1' in position 1: ordinal not in range(128)
```

### LINKED ISSUE #1254: BUG - Server error when describing sounds with particular unicode text characters.
Whilst describing sounds in the upload portal, a user will be returned with a server error when trying to input unicode U+00E1 [á] in the tags field

The bug is reproducible, tested also with unicode U+00E2; same result, so the affected characters may be extensive. 

Please look into this. 
Cheers
Sam

alastair:
**Issue(s)**
Fixes #1254
Fixes #1210

**Description**
We don't support any other characters, but we shouldn't raise an
uncaught exception either.

**Deployment steps**:
Resolve issue in sentry: https://logserver.mtg.upf.edu/sentry/freesound-web/issues/1133/
https://logserver.mtg.upf.edu/sentry/freesound-web/issues/1216/

### DIFF
--- a/utils/forms.py
+++ b/utils/forms.py
@@ -17,8 +17,12 @@
 # Authors:
 #     See AUTHORS file.
 #
+import re
 
 from django import forms
+from django.core import validators
+from django.core.exceptions import ValidationError
+
 from utils.text import clean_html, is_shouting
 from django.conf import settings
 from recaptcha.client import captcha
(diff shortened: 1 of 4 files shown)


utils/forms.py
```diff
@@ -38,23 +42,27 @@ def clean(self, value):
         value = super(HtmlCleaningCharField, self).clean(value)
         if is_shouting(value):
             raise forms.ValidationError('Please moderate the amount of upper case characters in your post...')
-        try:
-            return clean_html(value)
-        except (HTMLParseError, UnicodeEncodeError):
```
alastair: after discussion with @andrebola we realised that the switch to bleach no longer raises these exceptions

utils/forms.py
```diff
@@ -38,23 +42,27 @@ def clean(self, value):
...
+
+    def __init__(self, **kwargs):
+        super(TagField, self).__init__(**kwargs)
+        self.validators.append(
+            validators.MinLengthValidator(3, 'You should add at least 3 different tags. Tags must be separated by '
+                                             'spaces'))
+        self.validators.append(
+            validators.MaxLengthValidator(30, 'There can be maximum 30 tags, please select the most relevant ones!'))
+
+    def to_python(self, value):
```
alastair: the clean method in django Fields call to_python to convert the input into the correct data type, and then each of the validators in self.validators. This change also allows us to use CharField's to_python to check for empty values

utils/forms.py
```diff
@@ -38,23 +42,27 @@ def clean(self, value):
...
-                                        '(and/or commas).')
-        elif len(tags) > 30:
-            raise forms.ValidationError('There can be maximum 30 tags, please select the most relevant ones!')
-        return tags
+
+    def __init__(self, **kwargs):
+        super(TagField, self).__init__(**kwargs)
+        self.validators.append(
+            validators.MinLengthValidator(3, 'You should add at least 3 different tags. Tags must be separated by '
+                                             'spaces'))
```
alastair: after discussion with @ffont we removed commas from this error message, now we just insist on using spaces

Issue #1307, opened 2019-05-09
Update stripe checkout API


ffont:
We need to upgrade the Stripe checkout API we use for donations to the new version which supports Strong Customer Authentication (and which will be a requirement for some transactions after new EU directives come into play in September). Changes should be fairly simple.

Here is a migration guide form Stripe docs: https://stripe.com/docs/payments/checkout/migration

Here is an example (legacy) code using JS like we're currently doing: https://stripe.com/docs/checkout#integration-custom
Our code will most likely be very similar to that.

andrebola:
I contacted stripe and this was the reply:

`[...] So I have some documentation here explaining new Checkout, along with the quickstarts for the server-side and the client version: https://stripe.com/docs/payments/checkout I don't think the variable amounts option is available for new Checkout, but it may be something that could be added in the future`

I guess we will have to wait until they add this in the future.

andrebola:
After more discussion, we saw a way of using stripe's python library to have variable amount of donations.

Is important to create the webhook from stripe dashboard, with the following parameters:
URL: https://freesound.org/donations/donation-complete-stripe/
Event types: checkout.session.completed

PR #1324, opened 2019-05-21
New stripe api

andrebola:
**Issue(s)**
#1307 

**Description**
This PR will update the code to the last version of stripe API.
Now we create a session with the donation info from the view, then we pass the session_id to the frontend which makes a redirect to stripe with the given session.

After the user completes the transaction it's redirected to the 'success' page which displays a notification.

Now we create the Donation object only when stripe sends us a notification using webhooks.

**Deployment steps**:
Is important to create the webhook from stripe dashboard, with the following parameters:
URL: https://test.freesound.org/donations/donation-complete-stripe/
Event types: checkout.session.completed

Also, from the dashboard you need to copy the Signing secret for the created Webhook and then add it to the settings with the key "STRIPE_WEBHOOK_SECRET"


donations/views.py
```diff
@@ -77,43 +77,51 @@ def _save_donation(encoded_data, email, amount, currency, transaction_id, source
...
 
-        token = request.POST.get('stripeToken', False)
-        email = request.POST.get('stripeEmail', False)
+        try:
+            event = stripe.Webhook.construct_event(
+              payload, sig_header, endpoint_secret
+            )
+        except ValueError as e:
+            # Invalid payload
+            return HttpResponse(status=400)
```
ffont: This is stripe making a request to the URL we provide right? So if this error happens this will be the response to stripe's request?

donations/views.py
```diff
@@ -77,43 +77,51 @@ def _save_donation(encoded_data, email, amount, currency, transaction_id, source
...
+        except stripe.error.SignatureVerificationError as e:
+            # Invalid signature
+            return HttpResponse(status=400)
+
+        # Handle the checkout.session.completed event
+        if event['type'] == 'checkout.session.completed':
+            session = event['data']['object']
+
+            # Fulfill the purchase...
+            amount = int(session['display_items'][0]['amount'])/100
```
ffont: if I understand correctly this division will truncate in python2. should use `100.0`  for example

donations/urls.py
```diff
@@ -26,6 +26,8 @@
 urlpatterns = [
     url(r'^donate/$', views.donate, name="donate"),
     url(r'^donors/$', views.DonationsList.as_view(), name="donors"),
+    url(r'^donation-session-stripe/$', views.donation_session_stripe, name="donation-session-stripe"),
+    url(r'^donation-success-stripe/$', views.donation_success_stripe, name="donation-success-stripe"),
```
ffont: donation success is for when user comes back from stripe page (after making the donation) but still we have not received the webhook callback from stripe right?

for donating with paypal we also go to the paypal website and then go back to Freesound right? or how does it work? If the flows are similar it would be great to make sure all views are named following the same structure.

donations/views.py
andrebola: If for some reason we return an error Stripe will retry later, but also someone could call this url, in this case, we also want to return an error.

donations/views.py
andrebola: Stripe handles the amount as integer (multiplied by 100), that is why we divide it for storing it with decimals. 

You're right, this should 100.0 for python 2. Unfortunately, this was not captured by the test cases because of the amounts used.

donations/urls.py
andrebola: Yes, you're right. Donation success is where the user is redirected, it only shows a message saying that the donation will be processed and redirects to the donations page.

For Paypal the url 'donation-complete-paypal' also listens to paypal requests and creates the Donation object. Therefore, now the names are more consistent than before.

andrebola:
Ok, I will put this info in a wiki.

donations/views.py
ffont: this has not been fixed rigth? will you update the tests as well?

ffont:
Hi @andrebola, thanks for adding the wiki page, it looks great as is indeed very useful!
There are still two things that I need to clarify before merging this:

- The division issue has not been fixed yet? or you might have forgotten to push some changes...
- According to the wiki page, both methods go to 3rd party website (paypal or stripe) and then get redirected to freesound once process is done. For the stripe case it says users is redirected to `donation-success-stripe` URL name. How does the Paypal redirect work? Could we also use a `donation-success-paypal` URL so the structure is exactly the same for both methods? Also if there is a `donate` URL name that only refers to Paypal (as the equivalent for Stripe seems to be `donation-session-stripe`), could we rename this to something like `donate-paypal`?

donations/views.py
andrebola: Yes, I just submitted the changes.

andrebola:
Ok, I just added 'donation-session-paypal' and so it's the same process for stripe and paypal. I also changed 'donation-success-stripe' to 'donation-success', so now it's not only for stripe.

Issue #1248, opened 2018-10-13
Reduce Recaptcha use


qubodup:
The use of recaptcha to lower the impact of spam is smart. It is lately in my experience prone to being extremely annoying and broken/slow.

Its use punishes:
spammers (good)
moderators (at least we know what we force users to do)
people trying to get legal clearance for projects they are working on and using sounds for.
People who protect privacy by not having a google account (if you don't have one, you are forced to solve AI teaching image identification tasks).

Solution suggesiton:
- keep track of when the last message was sent. Don't require captcha if the last message was sent more than 10 minutes ago (this also means the very first message you ever send doesn't require it).

qubodup:
PS: I would also not require captcha if user is responding to a message received from the user they are writing to. Probably hard to track

qubodup:
PPS: other possible reasons to trust:
- being whitelisted
- having at least one approved sound upload

alastair:
Can you give an example of the pages where the captcha is causing you (or others?) problems? It sounds like the main page causing problems is message sending.
I agree that we could skip the captcha for trusted users - we've already started using this concept in other places so it shouldn't be much of an issue to extend. We already monitor the number of spam message reports from users, so we would need to extend this to see if messages sent without captcha tend to be reported as more or less spammy.

qubodup:
message sending is the main one. I'm not aware of other places where it disturbs. I haven't used the forum in a while.

Oh I just though of another thing: if the user downloaded a sound by the user they are writing to, we could try to assume that they are writing to ask for attribution terms or something like that and not require captcha. Since Freesound tracks downloads per-user this might be realistic to do without a lot of additional work.

ffont:
Hi, IIRC we don't ask for recaptcha in some of the situations you mention. However we could include more "whitelisted" cases as I also found this problem having to convince Freesound I'm not a robot. I'd say users that uploaded one sound never need recaptcha. Also users that posted more than X.

qubodup:
Another case: in registration, the form misinforms the user of acceptable usernames. Ideally (user forgets to click terms agreement) this form would only ask for captcha once to be successful:
https://github.com/MTG/freesound/issues/1258

PR #1327, opened 2019-05-28
 Don't show re-captcha form when trusted users write/reply messages

ffont:
**Issue(s)**
https://github.com/MTG/freesound/issues/1248

### DIFF
--- a/messages/tests/test_recaptcha_presence.py
+++ b/messages/tests/test_recaptcha_presence.py
@@ -1,0 +1,74 @@
+#
+# Freesound is (c) MUSIC TECHNOLOGY GROUP, UNIVERSITAT POMPEU FABRA
+#
+# Freesound is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Freesound is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#     See AUTHORS file.
+#
+
+from django.contrib.auth.models import User
+from django.test import TestCase
+from django.urls import reverse
+
+from messages.models import Message, MessageBody
+
+
+class RecaptchaPresenceInMessageForms(TestCase):
+    """
+    Test that whether the recapctha field should or should not be present in new message/reply message forms.
+    """
+
+    fixtures = ['initial_data']
+
+    def setUp(self):
+        # Create one user which is a potential spammer and one which is not
+        self.no_spammer = User.objects.create_user(username='noSpammer', email='noSpammer@example.com')
+        self.no_spammer.profile.num_sounds = 4  # Having sounds will make user "trustable"
+        self.no_spammer.profile.save()
+
+        self.potential_spammer = User.objects.create_user(
+            username='potentialSpammer', email='potentialSpammer@example.com')
+
+        self.message_sender = User.objects.create_user(username='sender', email='sender@example.com')
+
+    def test_captcha_presence_in_new_message_form(self):
+
+        # No spammer case, recaptcha field should NOT be shown
+        self.client.force_login(user=self.no_spammer)
+        resp = self.client.get(reverse('messages-new'))
+        self.assertNotIn('recaptcha', resp.content)
+
+        # Potential spammer (has no uploaded sounds), recaptcha field should be shown
+        self.client.force_login(user=self.potential_spammer)
+        resp = self.client.get(reverse('messages-new'))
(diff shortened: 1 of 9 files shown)


messages/views.py
```diff
@@ -184,13 +183,12 @@ def new_message(request, username=None, message_id=None):
...
-                previously_contacted_user2 = list(Message.objects.filter(user_to = request.user.id)\
-                        .values_list('user_from__username', flat='True').distinct())
+        # Only autocompleting for previously contacted users
+        previously_contacted_user = list(
+            Message.objects.filter(user_from=request.user.id).values_list('user_to__username', flat='True').distinct())
+        previously_contacted_user2 = list(
+            Message.objects.filter(user_to=request.user.id).values_list('user_from__username', flat='True').distinct())
 
-                results = list(set(previously_contacted_user + previously_contacted_user2))
+        results = list(set(previously_contacted_user + previously_contacted_user2))
```
alastair: I know you're only reformatting this code, but does it make sense for this to return the value of a set? This means that the order of these items could change. It'd be nice to see the list ordered by the frequency of contact.
It would also be nice to make only 1 query (to minimise load on the database)

messages/views.py
```diff
@@ -159,23 +163,18 @@ def new_message(request, username=None, message_id=None):
...
                 body = "> --- " + message.user_from.username + " wrote:\n>\n" + body
 
                 subject = "re: " + message.subject
                 to = message.user_from.username
 
-                if request.user.profile.num_sounds:
-                    form = MessageReplyFormNoCaptcha(initial=dict(to=to, subject=subject, body=body))
-                else:
-                    form = MessageReplyForm(initial=dict(to=to, subject=subject, body=body))
+                form = form_class(initial=dict(to=to, subject=subject, body=body))
```
alastair: I typically use dict literals here instead of the class syntax:
`{"to": to, "subject": subject}`

messages/views.py
```diff
@@ -107,28 +110,33 @@ def message(request, message_id):
...
         else:
             if form.is_valid():
                 user_from = request.user
                 user_to = form.cleaned_data["to"]
                 subject = form.cleaned_data["subject"]
                 body = MessageBody.objects.create(body=form.cleaned_data["body"])
 
-                Message.objects.create(user_from=user_from, user_to=user_to, subject=subject, body=body, is_sent=True, is_archived=False, is_read=False)
-                Message.objects.create(user_from=user_from, user_to=user_to, subject=subject, body=body, is_sent=False, is_archived=False, is_read=False)
+                Message.objects.create(user_from=user_from, user_to=user_to, subject=subject, body=body, is_sent=True,
```
alastair: I wonder if we should do something about the `is_sent` field. It always confuses me when I see it. I expect it to mean that the message has been successfully sent to someone, rather than indicating that this is the message in the output of the sender. Any suggestions on what else it could be called?

messages/tests/test_recaptcha_presence.py
```diff
@@ -0,0 +1,74 @@
...
+        self.no_spammer.profile.save()
+
+        self.potential_spammer = User.objects.create_user(
+            username='potentialSpammer', email='potentialSpammer@example.com')
+
+        self.message_sender = User.objects.create_user(username='sender', email='sender@example.com')
+
+    def test_captcha_presence_in_new_message_form(self):
+
+        # No spammer case, recaptcha field should NOT be shown
```
alastair: "Not a spammer case"

messages/views.py
```diff
@@ -107,28 +110,33 @@ def message(request, message_id):
     tvars = {'message': message}
     return render(request, 'messages/message.html', tvars)
 
+
 @login_required
 @transaction.atomic()
 def new_message(request, username=None, message_id=None):
+
+    if request.user.profile.num_sounds > 0 or request.user.profile.num_posts > 5:
```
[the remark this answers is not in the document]
ffont: I thought about it but then I saw it would not be re-used (by now). The thing is that different situations would require different levels of trustworthiness. For example, to decide whether or not to display the user's about text in profile we check for number of downloads. This is a less strict trustworthiness measure. I would be happy to unify different places where similar things are done and maybe have a `is_trustworthy` method in `Profile` with some sort of `level` parameter to decide how strict we are. This requires some investigation to see in which places this method could be used. Maybe I could implement the method only for this case and open an issue for the further investigations...

messages/views.py
ffont: that's a tricky one. maybe `is_sender_copy`?

messages/views.py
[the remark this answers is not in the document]
ffont: Added in 83a8a84ae96bee7da050771ca70019ba20d74a9c

messages/views.py
ffont: The order is not relevant because this list is used for the JS autocomplete feature, which will further filter the list and show results according to the characters that user is typing in the username box. I nevertheless updated the query so now we only make one. Addressed in bec9e953de2d51b12553146736fcf5a36be14874

messages/views.py
ffont: Addressed in 629ff4174be55a9d770a0dadc3ad36be502233fc

messages/views.py
alastair: let's leave it like this for now, but consider changing it if we make any further changes to messages

PR #1328, opened 2019-05-29
Allow bulk describe using custom permission

ffont:
This PR brings a new way to enable bulk describe for users who are granted a specific permission. The specific permission can be given to users by adding users to the `Group` with name `bulk_uploaders` via the Django admin (or the shell). 

The `Group` with the associated permission will be created automatically as part of a migration. However due to the nature in which migrations work this causes some problems and to make the migration work I had to follow a workaround as described in https://code.djangoproject.com/ticket/23422. It would be desirable to have an alternative for that.

**Deploy steps**

Load fixture which installs the `bulk_uploaders` `Group` with the appropriate permissions:
```
python manage.py loaddata bulk_uploaders_group
```


DEVELOPERS.md
```diff
@@ -29,4 +29,19 @@ that were made, and allow feedback if necessary
...
+
+If there is a need for defining custom permissions we should define them in the corresponding model's `Meta` class
+as [described in the Django docs](https://docs.djangoproject.com/en/dev/topics/auth/customizing/#custom-permissions). There
+is no generic model to **host** all custom permissions, but we'll add them to the most closely related model. These
+permissions should then be added to a `Group` so that we can manage them properly in the Django admin. The `Group` should
+be created in the corresponding database migration in which the permission is added to the model.
+
+Currently, we only use the following custom permissions:
+* `tickets.can_moderate` (in `Ticket` model, used to allow sound moderation)
+* `forum.can_moderate_forum` (in `Post` model, used to allowforum moderation)
```
alastair: allow forum

accounts/tests/test_upload.py
```diff
@@ -160,6 +160,32 @@ def test_describe_selected_files(self):
...
+        self.assertFalse(user.profile.can_do_bulk_upload())
+        user.profile.is_whitelisted = True
+        self.assertTrue(user.profile.can_do_bulk_upload())
+
+        # If user is assigned the bulk_uploaders group then bulk upload is allowed
+        user.profile.refresh_from_db()
+        self.assertFalse(user.profile.can_do_bulk_upload())
+        group = Group.objects.get(name="bulk_uploaders")
+        user.groups.add(group)
+        # Reload object from db to refresh permission's caches (Note that refresh_from_db() does clear permission cache)
```
alastair: does -> doesn't

tickets/fixtures/initial_data.json
```diff
@@ -105,5 +102,5 @@
...
+    }}, {"pk": 5, "model": "auth.group", "fields": {"name": "support", "permissions": [["can_moderate", "tickets", "ticket"], ["add_userannotation", "tickets", "userannotation"]]}}, {"pk": 3, "model": "auth.group", "fields": {"name": "moderators", "permissions": [["change_comment", "comments", "comment"], ["delete_comment", "comments", "comment"], ["change_pack", "sounds", "pack"], ["delete_pack", "sounds", "pack"], ["change_sound", "sounds", "sound"], ["delete_sound", "sounds", "sound"], ["can_moderate", "tickets", "ticket"], ["add_userannotation", "tickets", "userannotation"]]}},
```
alastair: What's this change? From what I understand when we were looking through some tickets related to the migration/permissions problem, you can use either migrations _or_ initial_data to put things into the database. An alternative to the migration to make the group would be to have a fixture that we can load in the test. This way we wouldn't need the hack, and we also wouldn't get the output of the hack running in our tests.
I'm not worried that fixtures are reported as being "deprecated"

ffont: Nice, I'll use a fixture for the group.
this change was to remove some unused permissions from the existing fixtures as well, although it does not really make any difference.

DEVELOPERS.md
```diff
@@ -29,4 +29,19 @@ that were made, and allow feedback if necessary
...
+
+## Specific notes
+
+### Custom Django permissions
+
+If there is a need for defining custom permissions we should define them in the corresponding model's `Meta` class
+as [described in the Django docs](https://docs.djangoproject.com/en/dev/topics/auth/customizing/#custom-permissions). There
+is no generic model to **host** all custom permissions, but we'll add them to the most closely related model. These
+permissions should then be added to a `Group` so that we can manage them properly in the Django admin. The `Group` should
+be created in the corresponding database migration in which the permission is added to the model.
```
alastair: now this should be that we create a fixture, right?

accounts/models.py
```diff
@@ -305,8 +305,20 @@ def can_post_in_forum(self):
...
-        return self.num_sounds >= settings.BULK_UPLOAD_MIN_SOUNDS or self.is_whitelisted
+        """
+        Returns True if the corresponding User is allowed to use the bulk upload feature. This will happen if one of
+        the following conditions is met:
+        * User has uploaded at least settings.BULK_UPLOAD_MIN_SOUNDS sounds
+        * User is whitelisted
+        * User has been granted special permissions for bulk upload
+
+        Returns:
+            bool: True if user ca do bulk upload, False otherwise.
```
alastair: can

general/tests.py
```diff
@@ -28,7 +28,7 @@
 
 class ReportCountStatusesManagementCommandTestCase(TestCase):
 
-    fixtures = ['initial_data']  # Needed for loading licenses, etc
+    fixtures = ['licenses']  # Needed for loading licenses, etc
```
alastair: comment doesn't make as much sense anymore, I would delete it

sounds/fixtures/bulk_uploaders_group.json
```diff
@@ -0,0 +1,10 @@
+[
```
alastair: any reason why you have a single fixture per group instead of putting them all in one file?

support/tests.py
```diff
@@ -25,7 +25,7 @@
 
 
 class SupportRequestsTest(TestCase):
-    fixtures = ['initial_data.json', 'moderation_test_users.json']
+    fixtures = ['moderation_test_users.json']
```
alastair: extensions for fixtures are optional. since no other test has them, I'd recommend that we remove it here

tickets/fixtures/moderation_groups.json
```diff
@@ -0,0 +1,51 @@
+[
+   {
+      "model":"auth.group",
+      "pk":3,
```
alastair: some of your fixtures have null pk, some are set. what's the difference?

tickets/fixtures/moderation_queues.json
```diff
@@ -0,0 +1,13 @@
+[
+  {
+    "pk": 1,
+    "model": "tickets.queue",
+    "fields": {
+      "notify_by_email": false,
+      "name": "sound moderation",
+      "groups": [
+        3
```
alastair: can this use the natural key to get the group instead of a pk?

utils/test_helpers.py
```diff
@@ -61,7 +61,7 @@ def create_user_and_sounds(num_sounds=1, num_packs=0, user=None, count_offset=0,
     Creates User, Sound and Pack objects useful for testing. A 'sound_counter' is used to make sound names unique as
     well as other fields like md5.
     NOTE: creating sounds requires License objects to exist in DB. Do that by making sure your test case loads
-    'initial_data' fixture, i.e. "fixtures = ['initial_data']".
+    'licenses' fixture, i.e. "fixtures = ['licenses']".
     :param num_sounds: N sounds to generate
```
alastair: not super important, but what I do when I make changes and see a related change is do that one too. You could have updated the docstring format here

alastair:
comments made

DEVELOPERS.md
ffont: 100%, addressed in c425e660b5edd489c74c9f7b8a6c41ee0056de67

sounds/fixtures/bulk_uploaders_group.json
ffont: basically for clarity and because when we load them manually to the database it is more explicit if groups are defined in separate files. however, given that loading a fixture will not duplicate existing objects we could also place them all in one file which is update when new groups are to be added. would you prefer that option @alastair?

tickets/fixtures/moderation_groups.json
ffont: only the new fixture I've made has `"pk":null,`, the "old" ones all have specific ID. when pk is set to null it assigns an ID at the time of creation. I think this works well because it does not assume a specific ID.

utils/test_helpers.py
ffont: I agree. I think I changed the format somewhere else but I must have forgotten this one.

sounds/fixtures/bulk_uploaders_group.json
alastair: I think that it makes more sense to include all groups in a single fixture file. So it seems that we'll be OK to add items to it in the future and continue to use loaddata to add a single new group? This sounds fine.

tickets/fixtures/moderation_groups.json
alastair: Yes, I think it's good that we can let django create the PK when needed. I'd prefer that we have all items in the fixture with null pk.

utils/test_helpers.py
ffont: Addressed inb1ae6e0e09564ede08626f4d0cff25e1610c10d0
It is interesting the case where we return tuples. I found an explanation somewhere about how to deal with that in the docstring, but I think there is no common way to do it. Anyway, if you like the way it has been approached here maybe we could copy this docstring as an example in the developers file.

tickets/fixtures/moderation_groups.json
ffont: I set a number of `pk` to `null` wherever possible (in an easy way) f5e643998d1350c2a37c8d9a7c4ab2fe91397051

I'd agree in further optimizing fixtures to completely remove all non-null `pk`s.

tickets/fixtures/moderation_queues.json
ffont: Done in f5e643998d1350c2a37c8d9a7c4ab2fe91397051

sounds/fixtures/bulk_uploaders_group.json
ffont: Moved all groups to a single fixture 7e920bc042706832a2f9b4df40dd4e53478b4e06

utils/test_helpers.py
ffont: Further formatted and added as example in `DEVELOPERS` 840e6700728e8e9116aea7fdc0fe1ba4e52fb26d

Issue #1282, opened 2019-02-20
Can't save when editing user profile 


ffont:
This is a bug reported through our support channels by a user who changed her username for an uppercase version of it. After that change, she can no longer edit profile options and save them because a validation error is thrown "This username is already taken or has been in used in the past.". 

This is the relevant form validation code: https://github.com/MTG/freesound/blob/15cd9ca68cb433c6e06f6047d995846f79060b8b/accounts/forms.py#L289-L311

We should make sure the logic works for the case when a user had a username with same characters but different case the past.

ffont:
usernames are case insensitive in Freesound, therefore the error here is that changing the username for a similar version with different capitalization should not count as username change and should not generate an `OldUsername`object.

PR #1333, opened 2019-06-11
Make username change case insensitive

ffont:
**Issue(s)**
https://github.com/MTG/freesound/issues/1282

**Description**
When users change their profiles we check if username has changed so
we can create `OldUsername` objects that we'll use to redirect URLs of
old username to new username. Because usernames are treated as case
insensitive, username changes should only create `OldUsername` objects
when the username update includes changes other than capitalization.
Otherwise this can create problems as described in the issue.

**Deployment steps**:
Run the following commands in console to remove all existing `OldUsername` objects that link versions of the same username with different cases (there should be ~85 cases in production)

```python
for ou in OldUsername.objects.select_related('user').all():
    if ou.username.lower() == ou.user.username.lower():
        print 'deleting:', ou
        ou.delete()
```

### DIFF
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -484,7 +484,7 @@ post_save.connect(create_user_profile, sender=User)
 def presave_user(sender, instance, **kwargs):
     try:
         old_username = User.objects.get(pk=instance.id).username
-        if old_username != instance.username:
+        if old_username.lower() != instance.username.lower():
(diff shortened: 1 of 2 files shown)


alastair:
There are some exceptions in sentry related to OldUsernames:
https://logserver.mtg.upf.edu/sentry/freesound-web/issues/1138/
We should double-check if it's related to this issue, or due to some other reason

Issue #1236, opened 2018-09-03
Improvements in geotags edit map


ffont:
A couple of simple improvements for the geotags edit map suggested by users in [this forum thread](https://freesound.org/forum/bug-reports-errors-and-feature-requests/41157/?page=1#post91908):

- Make map bigger
- When latitude, longitude or zoom are set manually, center the map at the corresponding position

PR #1334, opened 2019-06-11
Geotag edit map improvements

ffont:
**Issue(s)**
https://github.com/MTG/freesound/issues/1236

**Description**
Add a couple of improvements to the geotags edit map (i.e. the map used to add a geotag to a sound):

- Make map bigger.
- Make map update when changing `lat`, `lon` or `zoom` values from the input boxes next to the map.


templates/sounds/sound_edit.html
```diff
@@ -279,6 +279,14 @@ <h3>Delete</h3>
         $("#{{geotag_form.zoom.auto_id}}").val(Math.floor(zoom));
     }
 
+    function centerMapToLatLonZoomFields() {
+        var lat = $("#{{geotag_form.lon.auto_id}}").val();
+        var lon = $("#{{geotag_form.lat.auto_id}}").val();
+        var zoom = $("#{{geotag_form.zoom.auto_id}}").val();
+        window.map.setZoom(parseInt(zoom, 10) - 1);
```
alastair: What happens if these values aren't numbers?

templates/accounts/describe_sounds.html
```diff
@@ -297,6 +307,16 @@ <h3>Geotag</h3>
                                 $("#{{form.geotag.zoom.auto_id}}").attr('disabled', false);
                             }
                         });
+                        $("#{{form.geotag.lon.auto_id}}").change(function (event) {
+                            centerMapToLatLonZoomFields({{forloop.counter}});
```
alastair: Does this mean that the map is going to move every time that you press a key? This might be a bit annoying. I'm not sure what the better solution is. What's the result if it's deleted completely?
We could use blur(), when focus is removed from the element, but that's not great either because you have to click out for it to trigger. What about a timeout, to allow people to delete and write, but not have it update immediately? 
Do we expect people to paste values into here? A paste event might be useful too?

templates/accounts/describe_sounds.html
```diff
@@ -279,7 +289,7 @@ <h3>Geotag</h3>
             </p>
             <div class="gmap-wrapper" id="gmap-wrapper-{{forloop.counter}}">
                 <p>Drag this map to set a geotag:</p>
-                <div class="describe-gmap" id="gmap{{forloop.counter}}" style="width: 400px; height:300px;"></div>
+                <div class="describe-gmap geotag_edit_map_size" id="gmap{{forloop.counter}}"></div>
```
alastair: I wouldn't make a class that specifically exists to mark the _size_ of the map. Why can't you add the size info to the `describe-gmap` class?

Related - we use `gmap` a bit here, is this geo-map or google-map?

templates/sounds/sound_edit.html
ffont: The `<input>` type is `number` and the browser adds some validation (if not a number, undoes the change). If this did not happen I guess there would be a javascript error (unseen by the user) and the map would not update. That would still make sense as the input value would be invalid. To be honest I'm not too much worried by this behaviour and I guess we can trust the input type thing.

templates/accounts/describe_sounds.html
ffont: The event is not triggered until the element uses focus so the map is not constantly changing. This is as described in [JQuery docs](https://api.jquery.com/change/) (and as I have experienced as well) and is also the normal behaviour of the HTML [`onchange` event](https://www.w3schools.com/jsref/event_onchange.asp). 

We might want to use a paste event for the pasting use case, but I don't think this is really necessary as when user clicks somewhere else the map will be updated anyway.

templates/accounts/describe_sounds.html
ffont: The use of `geotag_edit_map_size` is because the same class is used in another place where a similar map is used which does not have `describe-gmap` class. I could probably investigate and refactor to have a common class for both cases, but this code is quite old and considering that we'll refactor all once implementing maps for the new interface I did not consider refactor was worth the time. The whole maps CSS thing might in fact need some refactor. Starting by class names in which `g` stands for Google and we're no longer using Google Maps.

templates/accounts/describe_sounds.html
ffont: After extra conversations and realizing that only updating the map when input loses focus could result in poor user experience, I updated the geotag edit map in sound edit and sound describe so that map re-sync gets triggered with the `input` event (instead of `change` event). I also added a timeout to perform the actual map syncing to avoid too many events triggering map re-syncs. This also works well for the use case in which a user pastes latitude/longitude or zoom values as `input` event is also triggered.

templates/accounts/describe_sounds.html
```diff
@@ -125,6 +125,29 @@
...
+                clearTimeout(centerMapTimers[idx]);
+            }
+
+            centerMapTimers[idx] = setTimeout(function (){
+                var idx0 = idx - 1; // Get map number
+                var lat = $("#id_" + idx0 + "-lon").val();
+                var lon = $("#id_" + idx0 + "-lat").val();
+                var zoom = $("#id_" + idx0 + "-zoom").val();
+                maps[idx].setZoom(parseInt(zoom, 10) - 1);
+                maps[idx].setCenter([parseFloat(lat), parseFloat(lon)]);
```
alastair: if I delete all of the text from a field, I get `lng_lat.js:29 Uncaught Error: Invalid LngLat object: (NaN, 23.8858)`

alastair: I can enter the letter e. It looks like it's so that exponent notation like 1e3 works, but e alone gives an error (because it's not text)

alastair: I can enter numbers greater than 180, and it seems to accept it and move the map, but I no longer see the red marker appear

ffont: Yeah I saw these errors, for some reason browser is not allowing numbers only. I could in fact enter other characters. Maybe it depends on the browser. However there is no problem if one of these values can't be parsed correctly as the map won't update. When the form is later validated in server (or by browser when clicking submit) then the errors will be highlighted.

alastair: the zoom input has up/down arrows on it, but doesn't allow me to go below 11, but I can type a number in and it's fine. I don't know what the acceptable range is here.
If I enter a negative zoom number I get a warning on the console `Geometry exceeds allowed extent, reduce your vector tile buffer size` and it resets to 1.

ffont: >  can enter numbers greater than 180, and it seems to accept it and move the map, but I no longer see the red marker appear

ok, this is an error. maybe doing some basic validation in the `var x = $("#id_" + idx0 + "-x").val();` would help.

> the zoom input has up/down arrows on it, but doesn't allow me to go below 11, but I can type a number in and it's fine. I don't know what the acceptable range is here.
If I enter a negative zoom number I get a warning on the console Geometry exceeds allowed extent, reduce your vector tile buffer size and it resets to 1.

one thing is the range the map accepts and the other the one Freesound accepts (we ask for a minimum zoom). It is normal that the map updates because you also want to see small zooms (while you're searching for some place for example). again, when form is validated later user wil be warned if zoom value is not correct.

alastair: If I move the map with the mouse I get a huge number of decimal places, e.g. `19.464546208718573`. This should be reduced

alastair: when I click/drag the map I can't change the latitude bigger than 36 or lower than -36

alastair: if I zoom with the mouse wheel, the map moves around a bit, depending on where the cursor is, causing the lat/long values to change. if it's possible to make the mapbox map stay still when zooming then that'd be great

alastair: it'd be nice to have instead of the "remove geotag" checkbox, a link which hides the map again

ffont: Most of these comments addressed in e3ecf82f2509410535259b316c52ca19dc8e7a6f

PR #1337, opened 2019-07-04
Correct User Sounds

aframires:
**Issue(s)**
 Where it should say User Sounds it said Similar Sounds

**Description**
 Corrected it to User Sounds

**Deployment steps**:
 None


ffont:
Thanks!
I hope this is the first PR of many to come :)

PR #1341, opened 2019-07-11
Restore the behaviour of using a cached count for the paginator

alastair:
**Description**
We have a Paginator subclass which will take a count of the number of
items being paginated so that we don't have to call count() on a
queryset. We used an implementation detail of django to write to
Paginator._count, but this was changed in a recent version of django so
our paginators were calling qs.count() again.
Add a new count property that bypasses calling .count() if a count is
provided to the paginator.

### DIFF
--- a/sounds/views.py
+++ b/sounds/views.py
@@ -172,7 +172,7 @@ def packs(request):
                      .exclude(is_deleted=True) \
                      .order_by(order)
     tvars = {'order': order}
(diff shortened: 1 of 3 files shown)


ffont:
Also got rid of `CachedCountProxy` as it was only being used for the packs page and the code was ugly.  Packs page is not visited very often and we decided the small optimization of `CachedCountProxy` was not worth the increased complexity of the code.

PR #1342, opened 2019-07-12
Reduce API num queries when serializing sounds

ffont:
**Issue(s)**
None

**Description**
Any resource in the API returning a sound list is expected to make a single query to DB to retrieve all sounds' information in one go. However, if the `url` field was requested, an extra query was being made for each returned sound. This could be up to 150 extra queries if the page size was set to 150. This PR fixes this problem by using the `sound.username` property (needed for building the URL) stored in the sound object (as it comes from `SoundManager.bulk_query`) instead of accessing `sound.user.username` which would trigger extra DB query.

### DIFF
--- a/apiv2/tests.py
+++ b/apiv2/tests.py
@@ -316,7 +316,8 @@ class TestSoundListSerializer(TestCase):
         # Test that we only perform one DB query when serializing sounds regardless of the number of sounds and the
         # number of requested fields
 
-        ContentType.objects.get_for_model(Sound)  # Make sure sound content type is cached to avoid further queries
+        # Make sure sound content type is cached to avoid further queries
+        ContentType.objects.get_for_model(Sound)
 
         field_sets = [
             '',  # default fields
--- a/apiv2/tests.py
+++ b/apiv2/tests.py
@@ -276,3 +276,20 @@ class TestSoundCombinedSearchFormAPI(SimpleTestCase):
         form = SoundCombinedSearchFormAPI(data={'target': target})
         self.assertTrue(form.is_valid())
         self.assertEqual(form.cleaned_data['target'], target)
+
+
+class TestSoundListSerializer(TestCase):
+
+    def test_num_fields(self):
+        # TODO: Test that serializer returns only fields included in fields parameter of the request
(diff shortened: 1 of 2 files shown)


apiv2/tests.py
```diff
@@ -276,3 +279,61 @@ def test_target_valid(self):
...
+
+    def test_num_fields(self):
+        # Test that serializer returns only fields included in fields parameter of the request
+
+        sounds_dict = Sound.objects.dict_ids(sound_ids=self.sids)
+
+        # When 'fields' parameter is not used, return default ones
+        dummy_request = self.factory.get(reverse('apiv2-sound-text-search'), {'fields': ''})
+        serialized_sound = SoundListSerializer(list(sounds_dict.values())[0], context={'request': dummy_request}).data
+        self.assertItemsEqual(serialized_sound.keys(), DEFAULT_FIELDS_IN_SOUND_LIST.split(','))
```
alastair: is this just luck that these things are ordered in the same way? I think when I do similar tests in dunya I wrap each side in sorted()

apiv2/tests.py
```diff
@@ -276,3 +279,61 @@ def test_target_valid(self):
...
+        fields_parameter = ','.join(SoundListSerializer.Meta.fields)
+        dummy_request = self.factory.get(reverse('apiv2-sound-text-search'), {'fields': fields_parameter})
+        serialized_sound = SoundListSerializer(list(sounds_dict.values())[0], context={'request': dummy_request}).data
+        self.assertItemsEqual(serialized_sound.keys(), fields_parameter.split(','))
+
+    def test_num_queries(self):
+        # Test that we only perform one DB query when serializing sounds regardless of the number of sounds and the
+        # number of requested fields
+
+        ContentType.objects.get_for_model(Sound)  # Make sure sound content type is cached to avoid further queries
```
alastair: comment above the line? (personal preference, it's OK to leave for now...)

apiv2/tests.py
ffont: `assertItemsEqual` works with unordered lists so it works here :)

Issue #1344, opened 2019-07-12
Pages where we make repeated database queries for every listed item


alastair:
We've been doing some monitoring of the number of sql queries and time that they take to run on the live freesound site.
We've found at least the following pages that loop through a list of something and then perform at least one additional query for each item in the list. This list isn't comphrensive, and the specific urls are just as an example:

* ~~/people/toiletrolltube/sounds/181311/ - tags_tag for each item~~
* ~~/people/Raffa Jaffa/comments/ - auth_user and sounds_sound for each item in the list~~
* ~~/apiv2/search/text/ - at least analysis rate, bitdepth, sample rate, channels~~,  -- #1351
* ~~/people/CodePanda/bookmarks/category/97703/ - count on bookmark category~~
* ~~/people/Benboncan/packs/4109/downloaders/ - accounts_profile and auth_user~~

PR #1351, opened 2019-07-17
Reduce number of queries when using text search on API (#1344)

alastair:
**Issue(s)**
#1344

**Description**
Reduce the number of queries made to the database on API endpoints. Although we thought that we fixed (and tested) this with https://github.com/MTG/freesound/pull/1342, it seems like there were still some fields that were getting selected from the database.

I saw the extra queries when actually accessing the view - the tests that we added only test the serialiser. I wonder if there's some differences here.

We should go through all of the other API endpoints and confirm the number of queries as well, and add any necessary changes to this PR.

Now we have a significant difference in the query for bulk_query and bulk_query_solr. I wonder if we should try and make this a bit more consistent. Either with a test to ensure that the fields that they return are the same, or reuse the list of fields to query in a variable somehow. Perhaps this isn't actually needed, if we are already selecting exactly the fields that we need for solr and nothing extra.


alastair:
I notice that we already have a test for the fields here: https://github.com/MTG/freesound/blob/7e920bc042706832a2f9b4df40dd4e53478b4e06/sounds/tests.py#L1377
but it treats solr and normal bulk separately, and doesn't check if anything additional has been returned other than what's listed in the test

apiv2/tests.py
```diff
@@ -329,12 +331,14 @@ def test_num_queries(self):
             with self.assertNumQueries(1):
                 sounds_dict = Sound.objects.dict_ids(sound_ids=self.sids[0])
                 dummy_request = self.factory.get(reverse('apiv2-sound-text-search'), {'fields': field_set})
-                SoundListSerializer(list(sounds_dict.values())[0], context={'request': dummy_request})
+                # Call serializer .data to actually get the data and potentially trigger unwanted extra queries
+                _ = SoundListSerializer(list(sounds_dict.values())[0], context={'request': dummy_request}).data
```
alastair: do you need `_ =`?

ffont: Pycharm complains otherwise because the method is not assigned. I can remove it but I thought it was fine.

alastair: I thought so. the alternative is to merge this test with the one that checks that the fields are in the returned data. no problem

alastair:
looks good, let's merge it!

PR #1353, opened 2019-07-17
Perform less queries when showing item flag links

alastair:
**Issue(s)**
#1344

**Description**
There was a lot of additional complexity in the flag_user template tag that wasn't used. By performing a join in the view and passing in some slightly different objects to the template tag, we can avoid running a query per item (comment, message, post) in order to determine if we
should show a report link.

Some testing still required to determine if this is faster in all cases (it's very slow in the 'all comments' view, this could be improved by trying to perform it only over the currently viewed page instead of all comments)

### DIFF
--- a/accounts/templatetags/flag_user.py
+++ b/accounts/templatetags/flag_user.py
@@ -19,32 +19,36 @@
 #
 
 from django import template
+
 from accounts.models import UserFlag
+
 register = template.Library()
 
 
 @register.inclusion_tag("accounts/flag_user.html", takes_context=True)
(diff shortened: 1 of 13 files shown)


ffont:
Ok, closing the PR then. If we identify such queries as being problematic in the future we can have a look again.

ffont:
Re-opening as this PR includes some optimizations that we can still apply.

alastair:
Given the major upgrades in freesound plus the new interface, I propose that we just close the PR and re-profile at a later stage. The changes were quite small anyway

PR #1356, opened 2019-07-17
Latest sounds

### LINKED ISSUE #1343: Remove DelayedQueryExecutor
This code is a bit magic, doesn't support query parameters, and we only use it in 2 places now. Once in
https://github.com/MTG/freesound/blob/cbc83b930beef5f76a2995d2c7c88b21214535b0/accounts/models.py#L234
but we never use the solr=False code path

and once in https://github.com/MTG/freesound/blob/4edd9fc23b6525512a0932353c8e3fdd1380059a/sounds/models.py#L87
but we can improve this by only calling this method from in a cache guard. Our recent changes have also made this query significantly faster, so the use of the delayed query executor is less important.

alastair:
**Issue(s)**
#1343

**Description**
The DelayedQueryExecuter was being evaluated on the sounds page, which
meant that it wasn't serving its purpose. Replace with Manager.raw,
which also isn't evaluated until it's used. Fixes #1343.

Share the latest sound block between the front page and the sounds page so that it is always loaded from cache when available.

Add an index to optimise the query to get the latest sounds so that it's fast even when the cache isn't available.

**Deploy steps**
Run migrations


sounds/models.py
```diff
@@ -84,34 +83,41 @@ def __unicode__(self):
...
+    def latest_additions(self, num_sounds, period_days=5):
         if settings.DEBUG:
-            # If in DEBUG mode it is likely that database does not contain sounds in the requested period. To compensate
-            # that we never use the interval filter in DEBUG mode and instead get the `num_sounds` latest additions.
-            use_interval = False
-        interval_query = ("and greatest(created, moderation_date) > now() - interval '%s'" % period) \
-            if use_interval else ""
+            # In DEBUG mode we probably won't have any sounds from the requested period, so we
+            # see what the most recent sound and go back from then instead
+            latest_sound = Sound.public.order_by('-created').first()
```
alastair: This needs #1352 merged to work properly

templates/sounds/latest_sounds.html
```diff
@@ -0,0 +1,20 @@
+{% load cache %}
+{% load display_sound %}
+
+<!-- recent additions -->
+{% cache 600 latest_sounds %}
```
alastair: 10 minute cache for the sounds

templates/sounds/latest_sounds.html
```diff
@@ -0,0 +1,20 @@
...
+<div id="recent_additions" class="content_box">
+    <h3>Recent Additions</h3>
+    {% for sound in latest_sounds %}
+        {% display_sound sound.id %}
```
alastair: display_sound will make a number of additional queries per sound to get all of the relevant information (username, tags, license). We could modify the latest sounds query to be like get_raw_sound and collect this information as well and then use display_raw_sound. However, we're also caching this for 10 minutes, so maybe it's worth leaving it as-is in exchange for making it a bit easier to understand (I'm sure this comment will come back and get us in the future)

sounds/models.py
```diff
@@ -84,34 +83,41 @@ def __unicode__(self):
 
 class SoundManager(models.Manager):
 
-    def latest_additions(self, num_sounds, period='2 weeks', use_interval=True):
+    def latest_additions(self, num_sounds, period_days=5):
```
alastair: all views use 2 days here, so change the default

templates/sounds/latest_sounds.html
alastair: we decided that this is OK for now, especially with the cache.

alastair:
@ffont ready for review/merge

Issue #1360, opened 2019-07-23
Further slow pages


alastair:
Continuing our work to make fs faster by optimising queries and removing duplicate queries (#1344, #1358, #1348)

The following views have slow queries or perform too many queries:
* /browse/ (sounds.views.sounds), gets `popular_sound_ids` by evaluating a query, causing it to not be cached. Add an index on the dates and consider using display_sound instead of display_raw_sound - even though it'll cause more queries sometimes, we'll be able to use the template cache (https://github.com/MTG/freesound/pull/1356#discussion_r304549120)
* User flag lookups, especially on /people/username/comments and sound view (#1353)
* pack download page looks up license for each sound in the pack individually
* Download page has extra query to license: https://github.com/MTG/freesound/blob/147c6ce0b604db259bf2f6a727e0e38d458fe5c6/sounds/views.py#L363

PR #1361, opened 2019-07-24
Improve speed of most downloaded sounds (#1360)

alastair:
**Issue(s)**
#1360

**Description**
We were evaluating this query in every view in order to use the
display_raw_sound template tag, causing the template cache to not have
an effect. Instead, use display_sound so that we can take advantage of
the cache (we do 5 lookups for tags, but these are small).
Also use greatest(created, moderation_date) in check condition to take
advantage of our indexes.
Query time has reduced from ~800ms to ~300ms, and then we cache this
value for an hour.

Also select related user information on pack downloads to reduce number
of queries


alastair:
I accidentally made the index on Pack.num_downloads when testing which is why I couldn't make the query faster. Actually making it on Sound works as expected :)

sounds/models.py
```diff
@@ -390,7 +390,7 @@ class Sound(SocialModel):
 
     # counts
     num_comments = models.PositiveIntegerField(default=0)  # Updated via django (sound view and delete comment view)
-    num_downloads = models.PositiveIntegerField(default=0)  # Updated via database trigger
+    num_downloads = models.PositiveIntegerField(default=0, db_index=True)  # Updated via database trigger
     avg_rating = models.FloatField(default=0)  # Updated via database trigger
     num_ratings = models.PositiveIntegerField(default=0)  # Updated via database trigger
```
ffont: All these fields have comments "updated via database trigger" which is no longer true right? Or is it?

alastair: good point, fixed

Issue #1367, opened 2019-08-30
Sound "is explicit" overwritten in moderation interface


ffont:
When users describe sounds they have a checkbox to indicate whether sounds are NSFW (called "is explicit"). Also, in the moderation interface moderators have the option to add the NSFW flag to sounds they are moderating.  Unfortunately, if a user marks a sound as NSFW but the moderator does not add the flag as well, the NSFW flag gets removed. This is a bug as if the moderator does not check the NSFW checkbox, we should keep whatever value was introduced by the author.

PR #1368, opened 2019-08-30
Fix bug in sound moderation with is_explicit flags

ffont:
**Issue(s)**
https://github.com/MTG/freesound/issues/1367

**Description**
Fixed bug in sound moderation which caused the "is explicit" information introduced by users to be overwritten by moderators choices even if moderators introduced no preference. 

After this fix the moderators will have the options to 1) keep user preference about "is explicit" flag (default), 2) add the flag to all selected sounds, 3) remove the flag from all selected sounds.

I have manually tested the fixed feature with several sounds and it works good. Also added unit tests.

### DIFF
--- a/tickets/forms.py
+++ b/tickets/forms.py
@@ -74,9 +74,9 @@ MODERATION_CHOICES = [(x, x) for x in
 IS_EXPLICIT_KEEP_USER_PREFERENCE_KEY = "K"
 IS_EXPLICIT_ADD_FLAG_KEY = "A"
 IS_EXPLICIT_REMOVE_FLAG_KEY = "R"
-IS_EXPLICIT_FLAG_CHOICES = ((IS_EXPLICIT_KEEP_USER_PREFERENCE_KEY, "Keep user preference"),
-                            (IS_EXPLICIT_ADD_FLAG_KEY, "Add \"is explicit\" flag"),
-                            (IS_EXPLICIT_REMOVE_FLAG_KEY, "Remove \"is explicit\" flag"))
+IS_EXPLICIT_FLAG_CHOICES = ((IS_EXPLICIT_KEEP_USER_PREFERENCE_KEY, 'Keep user preference'),
+                            (IS_EXPLICIT_ADD_FLAG_KEY, 'Add "is explicit" flag'),
+                            (IS_EXPLICIT_REMOVE_FLAG_KEY, 'Remove "is explicit" flag'))
 
 
 class SoundModerationForm(forms.Form):
--- a/tickets/forms.py
+++ b/tickets/forms.py
@@ -89,9 +89,10 @@ class SoundModerationForm(forms.Form):
                              error_messages={'required': 'No sound selected...'})
     
     is_explicit = forms.ChoiceField(choices=IS_EXPLICIT_FLAG_CHOICES,
-                                    initial=IS_EXPLICIT_FLAG_CHOICES[0][0],
+                                    initial=IS_EXPLICIT_KEEP_USER_PREFERENCE_KEY,
                                     required=True, 
                                     label=mark_safe("<i>Is explicit</i> flag"))
+
 
 class ModerationMessageForm(forms.Form):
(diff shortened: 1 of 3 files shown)


tickets/forms.py
```diff
@@ -70,6 +71,13 @@ def __init__(self, *args, **kwargs):
                        'Return',
                        'Whitelist']]
 
+IS_EXPLICIT_KEEP_USER_PREFERENCE_KEY = "K"
+IS_EXPLICIT_ADD_FLAG_KEY = "A"
+IS_EXPLICIT_REMOVE_FLAG_KEY = "R"
+IS_EXPLICIT_FLAG_CHOICES = ((IS_EXPLICIT_KEEP_USER_PREFERENCE_KEY, "Keep user preference"),
+                            (IS_EXPLICIT_ADD_FLAG_KEY, "Add \"is explicit\" flag"),
```
alastair: It'd be easier here to just use `"""` or `'` as string quotes so that you don't need to escape

tickets/forms.py
```diff
@@ -79,8 +87,11 @@ class SoundModerationForm(forms.Form):
 
     ticket = forms.CharField(widget=forms.widgets.HiddenInput,
                              error_messages={'required': 'No sound selected...'})
-    is_explicit = forms.BooleanField(required=False, label='Sound(s) contain explicit content')
-
+    
+    is_explicit = forms.ChoiceField(choices=IS_EXPLICIT_FLAG_CHOICES,
+                                    initial=IS_EXPLICIT_FLAG_CHOICES[0][0],
```
alastair: maybe more explicit here if you use `IS_EXPLICIT_KEEP_USER_PREFERENCE_KEY`?

tickets/views.py
```diff
@@ -403,12 +404,21 @@ def moderation_assigned(request, user_id):
...
+                    'is_index_dirty': True,
+                    'moderation_state': 'OK',
+                    'moderation_date': datetime.datetime.now()
+                }
+                is_explicit_choice_key = mod_sound_form.cleaned_data.get("is_explicit")
+                if is_explicit_choice_key == IS_EXPLICIT_ADD_FLAG_KEY:
+                    sounds_update_params['is_explicit'] = True
+                elif is_explicit_choice_key == IS_EXPLICIT_REMOVE_FLAG_KEY:
+                    sounds_update_params['is_explicit'] = False
+                elif is_explicit_choice_key == IS_EXPLICIT_KEEP_USER_PREFERENCE_KEY:
```
alastair: I guess you're adding this to make it explicit what happens when the key is set to this value, but I wonder if we could just have the following comment instead of the elif/pass

tickets/tests.py
```diff
@@ -215,3 +218,75 @@ def setUp(self):
...
+
+    def test_remove_is_explicit_flag_for_explicit_sound(self):
+        """Test that when apporving a sound it's 'is_explicit' flag is set to False if the moderator chooses to remove
+        the explicit flag, even if the sound was originally marked as explicit
+        """
+        self.ticket.sound.is_explicit = True
+        self.ticket.sound.save()
+        self._perform_action(u'Approve', is_explicit_flag_key=IS_EXPLICIT_REMOVE_FLAG_KEY)
+        self.ticket.sound.refresh_from_db()
+        self.assertFalse(self.ticket.sound.is_explicit)
```
alastair: newline

tickets/tests.py
```diff
@@ -215,3 +218,75 @@ def setUp(self):
...
+
+
+class TicketTestsIsExplicitFlagFromQueue(TicketTests):
+    """Test that the is_explicit flag of moderated sounds changes in accordance to moderator's choices"""
+
+    def setUp(self):
+        TicketTests.setUp(self)
+        self.ticket = self._create_assigned_ticket()
+
+    def _perform_action(self, action, is_explicit_flag_key=IS_EXPLICIT_KEEP_USER_PREFERENCE_KEY):
```
alastair: if the `is_explicit_flag_key` parameter is always set, I don't think there's a reason to set a default here

PR #1369, opened 2019-09-02
Return error if stripe POST method is called with GET

andrebola:
Users shouldn't get to this page with GET, but if they do now will
raise an error

### DIFF
--- a/donations/tests.py
+++ b/donations/tests.py
@@ -11,6 +11,7 @@ import donations.models
 import sounds.models
 from accounts.models import EmailPreferenceType, UserEmailSetting
 from sounds.models import License
+from collections import namedtuple
 
 
 class DonationTest(TestCase):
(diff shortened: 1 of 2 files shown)


alastair:
It'd be good to see a test for this and the paypal view to check that this works properly

donations/views.py
```diff
@@ -185,6 +185,8 @@ def donation_session_stripe(request):
               cancel_url=return_url_cancel,
             )
             return JsonResponse({"session_id":session.id})
+    # If request is GET return an error 400
```
alastair: maybe update the comment - this will also happen if the form is invalid

andrebola:
Added tests and also now I return form's errors in the same way that the paypal endpoint works.

donations/tests.py
```diff
@@ -178,7 +179,57 @@ def test_annon_donation_stripe(self):
...
+
+        with mock.patch('stripe.checkout.Session.create') as mock_create:
+            mock_create.return_value = session
+            data['amount'] = '0.1'
+            ret = self.client.post("/donations/donation-session-stripe/", data)
+            response =  ret.json()
+            # amount must be greater than 1
+            self.assertTrue('errors' in response)
+
+        with mock.patch('stripe.checkout.Session.create') as mock_create:
```
alastair: you could add a check to the tests that succeed to ensure that create() was called with the correct arguments. I understand that all of the ones that fail won't call this method, right?

donations/views.py
```diff
@@ -185,6 +185,10 @@ def donation_session_stripe(request):
               cancel_url=return_url_cancel,
             )
             return JsonResponse({"session_id":session.id})
+        else:
+            return JsonResponse({'errors': form.errors})
```
alastair: In the case that the form is invalid are you going to continue returning 200, or also 400?

donations/views.py
```diff
@@ -185,6 +185,10 @@ def donation_session_stripe(request):
...
+    # If request is GET return an error 400
+    return HttpResponse(status=400)
```
alastair: the paypal view returns a redirect to the homepage. Are we happy with the difference in behaviour?

andrebola:
I think the difference is that the url for paypal is "by default" in the action of the form. So, in the case that someone has disabled js then he will be redirected. Does this make sense?

donations/views.py
andrebola: 200, like in 'donation_session_paypal'

donations/views.py
andrebola: see comment below

donations/tests.py
andrebola: Done

ffont:
Is this good to be merged now? @alastair @andrebola

alastair:
yes, sure

PR #1370, opened 2019-09-03
Search Result Clustering [old PR]

### LINKED ISSUE #1269: Adding filter in search resets sorting criteria
As reported in https://freesound.org/forum/freesound-project/41317/?page=1#post92500, when adding or removing filters in the search page, the search criteria is lost and set to default. This should not happen.

xavierfav:
This PR is a Work in Progress. The aim is to iteratively discuss the integration of the search result clustering engine into Freesound.

**Search Result Clustering Functioning**
- A user submits a query to the Freesound collection. 
- The client receives the results and makes AJAX request to trigger the clustering of the top 1000 sounds in the results.
- The client periodically (every 0.5s) asks for the clustering result.
- When the clustering is finished, the result is cached for later uses and some cluster facets appear in the client page, with some tags and the number of sounds in each of them. 
- Clicking on a cluster facet will filter the results in order to retrieve only the sounds belonging to the selected cluster. Removing the filter can be done by clicking on the filter facet under the query definition as it is done with the other facets. The only particularity here is that only one cluster filter can be applied at a time.
- Clicking on an icon next to the cluster facet header will open a page with a 2D visualisation that enable quick exploration of the results.

**Notes**
- I've been moving code from `search/views.py` to `utils/search/general_search.py`, I order to enable the clustering engine to reuse Solr related functions for performing queries to the audio database.
- Celery (+ Redis) is used for performing async clustering.
- Config of cache to use Redis backend (easy to make Celery workers and Django sharing the same cache). This is configured in my `local_settings.py` file.
- I fixed #1269 on the way.

**TODO**
- Review code.
- Fix bug: when a cluster too big (>500 items), the Solr query made when applying the cluster facet produces a too long url (increase header size or switch to POST request).
- Make K-Nearest Neighbors Graph construction faster by parallelizing the NN searches. 
---

**How To**
- Add this to your local settings (change port if you need):
```python
# Redis backend cache
CACHES = {
    "default": {
        "BACKEND": "django_redis.cache.RedisCache",
        "LOCATION": "redis://localhost:6380",
        "OPTIONS": {
            "CLIENT_CLASS": "django_redis.client.DefaultClient"
        },
        "KEY_PREFIX": "example"
    }
}

# CELERY STUFF
CELERY_BROKER_URL = 'redis://localhost:6380'
CELERY_RESULT_BACKEND = 'redis://localhost:6380'
CELERY_ACCEPT_CONTENT = ['application/json']
CELERY_TASK_SERIALIZER = 'json'
CELERY_RESULT_SERIALIZER = 'json'

REDIS_HOST = 'localhost'
REDIS_PORT = 6380

# enable the use of the clustering feature.
ENABLE_SEARCH_RESULTS_CLUSTERING = True
```
- Configure the clustering engine:
`cp clustering/clustering_settings.example.py clustering/clustering_settings.py`
Open the file and set `INDEX_DIR` to the folder path were the Gaia dataset is (shared on [google drive](https://drive.google.com/open?id=1gqy1Jfxd71gBuLlhGFwVxwV7UJZ1nBXq)), and `INDEX_NAME_AS` as the name of the dataset file.
- Start Redis: 
`redis-server --port 6380`
- Start Cekery:
`env ENV_CELERY_WORKER=1 celery -A freesound worker -l info`
- Start Django server:
`python manage.py runserver`
- Start Solr as usual.


templates/search/search.html
```diff
@@ -201,13 +204,41 @@
         $("#filter_query").val(cleaned);
     }
 
+    // function for requesting clustering
+    function request_clustering()
+    {
+        $.get({% url 'cluster-sounds' %} + '?{{ query_params }}',
```
xavierfav: This is how I pass the url parameters that were present in the search url.

templates/search/search.html
```diff
@@ -201,13 +204,41 @@
         $("#filter_query").val(cleaned);
     }
 
+    // function for requesting clustering
+    function request_clustering()
+    {
+        $.get({% url 'cluster-sounds' %} + '?{{ query_params }}', 
+                {
+                    // query_params: '{{ query_params | safe }}',
```
xavierfav: This is how I used to pass all the parameters to the clustering view.

utils/search/search_general.py
```diff
@@ -22,19 +22,320 @@
...
+    # schema are '*_b' (for bool), '*_d' (for float), '*_i' (for integer) and '*_s' (for string). At indexing time
+    # we append these suffixes to the ac descirptor names so Solr can treat the types properly. Now we automatically
+    # append the suffices to the filter names so users do not need to deal with that.
+    for name, t in settings.AUDIOCOMMONS_INCLUDED_DESCRIPTOR_NAMES_TYPES:
+        filter_query = filter_query.replace('ac_{0}:'.format(name), 'ac_{0}{1}:'
+                                            .format(name, settings.SOLR_DYNAMIC_FIELDS_SUFFIX_MAP[t]))
+    return filter_query
+
+
+def search_prepare_parameters(request):
```
xavierfav: This function takes the request as parameter.
This way, the code remains the same as before.
However, it is not very piratical, maybe it would be better to explicitly enter the parameters that are relevant, with default None values for the one that were not present in the url query parameters. 
This would imply some changes in the way things are processed. Basically separating a first step that takes care of getting all the query parameters (request.GET.get()).

utils/search/search_general.py
```diff
@@ -22,19 +22,320 @@
...
+        'filter': filter_query,
+        'username': request.user.username,
+        'page': current_page,
+        'sort': sort[0],
+        'group_by_pack': grouping,
+        'advanced': json.dumps(advanced_search_params_dict) if advanced == "1" else ""
+    }))
+
+    # # we send the query parameters in the context for clustering
+    # query_params = {
```
xavierfav: These are the parameters that I used to store.
They correspond to all the parameters need to perform the second step explained before (create Solr query object).

templates/search/clustering_facet.html
```diff
@@ -0,0 +1,8 @@
+<ul>
+    {% for cluster_id, num_results, tags in cluster_id_num_results %}
+        <li class="facet_item">
+            <a href=".?{{url_query_params_string}}&cluster_id={{cluster_id|add:1}}">Cluster #{{cluster_id|add:1}}
```
xavierfav: By including this here, I have a bug that makes the url to grow everytime a clustering facet filter is applied.

search/views.py
```diff
@@ -44,20 +43,51 @@
...
+    query_params, advanced_search_params_dict, extra_vars = search_prepare_parameters(request)
+
+    # get the url query params for later sending it to the clustering engine
+    url_query_params_string = request.META['QUERY_STRING']
+
+    # get sound ids of the requested cluster when cluster filter facet
     cluster_id = request.GET.get('cluster_id', "")
-    in_ids = get_ids_in_cluster(request, cluster_id)
-    query_params.update({'in_ids': in_ids})  # for added cluster facet filter
+    in_ids = _get_ids_in_cluster(request, cluster_id)
```
ffont: If I understand correctly, `in_ids` should only be loaded when there is a clustering faced active. I see in `_get_ids_in_cluster` you check if `cluster_id` is "" and if that is the case you return empty list. I think it would be better to move the logic here. You could do:
```
cluster_id = request.GET.get('cluster_id')  # <- this will be None if cluster_id is not set
in_ids = []
if cluster_id is not None:
    in_ids = _get_ids_in_cluster(request, cluster_id)
```

search/views.py
```diff
@@ -96,14 +126,16 @@ def search(request):
...
+    """Get the sound ids in the requested cluster. Used for applying a filter by id when using a cluster facet.
+    """
     if requested_cluster_id == "":
         return []
     else:
         requested_cluster_id = int(requested_cluster_id) - 1
 
         # results are cached in clustering_utilities, features are: 'audio_fs', 'audio_as', 'audio_fs_selected', 'tag'
-        result = cluster_sound_results(request, 'audio_as')
+        result = cluster_sound_results(request, features='audio_as')
```
ffont: Don't use the magic string here

search/views.py
```diff
@@ -112,13 +144,18 @@ def get_ids_in_cluster(request, requested_cluster_id):
...
+    """Triggers the computation of the clustering, returns the state of processing or the clustering facet.
+    """
     # pass the url query params for later sending it to the clustering engine
     url_query_params_string = request.META['QUERY_STRING']
+    # remove existing cluster facet filter from the params since the returned cluster facets will include 
+    # their correspondinng cluster_id query parameter (done in the template)
     url_query_params_string = re.sub(r"(&cluster_id=[0-9]*)", "", url_query_params_string)
 
-    result = cluster_sound_results(request, 'audio_as')
+    result = cluster_sound_results(request, features='audio_as')
```
ffont: magic string

search/views.py
ffont: Also, where is the `in_ids` used? Inside `perform_solr_query`? if that is the case, add a comment as well `this is used in perform_solr_query to filter the sound ids included in the search

utils/search/search_general.py
```diff
@@ -54,27 +65,36 @@ def search_prepare_sort(sort, options):
...
-    # Solr (e.g. ac_tonality -> ac_tonality_s, ac_tempo -> ac_tempo_i). The dynamic field names we define in Solr
-    # schema are '*_b' (for bool), '*_d' (for float), '*_i' (for integer) and '*_s' (for string). At indexing time
-    # we append these suffixes to the ac descirptor names so Solr can treat the types properly. Now we automatically
-    # append the suffices to the filter names so users do not need to deal with that.
+    """Process the filter to replace humnan-readable Audio Commons descriptor names.
+
+    Used for the dynamic field names used in Solr (e.g. ac_tonality -> ac_tonality_s, ac_tempo -> ac_tempo_i).
+    The dynamic field names we define in Solr schema are '*_b' (for bool), '*_d' (for float), '*_i' (for integer) 
+    and '*_s' (for string). At indexing time we append these suffixes to the ac descirptor names so Solr can 
+    treat the types properly. Now we automatically append the suffices to the filter names so users do not
```
ffont: Wow, my comment was rather bad hehe. I should add an example here and improve the explanation.

utils/search/search_general.py
```diff
@@ -54,27 +65,36 @@ def search_prepare_sort(sort, options):
...
-    """ Process the query parameters.
-        Returns: - the params for search_prepare_query() in a dict
-                 - the variables for the template context
-    """
-    search_query = request.GET.get("q", "")
-    filter_query = request.GET.get("f", "")
-    filter_query_link_more_when_grouping_packs = filter_query.replace(' ','+')
-    cluster_id = request.GET.get('cluster_id', "")
+def split_filter_query(filter_query, cluster_id):
+    """Splits query filters.
```
ffont: This docstring should be a bit more detailed. The really useful information is in the return line. I guess this function is useful for the template to display the filter and the link to remove it right? also the fact that `cluster_id` is treated separately should be explained in the docstring as well instead of only in code comment as it is quite relevant.

utils/search/search_general.py
```diff
@@ -99,18 +119,41 @@ def search_prepare_parameters(request):
...
         filter_query_split.append({
             'name': "Cluster #" + cluster_id,
             'remove_url': filter_query,
         })
 
+    return filter_query_split
+
+
+def search_prepare_parameters(request):
+    """Construct parameters from the search request object.
```
ffont: I'd say this description should be a little bit different. Here the idea is that we parse and process all input parameters and aggregate them into a single dict that then is used to build a solr query. What about the first sentence being:

`Parse and pre-process search input parameters from the request object and return them as a dict`

utils/tests/test_search_general.py
```diff
@@ -0,0 +1,194 @@
...
+from utils.search.search_general import search_prepare_parameters, split_filter_query, search_prepare_query
+from search.forms import SEARCH_DEFAULT_SORT, SEARCH_SORT_OPTIONS_WEB
+
+
+class SearchUtilsTest(TestCase):
+
+    def setUp(self):
+        self.factory = RequestFactory()
+
+    def test_search_prepare_parameters_wo_query_params(self):
```
ffont: What does `wo` mean?

utils/tests/test_search_general.py
```diff
@@ -0,0 +1,194 @@
...
+            'grouping': True,
+            'in_ids': [],
+        }
+
+        query = search_prepare_query(**query_params)
+
+        self.assertEqual(query.params['q'], 'cat')
+        self.assertEqual(query.params['qf'], 'id^10')
+        self.assertEqual(query.params['fq'], 'duration:[1 TO 10] is_geotagged:1')
+        self.assertTrue(query.params['group'])
```
ffont: why not testing the rest of the parameters?

xavierfav: There are many parameters in the query.params dict and I am not sure I understand all of them.
Shall I copy/paste the dictionary obtained when executing the function manually, and just use `assertDictEqual()` for comparing at once the expected (copied) parameters and the resulting ones?

clustering/gaia_wrapper.py
```diff
@@ -104,21 +75,13 @@ def search_nearest_neighbors(self, sound_id, k, in_sound_ids=[], features='audio
...
-                nearest_neighbors = self.gaia_similiarity.view_pca.nnSearch(sound_id, 
-                                                                            self.gaia_similiarity.metrics['pca'], 
-                                                                            filter).get(k)[1:]
-            elif features == 'audio_fs_selected':
-                nearest_neighbors = self.fs_view.nnSearch(sound_id, self.fs_metric, filter).get(k)[1:]
-            elif features == 'audio_ac':
-                nearest_neighbors = self.ac_view.nnSearch(sound_id, self.ac_metric, filter).get(k)[1:]
+            gaia_view = getattr(self, '{}_view'.format(features))
+            gaia_metric = getattr(self, '{}_metric'.format(features))
+            nearest_neighbors = gaia_view.nnSearch(sound_id, gaia_metric, filter).get(k)[1:]
```
ffont: The code of this file has improved A LOT :)
I guess the only thing missing is to treat `tag_features` just like the others :)

ffont:
I'm closing this PR as we've opened a new version of it: https://github.com/MTG/freesound/pull/1388
(further explanations are given in the new PR)

PR #1389, opened 2019-10-04
Search results clustering

### LINKED ISSUE #1269: Adding filter in search resets sorting criteria

xavierfav:
**NOTE:** this PR is a new version of https://github.com/MTG/freesound/pull/1370, originally created by @xavierfav. After a couple rounds of reviews, @ffont messed up when merging from the external remote and the original PR included some commits which were already in master, showing wrong files changed. We decided to make a new branch (this time in the Freesound repo itself) and open a new PR to make it easier to continue with the review process. The previous comments of the review are still accessible at the old PR URL.

---

This PR is a Work in Progress. The aim is to iteratively discuss the integration of the search result clustering engine into Freesound.

**Search Result Clustering Functioning**
- A user submits a query to the Freesound collection. 
- The client receives the results and makes AJAX request to trigger the clustering of the top 1000 sounds in the results.
- The client periodically (every 0.5s) asks for the clustering result.
- When the clustering is finished, the result is cached for later uses and some cluster facets appear in the client page, with some tags and the number of sounds in each of them. 
- Clicking on a cluster facet will filter the results in order to retrieve only the sounds belonging to the selected cluster. Removing the filter can be done by clicking on the filter facet under the query definition as it is done with the other facets. The only particularity here is that cluster facets don't disappear when applying one, and you can quickly switch to another cluster by directly applying another cluster facet filter that will replace the existing one.
- Clicking on the "show" button next to the cluster facet header will open a modal with a 2D visualisation that enables quick exploration of the results.

**Notes**
- I've been moving code from `search/views.py` to `utils/search/general_search.py`, I order to enable the clustering engine to reuse Solr related functions for performing queries to the audio database.
- Celery (+ Redis) is used for performing async clustering.
- Config of cache to use Redis backend (easy to make Celery workers and Django sharing the same cache). This is configured in my `local_settings.py` file.
- I fixed #1269 on the way.

**TODO**
- Review code.
- Fix bug: when a cluster too big (>500 items), the Solr query made when applying the cluster facet produces a too long url (increase header size or switch to POST request). See [here ](https://stackoverflow.com/a/16205568)for a possible solution.
- Test in production settings. Optimizations such as the one done in [this branch](https://github.com/MTG/freesound/tree/src-opti-faiss) are maybe needed in order to make the clustering computation fast enough.
- Use Web Audio API in front for playing sounds instead of html audio.
---

**How To**
- Add this to your local settings:
```python
# enable the use of the clustering feature.
ENABLE_SEARCH_RESULTS_CLUSTERING = True
```
- Install dependencies:
Dependencies are installed within the clustering docker compose container when building containers (`docker-compose build`)
- Configure the clustering engine:
Add the dataset index file in `freesound_data/clustering_index/`. You can download the file from [google drive](https://drive.google.com/file/d/1z0Kz7RCsd5Ua-T7JCjxrjyBB3Oai4Xm5/view?usp=sharing)).
`INDEX_DIR` and `DATASET_FILE` in the `clustering_settings.py `file should not be changed because it is already set to where we mount the `freesound_data/clustering_index/` directory inside the docker container.
- Start freesound as usual:
`docker-compose up`
or 
`docker-compose up web search clustering redis`


clustering/clustering.py
```diff
@@ -0,0 +1,443 @@
...
+            features (str): name of the features to be used for nearest neighbors computation. 
+                Available features are listed in the clustering settings file.
+
+        Returns:
+            (nx.Graph): NetworkX graph representation of sounds.
+        """
+        # Create k nearest neighbors graph        
+        graph = nx.Graph()
+        graph.add_nodes_from(sound_ids_list)
+        k = int(np.ceil(np.log2(len(sound_ids_list))))
```
ffont: Why is `k` chosen in this way?

search/views.py
```diff
@@ -318,12 +121,109 @@ def search(request):
...
+        return JsonResponse({'status': 'failed'}, safe=False)
+    else:
+        return JsonResponse({'status': 'pending'}, safe=False)
+
+    num_sounds_per_cluster = [len(cluster) for cluster in results]
+    classes = {sound_id: cluster_id for cluster_id, cluster in enumerate(results) for sound_id in cluster}
+
+    # label clusters using most occuring tags
+    sound_instances = sounds.models.Sound.objects.bulk_query_id(map(int, classes.keys()))
+    sound_tags = {sound.id: sound.get_sound_tags() for sound in sound_instances}
```
ffont: Oh no! Don't call `get_sound_tags` as this will make an extra query for each sound. Because you used `bulk_query_id`, the sound objects are already loaded with an added `tag_array` which contains a list of the tags. Therefore you can do `sound_tags = {sound.id: sound.tag_array for sound in sound_instances}` and will save many queries.

This makes me think however that the `get_sound_tags` method could be updated and first check if `tag_array` exists in the model. Could be refactored to something like:

```python

    def get_sound_tags(self, limit=None):
        """
        Returns the tags assigned to the sound as a list of strings, e.g. ["tag1", "tag2", "tag3"]

        :param limit: The maximum number of tags to return
        """
        if hasattr(self, 'tag_array'):
            return self.tag_array[0:limit]
        else:
            return [ti.tag.name for ti in self.tags.select_related("tag").all()[0:limit]]
```

(docstring should be updated as well)

I'd suggest that you make this improvement to `get_sound_tags` and then you don't need to change the code here. Adding a unit test to check that using `get_sound_tags` triggers no queries if sound is loaded using `bulk_query_id` method is also a good idea :)

search/views.py
```diff
@@ -318,12 +121,109 @@ def search(request):
...
+    cluster_tags = defaultdict(list)
+    query_terms = {t.lower() for t in request.GET.get('q', '').split(' ')}
+    for sound_id, tags in sound_tags.iteritems():
+        cluster_tags[classes[str(sound_id)]] += [t.lower() for t in tags if t.lower() not in query_terms]
+    cluster_tags_with_count = {k: sorted([(t, tags.count(t)) for t in set(tags)], 
+                                         key=lambda x: x[1], reverse=True)
+                               for k, tags in cluster_tags.iteritems()}
+    cluster_most_occuring_tags = [' '.join(zip(*tags[:3])[0]) for tags in cluster_tags_with_count.values() if len(tags)>2]  # dict values sorted?!
```
ffont: All this logic here is a bit obscure. I think you could simplify it using [`collections.Counter`](https://docs.python.org/2/library/collections.html#collections.Counter). You can pass the raw list of tags (with repeated tags) and do something like `Counter(raw_tags).most_common(3)`. This will return the 3 most common ;)

search/views.py
```diff
@@ -318,12 +121,109 @@ def search(request):
...
+
+
+def clustered_graph(request):
+    """Returns the clustered sound graph representation of the search results.
+    """
+    result = cluster_sound_results(request, features=DEFAULT_FEATURES)
+    graph = result['graph']
+
+    results = sounds.models.Sound.objects.bulk_query_id([int(node['id']) for node in graph['nodes']])
+    preview_urls_name_tags_by_id = {s.id:(s.get_preview_abs_url()[21:],
```
ffont: use the Sound.locations() method :)

search/views.py
```diff
@@ -318,12 +121,109 @@ def search(request):
...
+                                          s.original_filename,
+                                          ' '.join(s.get_sound_tags()),
```
ffont: If you optimized the `get_sound_tags` before then you don't need to change anything here. Otherwise use `s.tag_array`.

search/views.py
```diff
@@ -318,12 +121,109 @@ def search(request):
...
+                                          s.get_absolute_url()) for s in results}
+
+    for node in graph['nodes']:
+        node['url'] = preview_urls_name_tags_by_id[int(node['id'])][0]
```
ffont: Maybe `preview_urls_name_tags_by_id` could be renamed to something like `sound_metadata`?

templates/search/clustered_graph.html
```diff
@@ -0,0 +1,347 @@
+{% load static %}
+<!DOCTYPE html>
+<html>
+
+<head>
+    <title>Force-Directed Layout</title>
```
ffont: This file I won't review as I understand this is a very prototype thing which will not be deployed for Freesound users.

templates/search/search.html
```diff
@@ -201,13 +206,44 @@
...
+    // function for requesting clustering
+    function request_clustering()
+    {
+        $.get({% url 'clustering-facet' %} + '?{{ url_query_params_string | safe }}', 
+                {}, 
+                function( data ) {
+                    if (data.status === 'pending') {
+                        setTimeout(() => {
+                            request_clustering();
+                        }, 500);
```
ffont: Should we have a maximum number of trials here?

utils/search/search_general.py
```diff
@@ -22,19 +22,356 @@
...
+    """
+    # Generate array with information of filters
+    filter_query_split = []
+    if filter_query != "":
+        for filter_str in re.findall(r'[\w-]+:\"[^\"]+', filter_query):
+            valid_filter = True
+            filter_str = filter_str + '"'
+            filter_display = filter_str.replace('"', '')
+            filter_name = filter_str.split(":")[0]
+            if filter_name != "duration" and filter_name != "is_geotagged":
```
ffont: This parsing code is so frightening... It deserves a refactor I think but that's be for a different PR.

utils/search/search_general.py
```diff
@@ -22,19 +22,356 @@
...
+        'description_weight': description_weight,
+        'username_weight': username_weight,
+        'pack_tokenized_weight': pack_tokenized_weight,
+        'original_filename_weight': original_filename_weight,
+        'grouping': grouping,
+    }
+
+    filter_query_link_more_when_grouping_packs = filter_query.replace(' ','+')
+
+    extra_vars = {
```
ffont: Add a message about why are these `extra_vars` used for

utils/tests/test_search_general.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def test_search_prepare_parameters_non_ascii_query(self):
+        # Simple test to check if some non ascii characters are correctly handled by search_prepare_parameters()
+        request = self.factory.get(reverse('sounds-search')+u'?q=Æ æ ¿ É')
+        query_params, advanced_search_params_dict, extra_vars = search_prepare_parameters(request)
+        self.assertEqual(query_params['search_query'], u'\xc6 \xe6 \xbf \xc9')
+
+    def test_split_filter_query_duration_and_facet(self):
+        # We check that the combination of a duration filter and a facet filter (CC Attribution) works correctly.
+        filter_query_string = u'duration:[0 TO 10] license:"Attribution"'
```
ffont: Maybe test a more complex filter here?

clustering/interface.py
```diff
@@ -0,0 +1,89 @@
...
+        Dict: contains either the state of the clustering ('pending' or 'failed') or the resulting clustering classes 
+            and the graph in node-link format suitable for JSON serialization.
+    """
+    query_params, _, _ = search_prepare_parameters(request)
+
+    cache_key = 'cluster-results-{search_query}-{filter_query}-{sort}-{tag_weight}-{username_weight}-{id_weight}-' \
+                '{description_weight}-{pack_tokenized_weight}-{original_filename_weight}-{grouping}'.format(**query_params)
+
+    cache_key += '-{}'.format(features)
+    cache_key = cache_key.replace(' ', '')
```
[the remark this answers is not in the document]
xavierfav: This is to deal with spaces added in the query parameters. 
For instance if a space is added at the end of the query.
I realize that with this, I am assuming for instance that "rim shot" and "rimshot" are the same. Which is not good...

I guess spaces at the end of the query should be removed somewhere else. Maybe in `search_prepare_parameters()`.

clustering/tasks.py
```diff
@@ -0,0 +1,25 @@
+from django.core.cache import cache
+from celery.decorators import task
+
+from clustering_settings import CLUSTERING_CACHE_TIME, CLUSTERING_PENDING_CACHE_TIME
+from . import CLUSTERING_RESULT_STATUS_PENDING, CLUSTERING_RESULT_STATUS_FAILED
+
+
+@task(name="cluster_sounds")
+def cluster_sounds(cache_key_hashed, sound_ids, features):
+    # this import would be better outside the function, but did not work
```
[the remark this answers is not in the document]
xavierfav: This import ensures that the engine is imported after it is re-assigned in __init__.py
There should be a better way to do it to avoid multiple imports that can decrease performance.

Simply importing outside of the scope of the function did not work and gives None value for engine.

clustering/interface.py
```diff
@@ -0,0 +1,89 @@
...
+
+    cache_key_hashed = hash_cache_key(cache_key)
+
+    # check if result is in cache
+    result = cache.get(cache_key_hashed)
+
+    if result and result not in (CLUSTERING_RESULT_STATUS_PENDING, CLUSTERING_RESULT_STATUS_FAILED):
+        # reset the value in cache so that it presists
+        cache.set(cache_key_hashed, result, CLUSTERING_CACHE_TIME)
```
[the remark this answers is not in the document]
xavierfav: I agree with what you say.
But I think we still need a little persistence behavior.
For instance, what happen if a user is relying on a clustering that has been done exactly 23 hours and 59 minutes ago (assuming that the persistence is set to 24 hours). He might search and get the facets, but when applying a clustering filter facet a minute later, we won't be able to retrieve the sounds of the corresponding cluster.

So I guess it can be a good idea to have a separate value `CLUSTERING_RESET_CACHE_TIME` for making the clustering persist while a user is interacting with it.

What do you think?

search/views.py
xavierfav: Wow, this `get_sound_tags` was actually making these extra queries and making the cluster facets take a lot of time to appear on the page. Thanks!

clustering/interface.py
```diff
@@ -0,0 +1,89 @@
...
+from . import CLUSTERING_RESULT_STATUS_PENDING, CLUSTERING_RESULT_STATUS_FAILED
+
+
+def get_sound_ids_from_solr_query(query_params):
+    """Performs Solr query and returns results as a list of sound ids.
+
+    This method performs a single query to Solr with a very big page size argument so all results are 
+    returned at once. A very big page size will make the clustering take a lot of time to be performed.
+    The number of results to retrieve is defined in the clustering settings file as MAX_RESULTS_FOR_CLUSTERING.
+
```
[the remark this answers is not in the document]
xavierfav: There is not a maximum number of results that SOLR can return. 
It seems that it is more a question of how feasible it is to transmitted the data.

Since I am only interested in retrieving the sound ids, I could set `include_facets` to False to reduce the amount of data that Solr returns.

clustering/interface.py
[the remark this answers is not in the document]
xavierfav: This behavior I mentioned might not happen so often, so for now we just don't make the clustering result persisting in cache for more than one day.

media/js/webAudioApiForDesigners.js
```diff
@@ -0,0 +1,219 @@
+// Browser detection required (guess which browsers are at faul =P)
```
alastair: Is this something that you wrote, or did you copy from somewhere else? we need to make sure we update our licensing information with details about this file

freesound/urls.py
```diff
@@ -104,6 +104,9 @@
 
     url(r'^contact/', support.views.contact, name="contact"),
     url(r'^search/$', search.views.search, name='sounds-search'),
+    url(r'^clustering_facet/$', search.views.clustering_facet, name='clustering-facet'),
+    url(r'^clustering_graph_visualisation/$', search.views.cluster_visualisation, name='clustering-graph-visualisation'),
```
alastair: what's the purpose of this page? Who needs to view it? Does it make sense to make it admin-only? Should it have a different URL if that's the case?

utils/search/search_general.py
```diff
@@ -22,19 +22,359 @@
...
+            sort = [sort, "num_ratings asc"]
+        else:
+            sort = [sort]
+    else:
+        sort = [forms.SEARCH_DEFAULT_SORT]
+    return sort
+
+
+def search_process_filter(filter_query):
+    """Process the filter to replace humnan-readable Audio Commons descriptor names.
```
alastair: human

utils/search/search_general.py
```diff
@@ -22,19 +22,359 @@
...
+
+    Used for the dynamic field names used in Solr (e.g. ac_tonality -> ac_tonality_s, ac_tempo -> ac_tempo_i).
+    The dynamic field names we define in Solr schema are '*_b' (for bool), '*_d' (for float), '*_i' (for integer) 
+    and '*_s' (for string). At indexing time we append these suffixes to the ac descirptor names so Solr can
```
alastair: descirptor

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
+import os, sys
```
alastair, line 1: formatting of imports - see https://www.python.org/dev/peps/pep-0008/#imports about grouping

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+import clustering_settings as clust_settings
+
+# The following packages are only needed if the running process is configured to be a Celery worker. 
+# We avoid importing them in appservers to avoid having to install unneeded dependencies.
+if settings.IS_CELERY_WORKER:
+    from gaia_wrapper import GaiaWrapperClustering
+    import numpy as np
+    from sklearn import metrics
+    from sklearn.feature_selection import mutual_info_classif
+    import json
```
alastair, line 15: things from the standard library can be in the main import block. Again, order these alphabetically to make it easier to follow

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+
+        Returns:
+            Tuple(List[List[Float]], List[Int]): 2-element tuple containing a list of evaluation features 
+                and list of classes (clusters) idx.
+        """
+        sound_ids_list, clusters = zip(*classes.items())
+        reference_features = self.gaia.return_sound_reference_features(sound_ids_list)
+
+        # Remove sounds that are not in the gaia reference dataset
+        idx_to_remove = set([idx for idx, feature in enumerate(reference_features) if feature == None])
```
alastair, line 62: use `is` when comparing against None

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+        Removes sounds with missing features.
+
+        Args:
+            classes (Dict{Int: Int}): clustering classes for each sound {<sound_id>: <class_idx>}.
+
+        Returns:
+            Tuple(List[List[Float]], List[Int]): 2-element tuple containing a list of evaluation features 
+                and list of classes (clusters) idx.
+        """
+        sound_ids_list, clusters = zip(*classes.items())
```
alastair, line 58: could you just use classes.keys() and classes.values() here? I _think_ the ordering will be the same

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+                'graph': graph_json,
+                'features': features,
+                'modularity': modularity,
+                'ratio_intra_community_edges': ratio_intra_community_edges,
+                'average_mutual_information': ami,
+                'silouhette_coeff': ss,
+                'calinski_harabaz_score': ci,
+                'communities': communities
+            }
+            json.dump(result, open('{}/{}.json'.format(clust_settings.SAVE_RESULTS_FOLDER, query_params[0]), 'w'))
```
alastair, line 222: os.path.join

alastair: use an open context manager (`with`) to open the file

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+        graph = nx.Graph()
+        graph.add_nodes_from(sound_ids_list)
+        # we set k to log2(N), where N is the number of elements to cluster. This allows us to reach a sufficient number of 
+        # neighbors for small collections, while limiting it for larger collections, which ensures low-computational complexity.
+        k = int(np.ceil(np.log2(len(sound_ids_list))))
+
+        for sound_id in sound_ids_list:
+            try:
+                nearest_neighbors = self.gaia.search_nearest_neighbors(sound_id, k, sound_ids_list, features=features)
+                # edges += [(sound_id, i[0]) for i in nearest_neighbors if i[1]<clust_settings.MAX_NEIGHBORS_DISTANCE]
```
alastair, line 245: can probably remove commented code if it's not used

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+
+        return graph
+
+    def create_common_nn_graph(self, sound_ids_list, features=clust_settings.DEFAULT_FEATURES):
+        """Creates a Common Nearest Neighbors Graph representation of the given sounds.
+
+        Args:
+            sound_ids_list (List[str]): list of sound ids.
+            features (str): name of the features to be used for nearest neighbors computation. 
+                Available features are listed in the clustering settings file.
```
alastair, line 262: value of the settings variable that defines this

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+            for j, node_j in enumerate(knn_graph.nodes):
+                if j > i:
+                    num_common_neighbors = len(set(knn_graph.neighbors(node_i)).intersection(knn_graph.neighbors(node_j)))
+                    if num_common_neighbors > 0:
+                        graph.add_edge(node_i, node_j, weight=num_common_neighbors)
+
+        # keep only k most weighted edges
+        k = int(np.ceil(np.log2(len(graph.nodes))))
+        for node in graph.nodes:
+            ordered_neighbors = sorted(list(graph[node].iteritems()), key=lambda x: x[1]['weight'], reverse=True)
```
alastair, line 284: iteritems isn't in python 3. You could just use items (remember that in py2 this makes a copy, so if it's huge this might cause problems. If you really need an iterable version, use `six.iteritems`

alastair: what's the difference between graph.nodes and graph[node]? Is there a way that allows you to get the name/values in the for loop?

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+            try:
+                neighbors_to_remove = zip(*ordered_neighbors[k:])[0]
```
alastair, line 286: I would prefer that we use a list comprehension here

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+        Returns:
+            Tuple(Dict{Int: Int}, int, List[List[Int]], float): 4-element tuple containing the clustering classes for each sound 
+                {<sound_id>: <class_idx>}, the number of communities (clusters), the sound ids in the communities and
+                the modularity of the graph partition.
+        
+        """ 
+        # Community detection in the graph
+        classes = com.best_partition(graph)
+        num_communities = max(classes.values()) + 1
+        communities = [[key for key, value in classes.iteritems() if value==i] for i in range(num_communities)]
```
alastair, line 313: this should be `value == i` - if you can, get pycharm or something to do a style cleanup of these files.

iteritems - use six or .items

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+
+        Returns:
+            Tuple(Dict{Int: Int}, int, List[List[Int]], float): 4-element tuple containing the clustering classes for each sound 
+                {<sound_id>: <class_idx>}, the number of communities (clusters), the sound ids in the communities and
+                the modularity of the graph partition.
+        
+        """ 
+        # Community detection in the graph
+        classes = com.best_partition(graph)
+        num_communities = max(classes.values()) + 1
```
alastair, line 312: out of interest, is this always going to be the same as len(classes) + 1, or len(classes)? or will class index values possibly be higher than the actual number of classes?

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+        Args:
+            graph (nx.Graph): NetworkX graph representation of sounds.
+
+        Returns:
+            Tuple(Dict{Int: Int}, int, List[List[Int]], None): 4-element tuple containing the clustering classes for each sound 
+                {<sound_id>: <class_idx>}, the number of communities (clusters), the sound ids in the communities and
+                None.
+        """ 
+        communities = [list(community) for community in k_clique_communities(graph, k)]
+        # communities = [list(community) for community in greedy_modularity_communities(graph)]
```
alastair, line 336: remove, or config item to choose between metric

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+        greedy_modularity_communities
```
alastair, line 337: what's this?

clustering/clustering.py
```diff
@@ -0,0 +1,446 @@
...
+        if len(communities) <= 2:
+            return graph, classes, communities, ratio_intra_community_edges
+        min_ratio_idx = np.argmin(ratio_intra_community_edges)
+        sounds_to_remove = communities[min_ratio_idx]
+        graph.remove_nodes_from(sounds_to_remove)
+        del communities[min_ratio_idx]
+        for snd in sounds_to_remove:
+            del classes[snd]
+        del ratio_intra_community_edges[min_ratio_idx]
+        for idx in range(min_ratio_idx, max(classes.values())):
```
alastair, line 370: at https://github.com/MTG/freesound/pull/1389/files#diff-40410ace29c087b6ebccb3110c2a4c6cR312 you do `range(max(classes.values())+1)`. Is this one supposed to be different? I wonder if you could use a separate method with an explicit name in order to make it clear what this is doing

clustering/gaia_wrapper.py
```diff
@@ -0,0 +1,116 @@
+import logging
```
alastair, line 1: import formatting

clustering/gaia_wrapper.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        gaia_dataset = DataSet()
+        gaia_dataset.load(self.__get_dataset_path(gaia_index))
+        gaia_view = View(gaia_dataset)
+        gaia_metric = DistanceFunctionFactory.create(gaia_metric, gaia_dataset.layout(), 
+            {'descriptorNames': gaia_descriptor_names})
+
+        setattr(self, '{}_dataset'.format(features), gaia_dataset)
+        setattr(self, '{}_view'.format(features), gaia_view)
+        setattr(self, '{}_metric'.format(features), gaia_metric)
```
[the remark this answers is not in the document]
alastair: I agree with this comment - I think that frederic's solution is more understandable instead of having to always use getattr/setattr whenever you want to get a feature. It makes the class a bit more self-documenting as well, as you can define this field in the class definition and don't need to read the code to work out that there are additional fields floating around

clustering/gaia_wrapper.py
```diff
@@ -0,0 +1,116 @@
...
+
+        except Exception as e:
+            logger.info(e)
+            return []
+
+    def return_sound_reference_features(self, sound_ids):
+        """Returns the reference features for the given sounds that are used for evaluating the clustering performance.
+
+        The reference features defined in the clustering settings file.
+        They tipically consists of features derived from the sounds' tags.
```
alastair, line 98: typically

clustering/interface.py
```diff
@@ -0,0 +1,96 @@
...
+    Args:
+        query_params (dict): contains the query parameters to replicate the user query.
+    
+    Returns
+        List[int]: list containing the ids of the retrieved sounds.
+    """
+    current_page = 1
```
alastair, line 27: you set this explicitly on the next line. is this needed?

clustering/interface.py
```diff
@@ -0,0 +1,96 @@
...
+    # page would have a clustering performed on filtered results.
+    query_params['filter_query'] = extra_vars['filter_query_non_facets']
+
+    cache_key = 'cluster-results-{search_query}-{filter_query}-{sort}-{tag_weight}-{username_weight}-{id_weight}-' \
+                '{description_weight}-{pack_tokenized_weight}-{original_filename_weight}-{grouping}'.format(**query_params) \
+                .replace(' ', '')
+
+    cache_key += '-{}'.format(features)
+
+    cache_key_hashed = hash_cache_key(cache_key)
```
alastair, line 70: I can't remember if we discussed this or not. I guess we do this because the memcache key size is quite low? In my experience, hashing cache keys is really annoying, because you can't inspect them easily. 
I guess this is all we can do for now, but I think that perhaps we should move to redis for the cache some time so that we can have large keys and remove the hashing

clustering/tasks.py
```diff
@@ -0,0 +1,25 @@
...
+    cache.set(cache_key_hashed, CLUSTERING_RESULT_STATUS_PENDING, CLUSTERING_PENDING_CACHE_TIME)
+
+    try:
+        # perform clustering
+        result = engine.cluster_points(cache_key_hashed, features, sound_ids)
+
+        # store result in cache
+        cache.set(cache_key_hashed, result, CLUSTERING_CACHE_TIME)
+
+    except Exception as e:
```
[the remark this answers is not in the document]
alastair: I don't mind having a blanket exception here, but we should log it to sentry. Does logger.error do this for us, or should we also manually raise an error?

search/views.py
```diff
@@ -318,12 +123,154 @@ def search(request):
...
+    cluster_tags = defaultdict(list)
+
+    # extract tags for each clusters and do not use query terms for labeling clusters
+    query_terms = {t.lower() for t in request.GET.get('q', '').split(' ')}
+    for sound_id, tags in sound_tags.iteritems():
+        cluster_tags[classes[str(sound_id)]] += [t.lower() for t in tags if t.lower() not in query_terms]
+
+    # count 3 most occuring tags
+    # we iterate with range(len(results)) to ensure that we get the right order when iterating through the dict 
+    cluster_most_occuring_tags = [' '.join(zip(*Counter(cluster_tags[cluster_id]).most_common(3))[0])
```
alastair, line 209: can you split this into a few different lines, it's quite difficult to follow exactly what it's doing in a single line.

search/views.py
```diff
@@ -318,12 +123,154 @@ def search(request):
...
+
+    # extract tags for each clusters and do not use query terms for labeling clusters
+    query_terms = {t.lower() for t in request.GET.get('q', '').split(' ')}
+    for sound_id, tags in sound_tags.iteritems():
+        cluster_tags[classes[str(sound_id)]] += [t.lower() for t in tags if t.lower() not in query_terms]
+
+    # count 3 most occuring tags
+    # we iterate with range(len(results)) to ensure that we get the right order when iterating through the dict 
+    cluster_most_occuring_tags = [' '.join(zip(*Counter(cluster_tags[cluster_id]).most_common(3))[0]) 
+                                  if len(cluster_tags[cluster_id])>0 else ''
```
alastair, line 210: `if cluster_tags[cluster_id]` will have the same behaviour as the len check

search/views.py
```diff
@@ -318,12 +123,154 @@ def search(request):
...
+                                  for cluster_id in range(len(results))]
+
+    # extract sound examples for each cluster
+    sound_ids_examples_per_cluster = [map(int, cluster_sound_ids[:7]) for cluster_sound_ids in results]
```
alastair, line 214: what's this 7?

search/views.py
```diff
@@ -318,12 +123,154 @@ def search(request):
...
+    }
+    sound_url_examples_per_cluster = [
+        [(sound_id, sound_urls[sound_id]) for sound_id in cluster_sound_ids] 
+            for cluster_sound_ids in sound_ids_examples_per_cluster
+    ]
+
+    return render(request, 'search/clustering_facet.html', {
+            'results': classes,
+            'url_query_params_string': url_query_params_string,
+            'cluster_id_num_results': zip(range(num_clusters), num_sounds_per_cluster,
```
alastair, line 229: It might be nicer to be a bit more explicit here. Could you make 'cluster_id_num_results' a dictionary, with each of these 4 items as values?

search/views.py
```diff
@@ -318,12 +123,154 @@ def search(request):
...
+    # if yes, filter nodes and links from the graph
+    query_params, _, _ = search_prepare_parameters(request)
+    if query_params['filter_query']:
+        nodes = graph['nodes']
+        links = graph['links']
+        graph['nodes'] = []
+        graph['links'] = []
+        sound_ids_filtered = get_sound_ids_from_solr_query(query_params)
+        for node in nodes:
+            if int(node['id']) in sound_ids_filtered:
```
alastair, line 250: any chance of these `int()`s failing?

templates/search/clustering_facet.html
```diff
@@ -0,0 +1,18 @@
...
+            </a> ({{num_results}}) 
+            <span>
+            &nbsp;<img class="cluster-audio-examples" src="{{media_url}}images/fugue-icons/icons/headphone.png" width="16" height="16" alt="listen" title="click to skip" />
+            &nbsp;<img class="cluster-link-button" cluster-id="{{cluster_id}}" 
+                src="{{media_url}}images/fugue-icons/icons/external.png" width="16" height="16" alt="listen" title=""
+            /></span>
+            <div>{{tags}}</div>
+        </li>
+    {% endfor %}
+</ul>
```
alastair, line 18: newline

templates/base.html
```diff
@@ -7,6 +7,9 @@
 <html xmlns:fb="http://www.facebook.com/2008/fbml" xmlns:og="http://ogp.me/ns#" xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en" dir="ltr">
 
     <head>
+        {% block extra_headers %}
+        {% endblock %}
```
alastair, line 11: can we consider loading external js after the body, to prevent loading slowdowns?

templates/search/search.html
```diff
@@ -4,6 +4,18 @@
 {% load paginator %}
 {% load search %}
 {% load util %}
+{% load static %}
+
+{% if clustering_on %}
+{% block extra_headers %}
+    <link rel="stylesheet" type="text/css" href="{{media_url}}css/graph-clustering.css"/>
+    <script src="//ajax.googleapis.com/ajax/libs/jquery/2.1.3/jquery.min.js"></script>
```
alastair, line 12: does this load a second jquery in addition to the one that is normally loaded on all pages? We should probably upgrade that one instead...

templates/search/search.html
```diff
@@ -77,6 +105,18 @@
         	}
         }
 
+        // Ajax request clustering
+        {% if clustering_on %}
```
alastair, line 109: I'm not sure about these types of template tags directly in javascript. I would prefer:
```js
const clustering_on = {{clustering_on}};
...
if (clustering_on) {
   ...
}
```

templates/base.html
xavierfav: I am getting a js error if I place this block after.
Maybe it is possible to split the extra files in two sets, loading some in the header, and others at the end of the body.

templates/search/search.html
xavierfav: Then I need to do something like this:

```js
const clustering_on = "{{clustering_on}}";
...
if (clustering_on == "True") {
    ...
}
```

Which does not look very good.

clustering/tasks.py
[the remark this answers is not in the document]
xavierfav: With default configuration the message should be logged to Sentry ([doc](https://docs.sentry.io/platforms/python/logging/)).

templates/search/search.html
alastair: you could make the value of `{{clustering_on}}` a string that looks like a javascript boolean. So the final output javascript will look like
```js
const clustering_on = true;
```

templates/base.html
alastair: we should look further into this error. Do you have more details about what the problem is?

xavierfav:
Clustering 1000 sound items takes around 10 sec to be computed on our test server.

When requests are received while the clustering container is down (every second the client requests the clustering with AJAX in order to trigger the clustering or receive the result), they will eventually all will be requested when the container is up again. The caching system used is not made properly to avoid repeating the same tasks several times if they are send one after the other very quickly. Perhaps a wrapper around [`cluster_sound_results`](https://github.com/MTG/freesound/blob/search-results-clustering/clustering/interface.py#L40) should be added for managing task keys in the cache.
The Celery documentation proposes [this approach](https://docs.celeryproject.org/en/latest/tutorials/task-cookbook.html#ensuring-a-task-is-only-executed-one-at-a-time) for ensuring that a task is only performed once. However, their approach is done with memcache, not completely sure it would work with redis-based cache (it is said: _In order for this to work correctly you need to be using a cache backend where the .add operation is atomic. memcached is known to work well for this purpose._)

templates/search/search.html
```diff
@@ -201,13 +242,157 @@
...
+                            }
+                        });
+
+                        $('#close-modal-button').click(function () {
+                            closeGraphModal();
+                        })
+
+                        // request clustered graph
+                        // set graph as a global variable
+                        $.get("{% url 'clustered-graph-json' %}"+"?{{url_query_params_string | safe}}", {
```
xavierfav, line 358: Here, I use `| safe` because I need to have the url unescaped. 
Did not find another way to do that, but it may not be recommended to do that.

xavierfav:
Several changes have been made based on discussion with @ffont. 
Among other minor things:

- refactor the audio feature store (remove Gaia stuff, use redis based cache db)
- improvement to the way the query filter string is being parsed (remove regex and use pyparsing)

utils/search/search_general.py
```diff
@@ -386,17 +404,17 @@ def search_prepare_query(search_query,
     return query
 
 
-def remove_facet_filters(filter_query):
+def remove_facet_filters(parsed_filters):
     """Process query filter string to keep only non facet filters
 
     Useful for being able to combine classic facet filters and clustering. Addtionaly, it returns
```
ffont, line 410: Maybe add some explanation of what a "facet filter" is? I assume these are the filters displayed as facets in the search interface. Maybe just adding this info would be enough. Also give an intuition about why this is "Useful for being able to combine classic facet filters and clustering"

ffont (commented):
Hi, I just left a couple more minor comments requesting small changes. Once this is done I'll merge. Thanks!

ffont (approved):

utils/search/search_general.py
```diff
@@ -85,41 +88,43 @@ def search_process_filter(filter_query):
     return filter_query
 
 
-def split_filter_query(filter_query, cluster_id):
-    """Parses and pre-process search filter parameters and returns the filters' information.
+def split_filter_query(filter_query, parsed_filters, cluster_id):
+    """Pre-process parsed search filter parameters and returns the filters' information.
 
     This function is used in the search template to display the filter and the link when removing them.
     The clustering facet filter is in a separate query parameter. It facilitates to distinguish them from
```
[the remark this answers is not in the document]
xavierfav: `cluster_id `is indeed not part of the filters. It is passed using an additional url query parameter.
I think having it separated from the other used to make sense when we did not have a good parser for the filters. Now it is true that the comment does not really make sense, because it is not really facilitating anything. Well, maybe it is still good to have the clustering filter very explicitly separated, because it's a different kind of filter that is not directly passed to Solr as such. I will update the docstring.

Issue #1382, opened 2019-09-27
Store user deletion requests in DB


ffont:
To semi-automate compliance with GDPR, we should deletion requests that we receive via email into a database. As a first implementation of this, we should create a `UserDeletionRequest` model and manage it through Django admin so we can add rows when needed and change their status. A suggestion of the fields this model should contain:

- user (nullable, so that if user is finally deleted this is set to null but object not deleted)
- username
- email
- date request received
- status of the request

Possible statuses could be (to be determined according to GDPR-compliant workflow):

- received
- waiting for user action
- closed (user has been deleted)
- closed (request cancelled)

In the admin list view we should also add some custom field that checks wether we have taken action in a timely manner and highlight requests that require attention.

Issue #1381, opened 2019-09-27
Don't allow reuse of usernames


ffont:
When a user account is fully deleted (the DB row removed) then it is technically possible that another user takes her username. We should forbid that. We can do that by checking against the `DeletedUser` table that should be created as described in #1379.

Issue #1380, opened 2019-09-27
Don't show downloaded sounds and similar pages for anonymized users


ffont:
When a user is anonymized we completely unlink it's personal data from the account and remove the links to user pages such as *downloaded sounds*, *bookmarks*, etc. We should make sure that these pages are not directly accessible by their URL and return a "deleted user page" or something like that.

Issue #1379, opened 2019-09-27
Remove OldUsername objects when users delete their accounts


ffont:
When users or admins delete user accounts but a DB row is still present (because user is anonymized instead of fully deleted), we should make sure that all `OldUsername` entries relating to that row are also deleted. This is to avoid redirects from URLs containing the username prior deletion to the anonymized user page.

PR #1390, opened 2019-10-17
Delete users refactor

ffont:
**Issue(s)**
Fixes https://github.com/MTG/freesound/issues/1378
Fixes https://github.com/MTG/freesound/issues/1379
Fixes https://github.com/MTG/freesound/issues/1380
Fixes https://github.com/MTG/freesound/issues/1381
Fixes https://github.com/MTG/freesound/issues/1382

**Description**
This PR brings some refactoring to user deletion behaviour as described in the issues listed above. As a summary:

- When users are deleted, we now create a `DeletedUser` object that stores some metadata including the reason for deletion. The original user object is normally anonymized and also kept in DB. In some cases (e.g. when deleting spammers) the original user object is deleted as well. I refactored `Profile.delete_user` method to make it more clear and to create these objects.
- When a user is anonymized, we now delete the associated `OldUsername` objects as well so that no redirects are made.
- Don't allow the reuse of usernames taken by deleted users. To do that we check against the `DeletedUser` table. This will not include old usernames of deleted users (if any) as `OldUsername` objects are deleted, but I don't think this is a big issue.
- When a user is anonymized, return 404 pages if someone's trying to visit pages like "downloads by user X". Only the main user profile page with the list of sounds (if any) and the individual sounds are allowed. This is the list of pages that now will return 404 for anonymized users:
  * list of sounds and packs downloaded by the user
  * user bookmarks
  * comments by a user and comments for a user 
  * user followers and following users/tags
  * geotags by user
- If `Sound` or `Pack` objects are to be deleted, the refactored `delete_user` method now also makes sure that `num_sound_downloads` and `num_pack_downloads` won't violate DB constraints when updated. We do that by using f-expressions and "greater" function in the update query.
- This PR adds a new `UserDeletionRequest` which is used to save a record when a user requests the deletion of her account via email or via the website itself. `UserDeletionRequest` are created and managed via the Django admin or automatically when a users deletes himself through the settings page. The status and evolution of the request should be done manually except for when the user is finally deleted. In that case the status is automatically updated by the `Profile.delete_user` method. In the future we could have a management command that could check the status of existing `UserDeletionRequest` objects and tell us when we have to take any actions.
- After this PR is merged, users deletion will always be done asynchronously if triggered from the admin panel or from the settings page when a user deletes herself. We observed this functionality fails some times. In this PR there are some fixes that will prevent it from failing in the future, but we should keep an eye to see if unexpected problems still occur. Also, sometimes the gearman worker seems to stop working after some time and needs restarts. In this PR we add a `check_async_deleted_users` management command that will double check that users that should have been deleted, were indeed deleted. We know what users should have been deleted by querying the `UserDeletionRequest` table mentioned above.

**Deployment steps**:
This PR includes some DB changes but everything is handled nicely in migrations.
Consider the option of adding `check_async_deleted_users` to crontab and updating one of the graylog panels to show that reuslts of this command in a time series. We can also set an alarm if the number of users not deleted goes above 0.


ffont:
Hi, I'd say this PR is no longer in draft mode. I'd be nice if it can be reviewed. We can talk about it when we meet at the office.

accounts/models.py
```diff
@@ -350,55 +381,123 @@ def num_sounds_pending_moderation(self):
...
+                    email=self.user.email,
+                    date_joined=self.user.date_joined,
+                    last_login=self.user.last_login,
+                    reason=deletion_reason)
+
+            # If UserGDPRDeletionRequest object(s) exist for that user, update their status and deleted_user property
+            UserGDPRDeletionRequest.objects.filter(user_id=self.user.id)\
+                .update(status="de", deleted_user=deleted_user_object)
+
+            # If Sound/Pack objects will be deleted, we need to make sure that num_sound_downloads and
```
ffont: This bit could be simplified if we somehow removed the >=0 constrain in the count fields...

ffont:
To add: make sure no sections like "followers" are shown in deleted user profile pages.

accounts/models.py
```diff
@@ -350,55 +381,123 @@ def num_sounds_pending_moderation(self):
...
+            # num_pack_downloads fileds from users that downloaded packs or sounds from the user being delted are
+            # correct. Otherwise this will make the deletion process to fail (violating DB constraints)
+            if delete_user_object_from_db or remove_sounds:
```
alastair, line 459: If this is a popular sound, then we'll perform a lot of queries here, 2 per downloader - one to get the count and one to save it. If a user has download a sound twice, we'll process it twice.
We should
 * do a unique select when getting the downloaders
 * Do only 1 query to get all the download counts (i.e. `user_id__in=[sound_downloaders]`), to minimise the number of queries that we make here

accounts/models.py
```diff
@@ -350,55 +381,123 @@ def num_sounds_pending_moderation(self):
...
+                self.user.username = 'deleted_user_%s' % self.user.id
+                self.user.email = 'deleted_user_%s@freesound.org' % self.user.id
+                self.has_avatar = False
+                self.is_anonymized_user = True
+                self.user.set_unusable_password()
+
+                self.about = ''
+                self.home_page = ''
+                self.signature = ''
+                self.geotag = None
```
alastair, line 487: I worry that if we add a new field to the profile that we'll forget to update it here. Anything that we can do?

accounts/models.py
```diff
@@ -350,55 +381,123 @@ def num_sounds_pending_moderation(self):
...
+                self.save()
+                self.user.save()
+
+                # Remove existing OldUsername objects so there are no redirects to the anonymized/deleted user page
+                OldUsername.objects.filter(user=self.user).delete()
+
+                # Remove sounds and packs if requested
+                if remove_sounds:
+                    Sound.objects.filter(user=self.user).delete()
+                    Pack.objects.filter(user=self.user).update(is_deleted=True)
```
alastair, line 498: why do we actually delete the sounds, but only mark the packs as deleted with a flag?

accounts/models.py
```diff
@@ -557,3 +656,21 @@ class Meta:
...
+
+class UserGDPRDeletionRequest(models.Model):
+    """
+    This model is used to store information about deletion requests received via email and to help comply with GDPR
+    """
+    user = models.ForeignKey(User, null=True, on_delete=models.SET_NULL, related_name='gdpr_deletion_requests')
+    deleted_user = models.ForeignKey(DeletedUser, null=True, on_delete=models.SET_NULL)
+    username = models.CharField(max_length=150, null=True, blank=True)
+    email = models.CharField(max_length=200)
+    date_request_received = models.DateTimeField(auto_now_add=True)
```
alastair, line 669: Would it be useful to also have date fields for the processes that we go through? Maybe we could just have a jsonb field or a char array field for log messages, and just append messages to it

accounts/tests/test_user.py
```diff
@@ -178,73 +180,209 @@ def test_user_delete_make_invalid_password(self):
...
+        delete_sound_from_gaia.assert_has_calls(calls, any_order=True)
+
+    def test_user_delete_include_sounds_using_web_form(self):
+        # Test user's option to delete user account including the sounds
 
-    def test_user_delete_using_form(self):
-        # This should set user's attribute active to false and anonymize it
         user = self.create_user_and_content(is_index_dirty=False)
-        a = self.client.login(username=user.username, password='testpass')
+        self.client.login(username=user.username, password='testpass')
```
alastair, line 287: can use force_login here to not need a password

accounts/tests/test_user.py
```diff
@@ -178,73 +180,209 @@ def test_user_delete_make_invalid_password(self):
...
+
+    def test_fail_user_delete_include_sounds_using_web_form(self):
+        # Test delete user account form with wrong password does not delete
+
+        # This should try to delete the account but with a wrong password
+        user = self.create_user_and_content(is_index_dirty=False)
+        self.client.login(username=user.username, password='testpass')
+        form = DeleteUserForm(user_id=user.id)
+        encr_link = form.initial['encrypted_link']
+        self.client.post(reverse('accounts-delete'),
```
alastair, line 329: we could test that the returned text doesn't have "you've deleted your stuff", or that it includes "your password is wrong", or whatever other message you put here

accounts/forms.py
```diff
@@ -147,6 +147,34 @@ def __init__(self, required=True):
             required=required)
 
 
+def username_taken_by_other_user(username):
```
alastair, line 150: nice

utils/username.py
```diff
@@ -57,11 +57,45 @@ def redirect_if_old_username_or_404(func):
...
-        request.parameter_user = new_user
+
+        # Save user object in the request so it can be used by the view and/or other decorators
+        request.parameter_user = user
+        return func(request, *args, **kwargs)
+
+    return inner
+
+
+def raise_404_if_user_is_deleted(func):
```
alastair, line 79: It looks like we have to apply these two decorators in the correct order, and on very specific views. I wonder if it makes sense to start a development guide that says things like this. For example, this could be in a section called "URLs that include a username" or something

accounts/models.py
```diff
@@ -350,55 +381,123 @@ def num_sounds_pending_moderation(self):
                 ~Q(sound__moderation_state='OK') &\
                 ~Q(status='closed')))
 
-    def get_info_before_delete_user(self, remove_sounds=False, remove_user=False):
+    def get_info_before_delete_user(self, include_sounds=False, include_other_related_objects=False):
```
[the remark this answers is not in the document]
ffont: hah, I did not really invert the behaviour, it only looks like that because of the variable names. The method gets info about the things that will be removed. If you use `include_sounds`, it will list sounds as well. Before, using `remove_sounds` meant that the deletion action would remove sounds, so I want the info for the case of sounds being removed (so include sounds). I changed it to be clearer. There are 4 usages of that method and they use different settings for the parameters so parameters are needed.

accounts/models.py
```diff
@@ -350,55 +381,123 @@ def num_sounds_pending_moderation(self):
...
-        else:
-            Sound.objects.filter(user=self.user).update(is_index_dirty=True)
+        # Run all deletion operations in a single transaction so if there is an error we don't create duplicate
+        # DeletedUser objects
+        with transaction.atomic():
+
+            # Create a DeletedUser object to store basic information for the record (first check if it
+            # already exists because user was deleted previously but db object preserved
+            try:
+                deleted_user_object = DeletedUser.objects.get(user=self.user)
```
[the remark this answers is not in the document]
ffont: I think I did not do it because in the "create" case I have to provide more info and I don't know if `get_or_create` will use all the fields I pass in the query for deciding if an object actually exists. Also I think this way the code is quite clear in this case.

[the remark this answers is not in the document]
alastair: this is in the [documentation for get_or_create](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#get-or-create) - you pass in the filter objects as regular parameters, and pass in new fields to fill in as the `defaults` parameter, so it's possible to do this behaviour. No problem

accounts/models.py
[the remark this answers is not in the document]
alastair: ok, that's great that you clarified it then, thanks

accounts/models.py
ffont: You mean to store when `status` changes right? Makes sense, it could just be a text field I guess with `date status` per line. And then have a post save hook in the `UserGDPRDeletionRequest` that will automatically populate that field when `state` changes. If this sounds good I'll go ahead with this idea.

alastair: yes, right. I'm not sure what the best data type is here. We could have an actual array of messages, instead of manually constructing them, it might be clearer and less prone to errors.
I guess we have something similar with the processing logs.

accounts/tests/test_user.py
ffont: Refactored all fs tests to use `force_login`. This is something I wanted to do some time ago... e153bcbefcd1fd1de30d11f65cb2c4b7082eb515

accounts/tests/test_user.py
ffont: Addressed e769296c26d70aba8fadfd232402727c7dfc014b

utils/username.py
ffont: Good idea, added documentation 56df8a8d211b99c0bf2b16708dea75f6be518aa9

ffont:
Thanks for the review @alastair. I made some changes and left comments where appropriate. There's a little more implementation to do depending on your answers to my comments. I'll wait for further feedback.

accounts/models.py
alastair: did we do something with this in the end? Open a separate ticket for it?

accounts/models.py
alastair: didn't see a reply from you about this

accounts/models.py
alastair: didn't see a reply from you about this

accounts/models.py
alastair: didn't see a reply from you about this

accounts/admin.py
```diff
@@ -155,91 +108,145 @@ def clean_username(self):
 
 class FreesoundUserAdmin(DjangoObjectActions, UserAdmin):
     search_fields = ('=username', '=email')
-    actions = (disable_active_user, disable_active_user_preserve_sounds, )
+    actions = ()
     list_display = ('username', 'email')
     list_filter = ()
     ordering = ('id', )
     show_full_result_count = False
     form = AdminUserForm
```
alastair, line 116: the `clean_username` method in this form doesn't use `username_taken_by_other_user`, therefore it won't check deleted users

accounts/admin.py
```diff
@@ -155,91 +108,145 @@ def clean_username(self):
...
             gm_client.submit_job("delete_user",
-                    json.dumps({'user_id': obj.id, 'action': DELETE_USER_KEEP_SOUNDS_ACTION_NAME}),
-                wait_until_complete=False, background=True)
+                                 json.dumps({'user_id': obj.id, 'action': DELETE_SPAMMER_USER_ACTION_NAME}),
+                                 wait_until_complete=False, background=True)
             messages.add_message(request, messages.INFO,
-                                 'User \'%s\' will be soft deleted asynchronously. Comments and other content '
-                                 'will appear under anonymised account' % username)
+                                 'User \'%s\' will be deleted asynchronously including sounds and all of its related'
+                                 'content.' % username)
```
alastair, line 202: double-check here. because you're using the split line strings, you need the space between the words related and content

accounts/admin.py
```diff
@@ -248,9 +255,45 @@ class OldUsernameAdmin(admin.ModelAdmin):
...
+            return '-'
+        return '<a href="{0}" target="_blank">{1}</a>'.format(reverse('admin:accounts_deleteduser_change', args=[obj.deleted_user_id]),
+                                                              obj.deleted_user.username)
+    deleted_user_link.allow_tags = True
+    deleted_user_link.admin_order_field = 'deleted_user'
+    deleted_user_link.short_description = 'Deleted user object'
+
+
+class DeletedUserAdmin(admin.ModelAdmin):
+    readonly_fields = ['user']
```
alastair, line 291: I added this, otherwise the page fails to load (as it's trying to get all of the users to put in a dropdown for the user relation).
However, I think it might be a good idea to make _all_ of these fields readonly... what do you think? There should be no need to ever edit this object.

accounts/management/commands/monitor_async_deleted_users.py
```diff
@@ -0,0 +1,127 @@
...
+from accounts.admin import DELETE_SPAMMER_USER_ACTION_NAME, FULL_DELETE_USER_ACTION_NAME, \
+    DELETE_USER_DELETE_SOUNDS_ACTION_NAME, DELETE_USER_KEEP_SOUNDS_ACTION_NAME
+from accounts.models import Profile
+
+logger = logging.getLogger('console')
+
+
+class Command(BaseCommand):
+    help = 'Gets a list of users that should have been asynchronously deleted recently and makes sure that these were' \
+           'indeed deleted. The list of users is taken from Graylog "Delete user request" log messages.'
```
alastair, line 37: I wonder if we can improve this in the future? Our Graylog setup is pretty stable, except for the times when it isn't. Do we think that we might ever have a situation where an account is deleted, gearman is down, and graylog doesn't record the log message?

I think that my solution for ensuring that we don't lose these messages would be to make a new table, DeleteRequest, and then have the gearman job remove the row from this table when completed.

This isn't something that I think needs to be changed in this PR, but we could make a ticket for it to improve later.

accounts/management/commands/monitor_async_deleted_users.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def add_arguments(self, parser):
+
+        parser.add_argument(
+            '-n', '--n_days',
```
alastair, line 42: just fyi, I don't need this to be changed: typically "long" options only use `-` in them, so maybe --n-days or --num-days

accounts/management/commands/monitor_async_deleted_users.py
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        while results_remaining:
+            page = (params['offset'] / messages_per_page) + 1
+            logger.info('Getting user delete requests info (page {0})'.format(page))
+
+            try:
+                total_results, new_messages = get_messages_for_params(params)
+            except requests.HTTPError as e:
+                logger.error('Can\'t access Graylog server. Error: ' + str(e))
```
alastair, line 77: Prefer to use `"` instead of escaping quotes inside the string. If I need both `"` and `'` I will triple-quote the string

alastair: Interesting bug here which I found recently in AcousticBrainz. If the message in `e` contains a unicode string, this will fail to render. Instead, use `u'got an error: {}'.format(e)`

accounts/management/commands/monitor_async_deleted_users.py
```diff
@@ -0,0 +1,127 @@
...
+                if User.objects.filter(id=user_id).exists():
+                    # If user object still exists with that id, action was not properly taken
+                    users_not_properly_deleted.append((user_id, 'User object deletion was requested because of being '
+                                                                'spammer, but User object still exists'))
+
+            elif DELETE_USER_DELETE_SOUNDS_ACTION_NAME in msg:
+                if Profile.objects.filter(user_id=user_id, is_anonymized_user=False).exists():
+                    # If user object still exists with that id and is not anonymized, action was not properly taken
+                    # Note that we use Profile objects as a proxy for User
+                    users_not_properly_deleted.append((user_id, 'User anonmymization (with sound deletion) was '
```
alastair, line 102: anonymization

accounts/admin.py
```diff
@@ -155,91 +108,145 @@ def clean_username(self):
...
+        tvars = {'users_to_delete': [], 'type': 'full_delete'}
+        tvars['users_to_delete'].append(user_info)
+
         return render(request, 'accounts/delete_confirmation.html', tvars)
-    delete_preserve_sounds.label = "Soft delete user (preserve sounds)"
-    delete_preserve_sounds.short_description = disable_active_user_preserve_sounds.short_description
 
-    change_actions = ('full_delete', 'delete_include_sounds', 'delete_preserve_sounds', )
+    full_delete.label = "Full delete"
+    full_delete.short_description = 'Completely delete user from db'
```
alastair, line 247: I think this one should have a bit more information. I see the comment in the gearman worker says "WARNING: This functions creates no DeletedSound nor DeletedUser objects". It would be nice to have something similar here, and maybe an indicator about when this should actually be used? Can you give an example of when we might want to use this instead of "Delete User and Sounds"  ?

accounts/views.py
```diff
@@ -1051,7 +1051,8 @@ def delete(request):
         else:
             delete_sounds =\
                 form.cleaned_data['delete_sounds'] == 'delete_sounds'
-            request.user.profile.delete_user(remove_sounds=delete_sounds)
+            request.user.profile.delete_user(remove_sounds=delete_sounds,
```
alastair, line 1054: This will run in the request, which means that if it takes a long time to run, there's a possibility of a timeout.
What options do we have here? If we do it in the background, then the user object will still exist. We could mark password as invalid or mark the account as inactive, and this would prevent the user from logging in. We can always show a message saying "we will process your deletion request and your account/sounds will be deleted shortly". This is especially important if the user has a popular sound and we need to rewrite user download counts... this process could take a long time

accounts/admin.py
ffont: Addressed in 0816d32316556e2e6f1da7a94869c259b4c47bef

accounts/admin.py
ffont: Addressed in 0f346f6f1b5b4542a9282530962111bc54b3c1e8

accounts/admin.py
ffont: Addressed in fa0c6c256904c9de2d60b7f7beba8081da961e5f

accounts/management/commands/monitor_async_deleted_users.py
ffont: Addressed 671fbe4de71dc952484515ca669bbfd7878d1b25

accounts/management/commands/monitor_async_deleted_users.py
ffont: Addressed 671fbe4de71dc952484515ca669bbfd7878d1b25

accounts/management/commands/monitor_async_deleted_users.py
ffont: Addressed 671fbe4de71dc952484515ca669bbfd7878d1b25

accounts/admin.py
ffont: I guess the answer is "ideally we should never use that". So maybe I just remove the option for now and in the advent of an unexpected case that requires it, we can do it using the console or we can re-enable the option.

accounts/views.py
ffont: You're completely right. We should show a message like you say and send the job to gearman to make it work like async user delete in the admin. I'm still considering to do the refactoring of storing delete requests in the DB (as you suggested before) in this PR. If I do that, then it will work using the DB table for deletion requests.

ffont: Similar fix should be done for sound deletion, but this will go in Sound deletion refactoring PR.

ffont: Addressed in 7f3e0a8c0b5f7cee8f500fdaf998787f937080e2 (using gearman to delete the user async)

accounts/admin.py
ffont: Addressed in f07a66a16c060f948b49ed849b47c04692683f9d (removed the non-necessary option)

accounts/models.py
ffont: addressed in e2f191ea8e7b657aa533ca7434746989292211d4

accounts/models.py
ffont: Addressed in c65491773db20f10928cd6c2b7288f0f52292c86. Now we use PG's `Greatest` to make sure the count fields are not set to < 0 and therefore we avoid the need of updating the fields before deletion.

accounts/models.py
```diff
@@ -350,55 +381,123 @@ def num_sounds_pending_moderation(self):
...
+                    user=self.user,
+                    username=self.user.username,
+                    email=self.user.email,
+                    date_joined=self.user.date_joined,
+                    last_login=self.user.last_login,
+                    reason=deletion_reason)
+
+            # If UserGDPRDeletionRequest object(s) exist for that user, update their status and deleted_user property
+            UserGDPRDeletionRequest.objects.filter(user_id=self.user.id)\
+                .update(status="de", deleted_user=deleted_user_object)
```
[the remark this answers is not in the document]
ffont: now it does with the other changes I implemented about keeping gdpr delete requests changes history

accounts/models.py
ffont: I'm not sure if we can do much, hopefully this should not happen often. maybe add it in the devs documentation, but we should be in the habit of reading it as well

accounts/models.py
ffont: Addressed in c654917. Now we use PG's Greatest to make sure the count fields are not set to < 0 and therefore we avoid the need of updating the fields before deletion.

accounts/management/commands/monitor_async_deleted_users.py
ffont: Addressed in 82194ac4d782e4ad84a29d22fe99ccdb44def351. We now use `DeletionUserRequest` table to check for users that should have been deleted and were not.

accounts/models.py
ffont: this discussion is happening outside this PR. could be something to consider for the delete sounds refactor that should come after this PR

ffont:
As discussed with @alastair, here is a list of detailed test cases for manually testing the new functionality of this PR. You should make sure that the  `worker_async_tasks` service is running when testing. Do: `docker-compose up web worker_async_tasks`. I think you should be able to do all checks in the Django admin, you'll need an account with permissions to access it.

1. Create an account, login with that account and delete the account via user settings page. Check that:
  - [ ] `DeletedUser` object was created and has appropriate deletion reason
  - [ ] `UserDeletionRequest` object was created and the status history show an appropriate change of states
  - [ ] Check that if you access the user profile page with the "old" url (with the username), you get 404
  - [ ] Check that if you access the user profile page with the new username (you'll have to get that from the admin?), you see a the "deleted user XXX" page

2. Delete an existing account using the Django admin with the option "delete spammer". Check that:
  - [ ] `DeletedUser` object was created and has appropriate deletion reason
  - [ ] The original `User` object is deleted from the DB
  - [ ] `UserDeletionRequest` object was created and the status history shows an appropriate change of states

3. Delete an existing accoun that has sounds using the Django admin with the option "delete user, keep sounds". Check that:
  - [ ] `DeletedUser` object was created  and has appropriate deletion reason
  - [ ] Check that if you access the user profile page with the new username (you'll have to get that from the admin?), you see a the "deleted user XXX" page with the list of uploaded sounds
  - [ ] Check that trying to access deleted user's bookmarks page, list of downloads, list of comments, followers and geotags returns 404. Only profile page, list of uploaded sounds and sound pages are available

4. Delete an existing accoun that has sounds using the Django admin with the option "delete user including sounds". Check that:
  - [ ] `DeletedUser` object was created  and has appropriate deletion reason
  - [ ] Check that if you access the user profile page with the new username (you'll have to get that from the admin?), you see a the "deleted user XXX" page but no sounds
  
5. Choose a user that has sounds and at least 1 pack, update its profile using the admin to set the "num_sounds", "num_sound_downloads", "num_pack_downloads" to wrong values (set them all to 1 for example). Also, edit a sound uploaded by that user and set the "num_downloads" field also to to a wrong value of 1. Also, edit a pack uploaded by that user and set the "num_downloads" field also to to a wrong value of 1. Delete the user using the admin with the option "delete user including soudns". Check that:
  - [ ] `DeletedUser` object was created  and has appropriate deletion reason
  - [ ] Check that if you access the user profile page with the new username (you'll have to get that from the admin?), you see a the "deleted user XXX" page but no sounds and no packs

6. Using the Django admin, manually create a `UserDeletionRequest` object and link the deletion request to an existing user account. Check that:
  - [ ] When changing the `UserDeletionRequest` states via the admin, the status history field gets updated accordingly.
  - [ ] If deleting the user via the admnin, the status of the `UserDeletionRequest` object gets automatically updated

7. Stop the `worker_async_tasks` service and delete 1 existing user using the admin. Temporarily set the `CHECK_ASYNC_DELETED_USERS_HOURS_BACK` setting to 0. Run the `check_async_deleted_users` management command. Check that:
  - [ ] The command reports that 1 user should have been deleted but it was not
  - [ ] After starting again `worker_async_tasks` and re-deleting the user, check that re-running the command no longer results in 1 user being reported.

PR #1399, opened 2020-01-06
Add typeahead on frontpage search input

giuband:
Screenshot 2020-01-06 at 18 12 28

Hey team, since this is quite a big change that will also require some JS code from your side to actually fetch the suggestions from your BE let me know how I can help guiding you through it.

### DIFF
--- a/freesound/static/bw-frontend/src/pages/front.js
+++ b/freesound/static/bw-frontend/src/pages/front.js
@@ -25,19 +25,10 @@ const checkShouldShowSearchInNavbar = throttle(() => {
 window.addEventListener('scroll', checkShouldShowSearchInNavbar)
 
 const fetchSuggestions = async query => {
-  const fakeSuggestions = [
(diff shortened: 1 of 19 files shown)


freesound/static/bw-frontend/src/pages/front.js
```diff
@@ -1,18 +1,43 @@
...
   if (shouldShowSearchBar !== isShowingSearchBar) {
-    navbar.classList.toggle('bw-nav--expanded');
+    navbar.classList.toggle('bw-nav--expanded')
   }
-}, SCROLL_CHECK_TIMER);
+}, SCROLL_CHECK_TIMER)
+
+window.addEventListener('scroll', checkShouldShowSearchInNavbar)
+
+const fetchSuggestions = async query => {
```
giuband: This is the function that you'll need to change in order for it to fetch actual suggestions from your BE

PR #1411, opened 2020-02-05
Similarity docker

alastair:
**Description**
Set up similarity server to use docker

**Deployment steps**:

...pending

### DIFF
--- a/similarity/docker-compose.yml
+++ b/similarity/docker-compose.yml
@@ -1,0 +1,13 @@
+version: "3.7"
+
+services:
+    similarity:
+        build: .
+        init: true
+        command: python similarity_server.py
+        volumes:
+            - .:/code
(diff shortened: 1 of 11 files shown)


similarity/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+
+services:
+    similarity:
+        build: .
+        init: true
+        command: python similarity_server.py
+        volumes:
+            - .:/code
+            - ../freesound-data/similarity_index:/gaia_index
+            - ../freesound-data/analysis
```
alastair: this file is invalid, it should be fixed

ffont: it worked in my localhost, but we're no trying to access the directory and that'd fail I guess

PR #1412, opened 2020-02-06
Dockerize tag recommendation service

ffont:
This PR adds Docker configuration files to be able to run the tag recommendation service under Docker environment. It also includes the upgrades of some internal packages used by the system. It follows a very similar structure to the Similarity docker setup.


tagrecommendation/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+	; \
+	pip --version; \
+	\
+	find /usr/local -depth \
+		\( \
+			\( -type d -a \( -name test -o -name tests \) \) \
+			-o \
+			\( -type f -a \( -name '*.pyc' -o -name '*.pyo' \) \) \
+		\) -exec rm -rf '{}' +; \
+	rm -f get-pip.py
```
alastair: If we are duplicating this in all of our dockerfiles, perhaps we should have a base dockerfile which does:
 - sets PYTHONUNBUFFERED
 - installs base packages
 - installs pip
 - creates fsweb user
 - adds entrypoint file

tagrecommendation/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: "3.7"
+
+services:
+    tagrecommendation:
+        build: .
+        init: true
+        command: python tagrecommendation_server.py
+        volumes:
+            - .:/code
+            - ../freesound-data/tag_recommendation_models:/tag_recommendation_models
```
alastair: looking at this again, I wonder if we should standardise on a `/data/` directory or similar in all images to store data. we can then have sub-directories named like this (`/tag_recommendation_models`)

ffont: I was thinking about this. Yes we should, so all images share the same paths to data files as well. Eg: `/freesound-data/tag_recommendation_models/`, `/freesound-data/analysis/ ` ...

ffont: This is also important because, for example, similarity has an endpoint to add a point to the gaia index and it takes as input the path to an analysis file. This path is sent from appserver. therefore appserver and similarity need to have the same path

Issue #1419, opened 2020-03-26
Improve error message when uploading large files


ffont:
When uploading files larger than the limit (which should be 1GB), a "413 Request entity too large" error is returned from the server, but our uploader UI only shows a generic error message like "Something went wrong". This creates confusion. When the response from the server is 413, we should show an error message notifying users the file is too big (and indicating the maximum filesize).

alastair:
I think this error comes from nginx, which means we should edit the error template files in freesound-deploy. In the case that we use the ajax uploader, this error message should be able to be read by the javascript, in which case we can make a nice error message.
Additionally, check if the ajax uploader can see the size of the file _before_ it's uploaded, which would save us an additional step

ffont:
When you click upload I think we tell nginx how big the whole upload it will be, and  nginx responds immediately with 413 if it is too large. So we have to click "upload" but not wait for the upload to fail after X minutes (or hours!). Nevertheless this could be optimized by actively asking before clicking upload (I guess). One thing I also discovered is that when uploading multiple files, the size is added, so the 1GB file size limit is for all files combined.

alastair:
If it's a regular "choose file and click upload" that sends a POST then you're right, we'll immediately get the error message. In this case it will be nginx returning it though, before we get to django, so we'll have to decide how to format this message.
I don't think we will have access to information about this file in javascript, without requesting access to the user's filesystem, which I don't think we can easily do.

In the JS upload, [we can get the file size before uploading](https://w3c.github.io/FileAPI/#attributes-blob). I think the fact that we send all sounds as a single request is an implementation detail. Maybe it won't be too difficult to upload these files one at a time instead, which would improve this workflow.

ffont:
Closing this issue as is was implemented in PR #1421

PR #1421, opened 2020-04-01
Added javascript check to keep upload size below 1 Gb

P-D-E:
**Issue(s)**
#1419  

**Description**
This PR aims to guard against uploads over 1 Gb, by checking the total size of the files in list, discarding files that won't fit and, in that case, alerting the user.
If the limit ever gets changed in nginx, the code here will have to be updated accordingly.


alastair:
awesome, thanks for the suggestion about how to handle this issue! I think we can make a few changes to it, but this is a great start. 

We could have an indicator/progress bar next to the file list that shows the total size of all files (perhaps only if it reaches 700mb or so, say "700/1000MB used" and make it get red the closer to 1gb it gets. Then we can show a message here in red "Cannot add the previous file, too big" instead of having an alert. What do you think?

P-D-E:
I've replaced the alert with a call to resetForm to have the red message, painless :)
Now, onto the indicator; keep feedback and further ideas coming!

P-D-E:
I've added a graphic bar below the file list, visible as usage gets above 50%, with green, orange, and red colors, let me know how it goes.

ffont:
Hi, I made some small changes to improve this PR and have it ready for merging. I'd like to push the changes to the same branch in your fork but I don't seem to be able to do it because of lack of permissions. @P-D-E can you give me permissions to push to your fork?

Here are some screenshots of the updates:

less_than_50_no_warning

green_warning

yellow_warning

limit_exceeded

P-D-E:
@ffont : Really nice! You should be able to push your changes now, let me know.

ffont:
Ok, I was able to push now, although I think that I might have a typo in my branch name and this was the reason why I could not push before, so maybe special permission was not really needed at all. Anyway, here's the update. @alastair if you want to have a look, and @P-D-E if it looks good to you; then I'll merge and deploy :)

P-D-E:
@ffont It looks good indeed, happy merging :)

PR #1423, opened 2020-04-14
Don't use control chars in APIv2 browser-version docs memacahe keys

ffont:
**Issue(s)**
https://github.com/MTG/freesound/issues/1422

**Description**
I suspect a recent upgrade in memacache made the browsable API to fail because it tries to save and load data from cache using invalid cache keys. This is quite urgent to fix because it means browsable API does not work at al right now :)

### DIFF
--- a/apiv2/apiv2_utils.py
+++ b/apiv2/apiv2_utils.py
@@ -46,6 +46,7 @@ from apiv2.exceptions import RequiresHttpsException, UnauthorizedException, Serv
 from examples import examples
 from search.views import search_prepare_query
 from similarity.client import SimilarityException
+from utils.encryption import create_hash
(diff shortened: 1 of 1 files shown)


apiv2/apiv2_utils.py
```diff
@@ -74,7 +75,8 @@ def hash(self):
 def get_view_description(cls, html=False):
     description = ''
     if getattr(cls, 'get_description', None):
-        cache_key = cls.get_view_name()
+        cache_key = create_hash(cls.get_view_name(), add_secret=False, limit=32)
+        print cache_key
```
alastair, line 79: no print needed, but otherwise it's fine

ffont: oups hehe, shame on me

Issue #1323, opened 2019-05-21
Remove EmailBounce objects when email is changed via admin


ffont:
When we try to send emails to a user and these emails bounce, we create a `EmailBounce` object associated to the `User` object. When this happens and the user logs in, she sees a messages asking to reset the email. If email is reset following that process the `EmailBounce` object is deleted accordingly. However if the email address is changed directly in the admin interface (by an admin for example) then the `EmailBounce` is not deleted.

We should check when a user is saved if the `email` field has changed, and if that's the case delete `EmailBounce` object(s) related to the user.

ffont:
Also, we should add EmailBounce in the Django admin and enable search by email address (and user id/username maybe) so that admins can also deal with these objects.

PR #1428, opened 2020-04-29
Clear EmailBoucne objects even if changing email from admin

ffont:
**Issue(s)**
Fixes https://github.com/MTG/freesound/issues/1323

**Description**
This PR changes the way in which `EmailBounce` objects are cleared so that it works both when users change their email through the email reset form and when an admin changes user's email though the admin page. Now `EmailBounce` objects are cleared in a `User` `pre_save` hook.

### DIFF
--- a/accounts/admin.py
+++ b/accounts/admin.py
@@ -31,7 +31,6 @@ from django.contrib.auth.models import User
 from django.core.paginator import Paginator
 from django.utils.functional import cached_property
 from django.db import connection
-from django.core.cache import cache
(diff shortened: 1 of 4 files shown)


alastairporter:
looks good to me.

alastair:
did you also want to add the EmailBounce object to the admin here?

ffont:
> did you also want to add the EmailBounce object to the admin here?

absolutely hehe, just added it. will merge

Issue #1427, opened 2020-04-29
Add more logging


ffont:
To be better able to monitor Freesound status in Graylog (and set up alerts for example), we should add more logging. These are some ideas of things to add:

- A sound object is created: this is when a sound gets described. We can log this and add information about sound being uploaded through API, normal web form, bluk describe.
- An email is sent: log everytime an email is sent, we could add subject as well so we can see the different types of emails. We can include username to but email address not needed (?)
- Log workers activity (or improve logging when it is already there): process/analysis workers, including result and duration; async tasks
- Cron jobs: log start and end of cronjob, including command name, duration?

PR #1429, opened 2020-04-30
Refactor logging

ffont:
**Issue(s)**
Fixes https://github.com/MTG/freesound/issues/1427

**Description**
This PR does a general logging refactor in many places of the Freesound codebase. Changes are not very big (in many occasions simply renaming a logger variable name), but it affects many files. Also I optimized imports of most of the files I edited so PR is now big :) This is a list of things changed:

- Simplify logger facilities. Now use `workers` for workers, `commands` for management commands, and some other simplifications
- Add `LoggingBaseCommand` class to facilitate logging of some management commands, and use it where appropriate
- Refactored logging in workers. Now reported logs also provide duration of tasks.
- Log when we send an email
- Log when we create a sound object
- Use more specific logger names in all Django apps and management commands (e.g. `console_logger` instead of `logger`)
- Use `xxx_logger.error` instead of `xxx_logger.info` when it makes sense

**Deployment steps**:

- Deploy using updated loggers configuration in deploy repository (already on master branch there)
- Change `donation_mails` command name in appserver3 crontab for `send_donation_request_emails`
- Change `create_random_sound` command name in appserver3 crontab for `create_random_sounds`


apiv2/apiv2_utils.py
```diff
@@ -52,7 +52,7 @@
 from utils.similarity_utilities import api_search as similarity_api_search
 from utils.similarity_utilities import get_sounds_descriptors
 
-logger_error = logging.getLogger("api_errors")
+error_logger = logging.getLogger("api_errors")
```
alastair, line 55: in `apiv2/exceptions.py` this is called api_errors_logger

general/management/commands/gm_worker_async_tasks.py
```diff
@@ -91,84 +78,93 @@ def task_whitelist_user(self, gearman_worker, gearman_job):
...
+                workers_logger.info("Whitelisted user (%s)" % json.dumps(
+                    {'user_id': whitelist_user.id,
+                     'username': whitelist_user.username,
+                     'work_time': time.time() - local_start_time}))
 
             count_done = count_done + 1
-            self.write_stdout("Finished processing one ticket, %d remaining" % (len(tickets)-count_done))
+
+        workers_logger.info("Finished whitelisting users from tickets (%s)" % json.dumps(
+            {'n_tickets': len(tickets), 'work_time': time.time() - start_time}))
```
alastair, line 89: this will have lots of decimal places, normally when I do something like this I use `round()`

general/management/commands/report_index_statuses.py
```diff
@@ -41,72 +44,82 @@ def add_arguments(self, parser):
...
-        print "Freesound\t\t%i  (moderated, processed and analyzed)" % len(fs_mpa)
-
-        print "\n\n***************\nSOLR INDEX\n***************\n"
         in_solr_not_in_fs = list(set(solr_ids).intersection(set(set(solr_ids).difference(fs_mp))))
         in_fs_not_in_solr = list(set(fs_mp).intersection(set(set(fs_mp).difference(solr_ids))))
-        print "Sounds in solr but not in fs:\t%i" % len(in_solr_not_in_fs)
-        print "Sounds in fs but not in solr:\t%i" % len(in_fs_not_in_solr)
+
+        message = "\nNumber of sounds per index:\n--------------------------"
+        message += "Solr index\t\t%i" % len(solr_ids)
```
alastair, line 70: missing newlines after these messages. You could add them to a list and then do "\n".join(thelist)

utils/management_commands.py
```diff
@@ -0,0 +1,59 @@
...
+import sys
+import time
+
+from django.core.management.base import BaseCommand
+
+
+commands_logger = logging.getLogger('commands')
+
+
+class LoggingBaseCommand(BaseCommand):
```
alastair, line 33: would be nice to have some documentation to know what this command does, when it should be used, and how to use it

utils/sound_upload.py
```diff
@@ -66,10 +67,31 @@ def _remove_user_uploads_folder_if_empty(user):
     remove_empty_user_directory_from_mirror_locations(user_uploads_dir)
 
 
-def create_sound(user, sound_fields, apiv2_client=None, process=True, remove_exists=False):
+def create_sound(user,
```
alastair, line 70: changed format here for some reason?

sounds/models.py
```diff
@@ -80,6 +80,195 @@ def __unicode__(self):
         return self.name
 
 
+class BulkUploadProgress(models.Model):
```
alastair, line 83: Why was this copied from lower in the file?  (remember that if you need a model for a foreignkey you can use a "String" instead). Don't worry about moving it back, it's done now...

utils/management_commands.py
```diff
@@ -0,0 +1,59 @@
...
+    def log_start(self, data=None):
+        self.start_time = time.time()
+        self.command_name = self.find_command_name()
+        if data is None:
+            data = {}
+        data['command'] = self.command_name
+        commands_logger.info('Started management command ({0})'.format(json.dumps(data)))
+
+    def log_end(self, data=None):
+        command_name = os.path.basename(__file__)
```
alastair, line 53: what's this?

utils/management_commands.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def log_start(self, data=None):
+        self.start_time = time.time()
+        self.command_name = self.find_command_name()
+        if data is None:
+            data = {}
+        data['command'] = self.command_name
+        commands_logger.info('Started management command ({0})'.format(json.dumps(data)))
+
+    def log_end(self, data=None):
```
alastair, line 52: documentation would be good here

general/management/commands/similarity_update.py
```diff
@@ -111,24 +111,12 @@ def handle(self,  *args, **options):
...
-        logger.info("Finished similarity update. %i sounds added to the similarity index" % to_be_added.count())
-        # At the end save the index
-        #if options['indexing_server']:
-        #    Similarity.save_indexing_server()
-        #else:
-        #    Similarity.save()
+                console_logger.error('Unexpected error while trying to add sound (id: %i, %i of %i): \n\t%s'
+                                     % (sound.id, count+1, N, str(e)))
+
+        self.log_start({'n_sounds_added': to_be_added.count()})
```
alastair, line 122: should this be start or end?

utils/management_commands.py
```diff
@@ -0,0 +1,59 @@
...
+    start_time = None
+    command_name = None
+
+    def find_command_name(self):
+        for count, arg in enumerate(sys.argv):
+            if arg == 'manage.py':
+                return sys.argv[count + 1]
+        return None
+
+    def log_start(self, data=None):
```
alastair, line 44: none of the times you call log_start you pass in data, I wonder if it's necessary? Perhaps we could just set start_time and command_name in `__init__`?

utils/management_commands.py
```diff
@@ -0,0 +1,59 @@
...
+            data = {}
+        data['command'] = self.command_name
+        commands_logger.info('Started management command ({0})'.format(json.dumps(data)))
+
+    def log_end(self, data=None):
+        command_name = os.path.basename(__file__)
+        if data is None:
+            data = {}
+        if self.start_time:
+            data['work_time'] = time.time() - self.start_time
```
alastair, line 57: this should be rounded

utils/management_commands.py
```diff
@@ -0,0 +1,59 @@
...
+        data['command'] = self.command_name or self.find_command_name()
```
alastair, line 58: in what situations will self.command_name not be set?

utils/management_commands.py
```diff
@@ -0,0 +1,59 @@
...
+        commands_logger.info('Finished management command ({0})'.format(json.dumps(data)))
```
alastair, line 59: I'm not a huge fan of this format of message, because if there is a lot of data here, it's difficult to read. Is this specifically for graylog? If we're going to use a helper tool like this maybe it's better to have it output something easily readable by people, and send a message directly to graylog using a dedicated logger in a format that it understands.

accounts/management/commands/fill_downloads.py
```diff
@@ -19,10 +19,11 @@
...
 import logging
+
 from django.core.management.base import BaseCommand
 from django.db import connection
 
-logger = logging.getLogger("console")
+console_logger = logging.getLogger("console")
 
 
 class Command(BaseCommand):
```
alastair, line 29: You don't use the Logging command here. Is there a reason why?

apiv2/apiv2_utils.py
ffont: fixed in c9a63184c9ef42bd624493321a4d7d32e991f917

general/management/commands/gm_worker_async_tasks.py
ffont: Nice one! Addressed in c4ef4707a3719bb7ac444dbf2905eb13bea18109

utils/management_commands.py
ffont: c4ef4707a3719bb7ac444dbf2905eb13bea18109

general/management/commands/report_index_statuses.py
ffont: Addressed in 361ab791b6e78c01d165192a8c9ac6f8545f9217

utils/management_commands.py
ffont: Added ebb16b24e48d3e8da429a01a56a0cac3e8ad9e71

utils/management_commands.py
ffont: You're right, but I think in this way it is more generic. Maybe in some cases you do want to pass some data at the start of the command and some data at the end. It's just a couple of lines of code so I think it is fine keeping it. But if you have a strong feeling against it I'm fine removing.

utils/management_commands.py
ffont: Only if `self.log_start` was not called. It should have been called, but just in case. I guess I'm trying to protect against server errors if logging helpers not called properly. Maybe it is better to just let it raise?

utils/management_commands.py
ffont: This is for graylog yes. We use it in many places. To be honest I think this data is not to be read a lot by humans in log files. But if it needs to be read it's not too bad because most of the times its only a property or two that we set here. The idea however is that these things will be inspected in graylog, and there we can format them nicely. Using a different format for printing on screen means more lines of code which I don't think will be really needed. For the commands we typically want to see the output on the console, we also have a console_logger which prints the info (eg the report_index_statuses command)

accounts/management/commands/fill_downloads.py
ffont: Not really. Updated here fcaac4a31ad0d681e020aea7d1255856712cca3e

utils/sound_upload.py
ffont: Because Ia added an extra item and it looked cleaner this way.

ffont:
Thanks for the comments. I addressed some of them. Also answered some of your concerns. If the changes/answers look good to you, I think we're ready to go.

apiv2/views.py
```diff
@@ -1229,7 +1228,7 @@ def get_description(cls):
                % (prepend_base('/docs/api'), 'index.html')
 
     def get(self, request,  *args, **kwargs):
-        logger.info(self.log_message('api_root'))
+        api_logger.info(self.log_message('api_root'))
```
alastair, line 1231: out of interest, the log message format constructed by this is different to the format constructed with json and `()`. Is this something that should be unified eventually?

apiv2/views.py
```diff
@@ -1349,8 +1348,8 @@ def create_apiv2_key(request):
             api_client.accepted_tos = form.cleaned_data['accepted_tos']
             api_client.save()
             form = ApiV2ClientForm()
-            logger.info('%s <> (ApiV2 Auth:%s Dev:%s User:%s Client:%s)' %
-                        ('new_credential', None, request.user.username, None, api_client.client_id))
+            api_logger.info('%s <> (ApiV2 Auth:%s Dev:%s User:%s Client:%s)' %
+                            ('new_credential', None, request.user.username, None, api_client.client_id))
```
alastair, line 1352: small thing, but the first string interpolation probably isn't needed - can just put 'new_credential' directly in there, right?

apiv2/views.py
ffont: Yes, it is different because initially we were only sending "structured" data from the API logs, then I added a "generic" way of doing it. It could be unified at some point.

apiv2/views.py
ffont: absolutely!

apiv2/views.py
alastair: cool, let's open a ticket for it then. It's a good continuation of my comment about improving this in the future.

PR #1432, opened 2020-05-12
Make similarity client a stateful class

alastair:
**Description**
When I was looking at this similarity client for some gaia testing, I noticed that you can't have 2 clients, each pointing to a different host.

This change lets you pass in a host into the constructor. 

There are still some changes to make - we need to read the settings file to get the host where this client is created.
I also changed to using requests, but we should improve the methods to pass GET args as a dict instead of constructing them manually.

### DIFF
--- a/similarity/client/__init__.py
+++ b/similarity/client/__init__.py
@@ -18,12 +18,8 @@
 #     See AUTHORS file.
 #
 
-from django.conf import settings
-import json
-import urllib2
+import requests
 
-_BASE_URL                     = 'http://%s:%i/similarity/' % (settings.SIMILARITY_ADDRESS, settings.SIMILARITY_PORT)
-_BASE_INDEXING_SERVER_URL     = 'http://%s:%i/similarity/' % (settings.SIMILARITY_ADDRESS, settings.SIMILARITY_INDEXING_SERVER_PORT)
 _URL_ADD_POINT                = 'add_point/'
 _URL_DELETE_POINT             = 'delete_point/'
 _URL_GET_DESCRIPTOR_NAMES     = 'get_descriptor_names/'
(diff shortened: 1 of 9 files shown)


alastair:
I've updated this to create an instance of the similarity client for both the normal server and the indexing server. This removes the requirement to have separate methods for add/save.

There are still some things that need to be improved:
 * `requests` 'timeout' by default is infinite. We should set this to a specific limit, and then undo it for save if needed
 * Still passing constructed parameters as the url, need to switch to `params` argument to `requests.get`. I'd like to be able to find a way to keep the tests the same (with a fully constructed url) in order to verify that we make this change correctly. I need to think about how these tests will fit together a bit better in order to do this
 * move client file from similarity/client/init.py to similarity/client.py
 * Fix conflicts

I still need to test with a real running similarity server

similarity/client/__init__.py
```diff
@@ -169,3 +137,7 @@ def get_sounds_descriptors(cls, sound_ids, descriptor_names=None, normalization=
             url += '&only_leaf_descriptors=1'
 
         return _result_or_exception(_get_url_as_json(url))
+
+
+similarity_client = Similarity(settings.SIMILARITY_ADDRESS)
+indexing_similarity_client = Similarity(settings.INDEXING_SIMILARITY_ADDRESS)
```
ffont, line 143: cool, love that

alastair:
yeah, I found the `prepare().url` syntax too. It's just a matter of seeing if I can write a test that contains the query parameter as "?foo=bar&etc", and then make the change to the code and see the tests stay the same. I'd prefer to use request's `.get(baseurl, paramsdict)` syntax, but maybe there's a middleground where I can construct the full url and pass it to `.get()` and then once the tests pass, move to the other syntax. I'll keep looking.

alastair:
Hi @ffont, I can't remember why we never finished this work. Was it really just because I couldn't agree on how to construct the test URLs? If that's it, then I propose that we quickly fix this and merge it. I'll try and resolve the conflicts.

ffont:
Yeah, I guess that was it! Also similar thing should be done for tag recommendation, although I want to refactor it as a celery task and running on py3 so maybe there's no need to make this update now.

alastair:
Closing as our mid-term plans are to use solr for similarity and this will go away anyway

PR #1436, opened 2020-05-15
More small logging improvements

ffont:
**Issue(s)**
https://github.com/MTG/freesound/issues/1427

**Description**
Added a couple of extra improvements to the logging recent changes:

- Fix bug in `LoggingBaseCommand` that would not log command name properly when running from cron
- Add task names to workers (processing, analysis, async tasks) loggers
- Add `unified_subject` to email logger so that we can list email types in Graylog without "variable" subjects being considered as different email types

### DIFF
--- a/accounts/tests/test_spam.py
+++ b/accounts/tests/test_spam.py
@@ -81,7 +81,8 @@ class ReportSpamOffensive(TestCase):
 
         # Now check that after the new flags an email was sent
         self.assertEqual(len(mail.outbox), 1)  # Notification email sent
-        self.assertTrue("[freesound] Spam/offensive report for user" in mail.outbox[0].subject)
+        self.assertTrue(settings.EMAIL_SUBJECT_USER_SPAM_REPORT in mail.outbox[0].subject and
+                        settings.EMAIL_SUBJECT_PREFIX in mail.outbox[0].subject)
         self.assertTrue("has been reported" in mail.outbox[0].body)
 
         # Continue flagging object until it reaches blocked state
@@ -99,7 +100,8 @@ class ReportSpamOffensive(TestCase):
 
         # Now check that an extra mail was now sent (the email notifying user is blocked)
         self.assertEqual(len(mail.outbox), 2)  # New notification email sent
-        self.assertTrue("[freesound] Spam/offensive report for user" in mail.outbox[1].subject)
+        self.assertTrue(settings.EMAIL_SUBJECT_USER_SPAM_REPORT in mail.outbox[1].subject and
+                        settings.EMAIL_SUBJECT_PREFIX in mail.outbox[1].subject)
         self.assertTrue("has been blocked" in mail.outbox[1].body)
 
         # Flag the object again and check that no new notification emails are sent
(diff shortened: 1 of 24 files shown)


accounts/tests/test_spam.py
```diff
@@ -81,7 +81,8 @@ def __test_report_object(self, flag_type, object):
 
         # Now check that after the new flags an email was sent
         self.assertEqual(len(mail.outbox), 1)  # Notification email sent
-        self.assertTrue("[freesound] Spam/offensive report for user" in mail.outbox[0].subject)
+        self.assertTrue(settings.EMAIL_SUBJECT_USER_SPAM_REPORT in mail.outbox[0].subject and
```
alastair, line 84: because these are tests, perhaps it makes better sense to list them as 2 separate asserts, so that if one fails we can easily see which one it is

accounts/views.py
```diff
@@ -237,7 +237,7 @@ def send_activation(user):
         'username': username,
         'hash': uid_hash
     }
-    send_mail_template(u'Your activation link.', 'accounts/email_activation.txt', tvars, user_to=user)
```
alastair, line 240: interesting, I didn't see that some of our subjects had . at the end. I see you've removed them, good.

follow/management/commands/send_stream_emails.py
```diff
@@ -71,8 +71,7 @@ def handle(self, *args, **options):
             week_first_day_str = week_first_day.strftime("%d %b").lstrip("0")
             week_last_day_str = week_last_day.strftime("%d %b").lstrip("0")
 
-            subject_str = u'new sounds from users and tags you are following ('
-            subject_str += unicode(week_first_day_str) + u' - ' + unicode(week_last_day_str) + u')'
+            extra_email_subject = unicode(week_first_day_str) + u' to ' + unicode(week_last_day_str)
```
alastair, line 74: this no longer has (), but I guess it's a bit more difficult to do with the new format of subject line, right?

freesound/settings.py
```diff
@@ -207,7 +224,7 @@
 # -------------------------------------------------------------------------------
 # Freesound miscelaneous settings
 
-SUPPORT = ()
+SUPPORT = (('Support', 'admin@example.org'), )
```
alastair, line 227: I don't like using example.org here, can we change it? Is it needed for the tests? If so, better that we put it in test settings

utils/mail.py
```diff
@@ -102,15 +104,20 @@ def send_mail(subject, email_body, user_to=None, email_to=None, email_from=None,
     if settings.ALLOWED_EMAILS:  # for testing purposes, so we don't accidentally send emails to users
         email_to = [(username, email) for username, email in email_to if email in settings.ALLOWED_EMAILS]
 
+    full_subject = u'{0} {1}'.format(settings.EMAIL_SUBJECT_PREFIX, subject)
+    if extra_subject:
+        full_subject += u' - {0}'.format(extra_subject)
```
alastair, line 109: this feels a bit messy to me, having both `+=` and the format. Maybe it'd be a bit easier to understand to just say `full_subject = "{}- {}".format(full_subject, extra_subject)`

utils/mail.py
```diff
@@ -120,6 +127,7 @@ def send_mail(subject, email_body, user_to=None, email_to=None, email_from=None,
         for username, email in email_to:
             emails_logger.info('Email sent (%s)' % json.dumps({
                 'subject': subject,
+                'full_subject': full_subject,
```
alastair, line 130: do we need both of these? In this case, full_subject contains both `[freesound]` and the extra if it's been set, but previously they `subject` line just contained the subject without the prefix. We could construct the subject without the prefix in order to log here, but then add it in order to send the email. What do you think?

follow/management/commands/send_stream_emails.py
ffont: I also thought I was also nicer without it

freesound/settings.py
ffont: actually this is not for tests (in tests `SUPPORT` is already defined). this is for running locally, because otherwise sending to support fails (which only happens in the contact form). I'll add it to `local_settings.example.py` instead

utils/mail.py
ffont: I'm interested in `subject` really, because it will be constant per email type (eg, all topic reply notifications will have the same `subject`, whereas `full_sibject` includes the topic as well). I think it is nice to log both because we might be interested in the rest. But we could log `extra_subject` instead!

ffont: Done in 09377079f39a904c17a9e6dcfef699a7dfe02f8c

utils/mail.py
ffont: Addressed in 9f6f6a65292e534cb40f0854e28449691a4a5851

accounts/tests/test_spam.py
ffont: good point! Addressed in 9439414688155b780b9743e02530e6f0df673a9c

PR #1439, opened 2020-05-20
Use the database replica to read large tables when generating stats

alastair:
**Description**
Now that we have a database replica, we want to use it for some queries. Stats generation is one of the operations that historically has caused the database server to get slower, so use it for some queries.

I'm not sure about the best way to do the queries in this command. I've chosen a few of the larger tables, and manually selected the replica with the `.using()` method on a queryset.
I didn't do it with all models - what do we think here? Should we only use it for the large tables, or consistently use it on _all_ objects? The replica database will always be completely up-to-date.

Alternatively, we could use a [database router](https://docs.djangoproject.com/en/3.0/topics/db/multi-db/#using-routers) and somehow indicate that we want to use the replica for all read queries. This could be as simple as this command setting an environment variable, and us having a router that checks this environment variable and uses the replica database for all read queries.

Feedback welcome...

**Deployment steps**:
Needs local_settings in production to configure the `ro_replica` database

### DIFF
--- a/environment
+++ b/environment
@@ -1,1 +1,2 @@
 DJANGO_DATABASE_URL=postgres://freesound@db/freesound
+DJANGO_RO_REPLICA_DATABASE_URL=postgres://freesound@db/freesound
--- a/freesound/local_settings.example.py
+++ b/freesound/local_settings.example.py
@@ -5,6 +5,9 @@ import os
 DEBUG = True
 DISPLAY_DEBUG_TOOLBAR = True
 
+# Use a database replica when reading some large tables
+USE_REPLICA_FOR_READS = True
+
 # Data path of the mounted data volume in docker
 DATA_PATH = '/freesound-data/'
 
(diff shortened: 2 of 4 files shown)


alastair:
There's a problem with this setup:
https://stackoverflow.com/questions/14592436/postgresql-error-canceling-statement-due-to-conflict-with-recovery

We need to work out how to prevent these queries from being cancelled. The solution of `hot_standby_feedback = on` or `max_standby_streaming_delay = -1` are probably what we need to set.

alastair:
@ffont what do you think about this? We've really had no slowdown issues since we did our migrations and fixes to the fileserver. In this case, I kind of think that we should forget this PR and stay with the single database

ffont:
Yeah, considering that our main issues are not related to the DB structure but with the filesystem, I think it is good to keep things as they are and not add more complexity for now.

alastair:
Hasn't been a problem since we decided not to implement this

PR #1441, opened 2020-05-25
Use create_directories util method to always create without failing

### LINKED ISSUE #1440: Use consistent method to create directories
We have a number of different places that create directories using either `os.mkdir` or `os.makedirs`. Each use handles exceptions differently - some catch `Exception`, others catch `OSError`, some make a log message, some don't, etc.

We should replace all cases with a single utility method that works the same in all cases. Something like
```py
    try:
        os.makedirs(path)
    except OSError as exception:
        if exception.errno != errno.EEXIST:
            raise
```

alastair:
**Issue(s)**
Fixes #1440 

**Description**
We had many places where we were creating directories with different methods, and reporting different log messages depending on if the operation failed or succeeded.
Use the standard `create_directories` method that we already had, instead of using os.mkdir or os.makedirs.

**Deployment steps**:
Should be nothing specific

### DIFF
--- a/utils/mirror_files.py
+++ b/utils/mirror_files.py
@@ -3,8 +3,7 @@ import os
 import shutil
 import logging
 import errno
-from utils.filesystem import remove_directory_if_empty
-
+from utils.filesystem import remove_directory_if_empty, create_directories
 
 web_logger = logging.getLogger('web')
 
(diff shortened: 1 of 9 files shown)


utils/mirror_files.py
```diff
@@ -13,14 +12,7 @@ def copy_files(source_destination_tuples):
     for source_path, destination_path in source_destination_tuples:
         if settings.LOG_START_AND_END_COPYING_FILES:
             web_logger.error('Started copying file %s to %s' % (source_path, destination_path))
```
alastair, line 14: @ffont I noticed this when making the change - this notification shouldn't be an error, I think it makes more sense to be info or debug.

PR #1447, opened 2020-06-04
Switch to pysndfile for generating images

alastair:
**Description**
We were using scikits.audiolab to load wav files during the image generation stage. This library wraps libsndfile to read wav files.
scikits.audiolab has always given us problems when installing - its setup process required numpy to be installed in a separate step.
Additionally, it's been unsupported for many years, requires an older version of numpy, and doesn't support python 3.
This PR switches to using pysndfile, which is another wraper around libsndfile.

Apart from a few general code cleanups, the main algorithm is unchanged. I ran some tests and found that the new images are byte-equal with the originals (some examples below). I also improved wav2png in case it was still used.

I only tested the workflow with wav2png. I did change the signature of `create_wave_images` to remove a parameter, but I see that the audio_processing script doesn't use this parameter, so there should be no issues here. Nevertheless, it would be a good idea to run a sound through processing to ensure that it works as expected.

**Deployment steps**:
I upgraded numpy, as a newer version was required for pysndfile. There is some basic use of numpy in the image generation script, but it's mostly array management and some constants and window functions. I see that tagrecommendation is build in a separate docker image with an old numpy still, for compatibility with the sparse matrix library. I think that this change won't affect tagrec, but we should double-check.

### DIFF
--- a/docker/Dockerfile.travis_test_runner
+++ b/docker/Dockerfile.travis_test_runner
@@ -5,6 +5,13 @@
 
 FROM freesound:2020-02
 
+# Install specific dependencies needed for processing
+RUN apt-get update \
+	&& apt-get install -y --no-install-recommends \
+		sndfile-programs \
+		libsndfile1-dev \
+	&& rm -rf /var/lib/apt/lists/*
+
 RUN mkdir /code
 RUN mkdir /freesound-data
 WORKDIR /code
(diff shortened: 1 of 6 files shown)


alastair:
Some images

old:
219995__b-lamerichs__37 wav_s

new:
219995__b-lamerichs__37 wav_new_s


old
398087__nucleartape__glitchy-tones wav_w

new
398087__nucleartape__glitchy-tones wav_new_w


old
428508__santiaro99__12-grito-nino wav_s


new
428508__santiaro99__12-grito-nino wav_new_s


They're literally exactly the same

PR #1453, opened 2020-06-17
Register email address

### LINKED ISSUE #1214: Try and prevent email address misspellings on signup
After the changes to simplify the registration page (#1171, #1177) we've noticed that we're getting quite a large number of bounces to domain names that look almost but not quite like gmail.com (and other popular domains).
We wonder if this is due to people not typing in their email address twice for confirmation. Bram also commented that he's seeing more support requests from users who don't receive the confirmation mail where he can't find their account (which also sounds like invalid email address).
We've talked a bit about how we could improve this and came up with some ideas:
 * Revert to the two email address fields
 * Show a popup before submit "You are making an account with name [x] and email [y]. [ok?] [change]"
 * Show a notification _after_ submit "You have made an account with name [x] and email [y]. If its incorrect, contact us at support" (still makes support work for Bram)
 * Check domain name of email against a list of "common" domains (hotmail, gmail, yahoo, etc) and see if it's a small edit distance away. If so, show a notification, but still allow the user to continue (This won't pick up spelling mistakes on the local part of the address). Alternatively, we could make them confirm the email address a second time only if we think it's "wrong"

alastair:
**Issue(s)**
Fixes #1214

**Description**
We want to try and reduce the number of users who try and register with
an invalid email address, so go back to 2 fields to ensure that they
write it correctly


templates/api/minimal_login.html
```diff
@@ -61,7 +61,6 @@
         <input type="submit" value="Login" />
         <input type="hidden" name="next" value="{{ next }}" />
     </form>
-    <p>Do not have a Freesound account? <a href="{% url "apiv2-registration" %}">Register here</a>.</p>
```
ffont, line 64: I think it is still worth pointing to the normal registration page. I'll make that change

alastair: oh yeah, I considered that. thanks

Issue #1451, opened 2020-06-11
Download all attributes without paging


alastair:
Feedback from a user:
The attribution page listing the sounds that you downloaded has a pager on it, even if you request the html or text formats. This makes it difficult to get a list of _all_ of your downloaded sounds for your personal records.
It would be nice to have an option to download some of these formats (text, csv?) with all records at once.

Double-check how slow this will be for a user who has a lot of sounds... going through the download table may take some time.

P-D-E:
I've put a proof of concept here:
https://github.com/P-D-E/freesound/tree/Download-list-of-all-downloaded-sounds
If it's a good starting point, I can issue a pull request.

ffont:
@P-D-E thanks! This looks neat. I'd suggest you to make a pull request directly and include in the description of the PR an explanation of what is implemented and also some hints about how it is done. It can be short (specially for a short PR like this), but this helps review :)

Thanks again!

alastair:
I have some comments, but I think that @ffont is correct that it's probably easier to do a review in a pull request.

Here is a quick overview of my comments, in case you want to address them before you make the PR:
* let's make a new view to download this file instead of having a query parameter on the existing view
* Use  the `csv` module instead of creating lines manually
* Don't save to tempfile - I see that you initially just streamed the response but then later changed it to write to a temporary file. I think that because the view has to do a bunch of work anyway to generate the csv file, it's not much further work to have it serve the file itself. These files are quite small, so I don't think that it's worth the extra complexity that this method brings.

P-D-E:
@ffont @alastair 
Thank you very much for your feedback; yes, I hadn't issued a pull request yet, because I thought it would be better to address the pending issues first and choose the best options, thanks to your insight.
The way it was done was to have it work with a minimal impact on the code (DRY) while providing food for thought;

1. The idea of the query parameter was not to have the query code duplicated, OTOH a dedicated view is surely cleaner, ok (the tests will have to cover the new view accordingly)
2. The idea of not using the `csv` module was to have the same code deal with both text and csv cases, but if you think it's better/safer to use `csv` it can be done; what do we gain from it in this case?
3. The move to a tempfile was meant to play it safe, but if there are no concerns with resource usage and responsiveness, the initial idea of writing directly to the response can be restored.

alastair:
@ffont ready to close?

PR #1459, opened 2020-07-15
Download list of all downloaded sounds

P-D-E:
**Issue(s)**
https://github.com/MTG/freesound/issues/1451

**Description**
This PR provides users a way to download a list of all the sounds they've downloaded, as csv or plain text.
It's accessible via new links in the attribution page, pointing to a new dedicated view, `download_attribution`.
The new view has the same object selection as the `attribution` one it derives from, and serves the whole list without pagination as a downloadable file (direct write into the response with the appropriate headers), in the format specified via the `dl` query parameter. The currently handled values for the parameter are `txt` and `csv`.

### DIFF
--- a/accounts/views.py
+++ b/accounts/views.py
@@ -710,18 +710,17 @@ def describe_sounds(request):
 
 @login_required
 def attribution(request):
-
     qs_sounds = Download.objects.annotate(download_type=Value("sound", CharField()))\
         .values('download_type', 'sound_id', 'sound__user__username', 'sound__original_filename',
                 'license__name', 'sound__license__name', 'created').filter(user=request.user)
     qs_packs = PackDownload.objects.annotate(download_type=Value("pack", CharField()))\
         .values('download_type', 'pack_id', 'pack__user__username', 'pack__name', 'pack__name',
                 'pack__name', 'created').filter(user=request.user)
-    # NOTE: in the query above we duplicate 'pack__name' so that qs_packs has same num columns than qs_sounds. This is
+    # NOTE: in the query above we duplciate 'pack__name' so that qs_packs has same num columns than qs_sounds. This is
     # a requirement for doing QuerySet.union below. Also as a result of using QuerySet.union, the names of the columns
     # (keys in each dictionary element) are unified and taken from the main query set. This means that after the union,
     # queryset entries corresponding to PackDownload will have corresponding field names from entries corresponding to
-    # Download. Therefore to access the pack_id (which is the second value in the list), you'll need to do
+    # Download. Therefre to access the pack_id (which is the second value in the list), you'll need to do
     # item['sound_id'] instead of item ['pack_id']. See the template of this view for an example of this.
     qs = qs_sounds.union(qs_packs).order_by('-created')
 
@@ -732,7 +731,6 @@ def attribution(request):
 
 @login_required
 def download_attribution(request):
-
     content = {"csv": {"style": "csv", "pattern": "%s,%s,%s,%s\n"},
                "txt": {"style": "plain", "pattern": "%s: %s by %s | License: %s\n"}}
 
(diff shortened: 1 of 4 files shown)


accounts/views.py
```diff
@@ -729,6 +729,39 @@ def attribution(request):
...
+    qs_sounds = Download.objects.annotate(download_type=Value("sound", CharField()))\
+        .values('download_type', 'sound_id', 'sound__user__username', 'sound__original_filename',
+                'license__name', 'sound__license__name', 'created').filter(user=request.user)
+    qs_packs = PackDownload.objects.annotate(download_type=Value("pack", CharField()))\
+        .values('download_type', 'pack_id', 'pack__user__username', 'pack__name', 'pack__name',
+                'pack__name', 'created').filter(user=request.user)
+    # NOTE: see the above view, attribution.
+    qs = qs_sounds.union(qs_packs).order_by('-created')
+
+    download = request.GET.get("dl", "")
```
alastair, line 746: let's be consistent in the quote styles in this method

accounts/views.py
```diff
@@ -729,6 +729,39 @@ def attribution(request):
...
+    if download in ["csv", "txt"]:
+        response = HttpResponse(content_type='text/%s' % content[download]["style"])
+        response['Content-Disposition'] = 'attachment; filename="attribution.%s"' % download
```
alastair, line 749: it would be good to include the username and current date/time in this filename. We can make it in a separate variable to make this line simpler

accounts/views.py
```diff
@@ -729,6 +729,39 @@ def attribution(request):
     return render(request, 'accounts/attribution.html', tvars)
 
 
+@login_required
+def download_attribution(request):
+    content = {"csv": {"style": "csv", "pattern": "%s,%s,%s,%s\n"},
```
alastair, line 734: I just had a check, we have 14000 sounds in freesound where the name has a comma in it. This means that it's absolutely necessary to use the csv library to write the csv version.
You could use a StringIO object to write the result to, and then get the value of that to send to the HttpResponse

accounts/views.py
```diff
@@ -729,6 +729,39 @@ def attribution(request):
     return render(request, 'accounts/attribution.html', tvars)
 
 
+@login_required
+def download_attribution(request):
```
alastair, line 733: could you also try and write a test for this method? Something that has at least one sound and at least one pack, and then check that both formats give the expected result

accounts/views.py
P-D-E: Right, sorry! It was a copy-paste heritage from the `attribution` view; the next version of `download_attribution` is consistent, I'll push after writing the test.

accounts/views.py
P-D-E: Done and done, and ready for push; and now onto the test...

accounts/views.py
P-D-E: Should I integrate it into `test_views` or write a separate test file?

P-D-E: Thank you, I found that already, and saw that setting `num_packs=1` in the test setup shouldn't disturb the other tests; I'll go that way.

P-D-E: Push done with the above issues addressed and a first draft of the tests added.

1. Should I compare the content with the actual object properties instead of the fixed strings, in case the `create_user_and_sounds` ever gets changed?
2. Should I fix the pack name appearing instead of the pack license (and if so, how), or can it be ignored for the tests?
3. Should I add a test for the redirection to the `attribution` view in case of unexpected values for the `dl` parameter?

P-D-E: 1. answering myself, yes, fixed strings won't cut it. Besides, while the new tests were working via separate test calls, I broke the whole suit; back to the drawing board...

P-D-E: Ok, the tests are fixed now.

ffont:
Thanks for the PR, looks good now. I'll merge now and will deploy soon :)

PR #1467, opened 2020-08-28
Fix error importing create directories in similarity sever

ffont:
**Description**
A recent (untested) change in the similarity server makes it import `create_directories` from our main utils folder (https://github.com/MTG/freesound/commit/af4878738bf9139d0a3385e315019524e2db8753). However, the similarity server can't import that function as is because it is in a module in the parent folder of the similarity server code. This parent folder is not even passed to the Docker context when building the similarity image.

In this PR I propose a solution to this problem which adds more context to the similarity server image build and makes use of `sys.path.append` to be able to import the module in the relative path (remember this needs to be Python 2). Maybe other solutions would be better like specifically copying the `utils` module in `/code/similarity/utils` so that it can be imported without problems (and without the need of using `sys.path.append`) or maybe other more appropriate stuff. Please enlight me :)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -72,12 +72,12 @@ services:
     
     similarity:
         build:
-            context: ./similarity
-            dockerfile: ../docker/Dockerfile.similarity
+            context: ./
+            dockerfile: ./docker/Dockerfile.similarity
         init: true
         command: python similarity_server.py
         volumes:
-            - ./similarity:/code
+            - .:/code
             - ./freesound-data/similarity_index:/freesound-data/similarity_index
             - ./freesound-data/analysis:/freesound-data/analysis
         ports:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -72,8 +72,8 @@ services:
     
     similarity:
         build:
-            context: ./
-            dockerfile: ./docker/Dockerfile.similarity
+            context: ./similarity
+            dockerfile: ../docker/Dockerfile.similarity
         init: true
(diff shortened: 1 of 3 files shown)


alastair:
How are we going to deploy this in production? If we just have the similarity code in the similarity image, then this solution of modifying sys.path isn't going to work, because the code won't be there.
In this case I'd be happy with just duplicating this code in similarity.

ffont:
Well, I modified the similarity Dockerfile so that we also have the rest of the code. If you think duplicating the function is fine, then I would discard this PR and go this other way. It also makes sense to be as similarity server is indeed a different codebase.

alastair:
Ah, I missed that you changed the context in the docker-compose file. That's OK then

As the `similarity` folder is a module (`__init__.py`), then you could do `python -m similarity.similarity_server` from `/code`, and the main freesound folder will be in the path, so we wouldn't need to modify sys.path. However, we would need to make sure that all of the imports in similarity are correct, and not trying to do relative imports.

alastair:
@ffont any feedback on this proposal to change the way that we run the similarity server?

ffont:
Hi @alastair, I was looking again at this. I have mixed feelings. I think I don't like any solution. Maybe it would be better to simply duplicate `create_directories` code as, after all, similarity is a different app with different codebase. We could add a note to both implementations of `create_directories` to say that if changes need to be done, it should be wise to revise the other implementation as well. If we see this problem happening again in the future, we might want to make some sort of `freesound-utils` package that we can reuse. What do you think?

PR #1492, opened 2020-10-19
Fix ascii decoding error in paginator

ffont:
**Issue(s)**
#1491

**Description**
Paginator objects are passed a request object which includes a list of request GET parameters and values. The paginator uses this object to get parameters like the current page that is requested, and also to construct pagination links which contain all the same GET parameters as the initial request so that whatever things are determined there, will be preserved when moving to the next page. To do that the paginator iterates over all GET parameters and values. This PR adds proper handling of non-ascii characters in the GET parameters passed to the paginator. Even though the original issue was caused by some attacker trying things in the browse tags page, this PR should fix the same error happening in any page using the paginator.

Also in this PR I included a small refactor of tags view, see commit message in https://github.com/MTG/freesound/pull/1492/commits/fb6b85833496987b6f8fb68a379eb3388bfe2ae5

### DIFF
--- a/general/templatetags/paginator.py
+++ b/general/templatetags/paginator.py
@@ -49,7 +49,8 @@ def show_paginator(context, paginator, page, current_page, request, anchor="", n
     # although paginator objects are 0-based, we use 1-based paging
     page_numbers = [n for n in range(min_page_num, max_page_num) if n > 0 and n <= paginator.num_pages]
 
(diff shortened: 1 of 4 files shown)


alastair (commented):
I think this fix is fine for now, but I've mentioned before that I'm not a fan of this paginator. I would much prefer that we explicitly list the parameters that are available on each page, so that we don't propagate query parameters that people add when trying to find exploits.

That's probably too big of a job for this PR, but we should open a ticket to make sure that we have a record of it

Issue #1495, opened 2020-11-03
Front page


ffont:
The former developer started to implement that page, but it is not finished yet. Here is a maybe non exhaustive list of the things remaining.

 * Sounds sometimes break their layout and span 2 columns or other combinations. See example screenshot:

PR #1500, opened 2020-11-06
feature/front-page

GenisDiaz:
**Issue(s)**
[front-page](https://github.com/MTG/freesound/issues/1495)

**Description**
Sounds sometimes break their layout and span 2 columns or other combinations


ffont (approved):
Thanks, I just ported the change to the Django template and will merge and deploy. No need to add both me and Alastair as reviewers. I'll normally do the review and as Alastair for his input if needed. Thanks again!

Issue #1494, opened 2020-11-03
Search page


ffont:
The former developer started to implement the search page, but it is not finished yet. Here is a *maybe* non exhaustive list of the things remaining.

- [x] Pages are not prepared for mobile, should we set a minimum width for now so the layout is more under control?
- [x] Similar sounds icon should have hand icon
- [x] Bookmark sound icon should have hand icon
- [x] Empty checkboxes do not appear
- [x] Checkbox labels too separated from checkbox
- [x] "Sort by" options missing
- [x] Sound display textual description color should be darker
- [x] Star icons too much separated (stars should be closer)
- [x] Tag border a bit too dark?
- [x] Missing paginator [example here](https://projects.invisionapp.com/d/main#/console/11781900/257443463/preview)
- [x] Too much spacing between user icon, username and date? (check spacing generally as I see more spacing than in the designs)
- [x] Waveform looks smaller compared to the design?
- [x] In filters section you'll need to add the "related tags" type of filter (note the fading that the designer added) [example here](https://projects.invisionapp.com/d/main#/console/11781900/257443463/preview)
- [x] Missing loop button in sound player
- [x] Play button in sound player has some gray background that should be transparent
- [x] Start browsing text should have more weight
- [x] The cross icon (to remove query) should only appear when something has been typed and should be gray. Also it should appear next to the text. Also hand icon should be shown on hovering and clicking should clear input contents (btw: is this kind of behaviour supposed to be implemented in the template?) [example here](https://projects.invisionapp.com/d/main#/console/11781900/255631838/preview)
- [x] The cursor for typing search terms should appear in the left of the placeholder instead of the middle. See [example here](https://projects.invisionapp.com/d/main#/console/11781900/257441522/preview).
- [x] Hovering tags should change their background to black (invert colors) instead of have the shadow. See [example here](https://projects.invisionapp.com/d/main#/console/11781900/255631838/preview).
[...]

PR #1501, opened 2020-11-10
[BW] Search page 

GenisDiaz:
**Issue(s)**
https://github.com/MTG/freesound/issues/1494

**Description**

- Similar sounds icon should have hand icon
- Bookmark sound icon should have hand icon
- Empty checkboxes do not appear
- Checkbox labels too separated from checkbox
- "Sort by" options missing
- Sound display textual description color should be darker
- Star icons too much separated (stars should be closer)
- Tag border a bit too dark?
- Missing paginator
- Too much spacing between user icon, username and date? (check spacing generally as I see more spacing than in the designs)
- In filters section you'll need to add the "related tags" type of filter
- Missing loop button in sound player
- Play button in sound player has some gray background that should be transparent
- Start browsing text should have more weight
- The cross icon (to remove query) should only appear when something has been typed and should be gray.
- The cursor for typing search terms should appear in the left of the placeholder instead of the middle. See example here.
- Hovering tags should change their background to black (invert colors) instead of have the shadow. See example here.


ffont:
Hi, thanks for the PR! I reviewed the changes and I have some small comments comparing the current version and the version of Invision:

- [ ] Sort by dropdown should be styled. Example here: https://projects.invisionapp.com/d/main#/console/11781900/255631838/preview (not very clear, but I think you get the idea)
- [ ] I think sound rating stars should be even closer
- [ ] Next to stars, the text "Overall rating (XX)" is missing
- [ ] Some missing spacing after paginator
- [ ] I think there's still a bit too much spacing between username and date
- [ ] Sidebar filter title fontsize should be a bit bigger
- [ ] "Start browsing here" should be centered
- [ ] The soft gray color (e.g., checkbox borders) I think it is too soft. I don't know how many levels of gray are defined and used, but maybe the softer ones should be darkened a bit

I also found a couple of extra things missing in the search page that I'll add to the original issue as they are not that related with the changes of this PR

ffont:
All good, however the "start browsing here" still does not look completely centred to me. I think it is because there is some white space in the right of "start browsing here":

Screenshot 2020-11-11 at 12 04 55

ffont:
Thanks for the update. The duration filter I see it in two columns in Safari and in firefox (macOS):

Screenshot 2020-11-12 at 17 11 38

On chrome it is fine. Other that that changes are good.

ffont:
Nice!
Some extra comments:

 * On mobile, the "cross" icon in the right of "start here" is too much to the right (like there's no margin)
 * Also on mobile mode, I think the sounds should be first and then the filters. This would still not be optimal, but we are not optimising for mobile (priority is not small screens), so I'm happy if the flow is not the best. Having sounds first is important because that is what people will mostly want to see.
 * After seeing it working, I think the "240 sounds ***** sounds **** sort by dropdown" bar should not stay fixed and should scroll with the rest as it takes space and it is not useful to have it there all the time.

ffont:
I finished porting the templates to Django and this is ready to merge now. There are a number of pending things for the search page, some affecting backend stuff, but we'll look into that once more things get ported.

PR #1504, opened 2020-11-16
Revised Moderation Responses 

Headphaze:
Please let me know if my hyperlinks and line breaks would work the way i've done it. 

Cheers

### DIFF
--- a/tickets/__init__.py
+++ b/tickets/__init__.py
@@ -76,7 +76,7 @@ MODERATION_TEXTS = [
     ('Verify Details', "Hey there. Thanks for contributing to Freesound. "
                        " "
                        "Before we can moderate your upload, could you possibly update the description/tags? "
-                       "Any details such as how it was created, recording device, software used, date/location etc-
+                       "Any details such as how it was created, recording device, software used, date/location etc- "
                        "-are extremely useful.
                        " "
                        "If you can't find how to edit your sound, here's a little visual guide: "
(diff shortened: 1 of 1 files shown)


ffont:
Thanks for the PR @Headphaze. There are some syntax errors that I'll fix, but I'm not sure I understand the `hyperlink_format ` bit. These text templates I'm afraid can't contain HTML, so the links should be raw. Does that make sense? Also the template buttons add `*` before the text, but maybe now with these more complete templates this is not needed right? I could try to make these templates HTML-enabled also.  What is your preference?

Headphaze:
Damn, it was worth a try. I'm not great with Python so I looked up a way to add hyperlinks into the code. My only issue was that the FAQ links were long. If you could somehow make the templates HTML enabled that would be great.

Oh yes. I forgot to mention the asterisk symbol. Please remove it as it's not necessary. 

Thanks

ffont:
Ok, I'll work on that. But what do you mean the FAQ links were wrong? there were no FAQ links before right?

Headphaze:
There were no FAQ links before. I'm adding them in to direct user's for guidance. What I meant was that the raw URLs **are** long and ugly. :-P

I'll not make any commits to my master branch then, I'll leave it to you.

ffont:
perfect, I'll take it from here. Thanks Sam!

ffont:
closing this PR as the changes were added in d087e9aaacad7cd4bc1c173d0ff007ab4ba578a0

Issue #1499, opened 2020-11-03
Login modals


ffont:
BW frontend uses modals for login: https://projects.invisionapp.com/d/main#/console/11781900/248591521/preview

Current Freesound design has a dedicated page for login instead of using modals. However in the next version we want to be able to login using modals (as the designer suggested). This will require some changes in the backend to work properly. But I guess for now we just need the modal design in some way which is reusable for other potential modals.

PR #1510, opened 2020-11-18
[BW] Login

GenisDiaz:
**Issue(s)**
[Login modals](https://github.com/MTG/freesound/issues/1499)

**Description**
1. Modals created: 
- Login
- Registration
- New Password
- Logging problems
- Feedback registration
2. Logics between modals.
3. Links in navbar.
4. Query params `newPassword` and `feedbackRegistration` to open NewPasswordModal and FeedbackRegistrationModal


ffont:
Thanks @GenisDiaz!
That looks perfect, except for the notifications in the bottom which might be missing or require some other query param. Like here: https://projects.invisionapp.com/d/main#/console/11781900/251119332/preview The idea is that when you click on "send me the link", a message is shown below "Check your email bla bla". I guess we should be able to trigger that message with some javascript function.

I should start thinking how to integrate modal-based login into Freesound backend. @alastair did you ever do anything similar? I guess something like:

- When users click on "login" we make an ajax GET call to the `login_modal` URL which will return the HTML for the modal
- Users fill in the credentials and then we make an ajax POST call to the `login_modal` URL which should return the updated HTML for the modal if login failed (with form errors), or something else to indicate that logging succeeded.
- If login failed, we replace the modal html with the returned one (to show errors), if login succeeds we reload the current page (so that it is updated with the user logged in)

The issue is with the way `@login_required` works in Django, because it will redirect to a login URL (with the `next` parameter to know where the user was when he was redirected). We should therefore have a dedicated login page. Maybe we can have an empty page with the login modal only?

GenisDiaz:
Hello! @ffont @alastair 

I've fixed conflicts and I've added the toast component to show the quick feedback. 

If you want to see the toast you only have to click on the 'Send me that Link' button.

If you have any doubt about how to use it, let me know.

PR #1515, opened 2020-11-23
[BW] Sound page

GenisDiaz:
**Issue(s)**
[Sound page](https://github.com/MTG/freesound/issues/1496)

**Description**
1. Sound page
2. Rating component
3. Collapsable component
4. Fontsizes classes


ffont:
Thanks a lot @GenisDiaz, this looks super cool 👍 
Some comments:

- [ ] navbar should have the "search sounds" widget at the top left (next to the logo) - this is already implemented in the front page
- [ ] missing "your rating" widget? [example here](https://projects.invisionapp.com/d/main#/console/11781900/252809910/preview)
- [ ] missing hover styles for black and red buttons: white button becomes inverted on hover, but the other buttons don't have hover styles. I guess sll buttons should go "inverted" on hover like the white one?
- [ ] sound description spacing, size and weight does not seem right
Screenshot 2020-11-25 at 11 03 26
- [ ] Also the tags are formatted slightly differently. This is the same reported in other pages. I guess it was fixed in the search sounds branch but not in others.
- [ ] The show more/show less text is inverted and I guess should start collapsed?
- [ ] The "write your comment" input should have a small border but I don't see any in the implemented version. Maybe this was happening also in other pages with similar input. I guess the design should be consisted regarding that. I like the small border.
Screenshot 2020-11-25 at 11 09 07
- [ ] The "comment" button we added I think it would be better placed next to the "write your comment here" input box (so the input box would be a bit smaller). I'd call it "post comment".
- [ ] In similar sounds section, the sound description is too separated vetically from user name and I think it is smaller that qhat it should be. 
Screenshot 2020-11-25 at 11 10 03

GenisDiaz:
> * missing hover styles for black and red buttons: white button becomes inverted on hover, but the other buttons don't have hover styles. I guess sll buttons should go "inverted" on hover like the white one?

Those buttons have the hover style, they become lighter, but maybe it's too small that effect. I think I used the same style the project has for other buttons, Some of them has inverted hover and other ones don't have it (as this case)

GenisDiaz:
@ffont @alastair I've pushed a new commit with the fixes (you have some comments above) if no comment is because I did just the change. Let me know anything about my changes and if I have to do anything else here.

PR #1558, opened 2021-06-04
Add basic rate limitng to search and simialrity pages

ffont:
This PR adds basic rate limiting to the search and sound similarity pages. The limiting is based on IP and can be configured separately for the search and similarity pages. Check `RATELIMITS` in Django settings.

This PR also adds basic IP blocking by setting a list of IPs in `BLOCKED_IPS` in the Django settings. This does not allow to specify ranges or the like, this is possibly something to improve.


freesound/settings.py
```diff
@@ -530,6 +530,17 @@
...
+# -------------------------------------------------------------------------------
+# Rate limiting
+
+RATELIMIT_SEARCH_GROUP = 'search'
+RATELIMIT_SIMILARITY_GROUP = 'similarity'
+RATELIMITS = {
+    RATELIMIT_SEARCH_GROUP: '2/s',
+    RATELIMIT_SIMILARITY_GROUP: '2/s'
+}
+BLOCKED_IPS = []
```
alastair, line 542: this will require a deploy each time we need to change an ip address.
As we're not in an urgent rush right at the moment, I'd like to adapt it to allow it to read these values from a cache.

We should probably not make it read every single request, as that would make 1 more redis/memcache read on every single request. Instead we could save a global variable along with a "last updated" flag, and if the last updated is greater than some time (5 mins?) then we do a cache call to update the list.

templates/403.html
```diff
@@ -0,0 +1,9 @@
+{% extends "base.html" %}
```
alastair, line 1: can we make the rate limiting return http 429 instead of 403? (https://django-ratelimit.readthedocs.io/en/latest/cookbook/429.html)

utils/ratelimit.py
```diff
@@ -0,0 +1,42 @@
...
+    return ip
+
+def key_for_ratelimiting(group, request):
+    return '{}-{}'.format(group, get_ip_or_random_ip(request))
+
+def rate_per_ip(group, request):
+    ip = get_ip_or_random_ip(request)
+    if ip in settings.BLOCKED_IPS:
+        return '0/s'
+    return settings.RATELIMITS[group]
```
alastair, line 42:
```suggestion
    return settings.RATELIMITS.get(group, settings.DEFAULT_RATELIMIT)
```

ffont:
All comments have been addressed, please re-review. Specially for the double-cache thing of the blacklisted ips :)

utils/ratelimit.py
```diff
@@ -0,0 +1,112 @@
...
+    if now - last_cached_blocked_ips_timestamp > settings.CACHED_BLOCKED_IPS_TIME:
+        last_cached_blocked_ips = cache.get(settings.CACHED_BLOCKED_IPS_KEY, None)
+        last_cached_blocked_ips_timestamp = now
+    if last_cached_blocked_ips is not None:
+        return list(set(settings.BLOCKED_IPS + last_cached_blocked_ips))
+    else:
+        return settings.BLOCKED_IPS
+
+
+def add_new_ip_to_block(ip):
```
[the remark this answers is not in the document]
alastair: what happens with a parameter which isn't a valid ip address? can we use the `ipaddress` module to verify it?

utils/ratelimit.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    Add a new IP to the cached list of IPs to block so that further requests to that IP will get blocked.
+    Note that it might take up to settings.CACHED_BLOCKED_IPS_TIME before that IP actually starts to
+    get blocked.
+
+    This method is intended to be used from the console as a quick way og blocking an IP. However, note that
+    if the cache gets cleared, that IP will no longer be clocked. For long-time persistent IP blocking,
```
alastair, line 63: blocked?

utils/ratelimit.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    cached_ips_to_block = cache.get(settings.CACHED_BLOCKED_IPS_KEY, None)
+    if cached_ips_to_block is not None:
+        cached_ips_to_block += [ip]
+    else:
+        cached_ips_to_block = [ip]
+    cache.set(settings.CACHED_BLOCKED_IPS_KEY, cached_ips_to_block)
+
+
+def ip_should_be_blocked(ip):
```
alastair, line 77: ip_is_blocked

alastair (commented):
looks pretty good!

Issue #1526, opened 2021-01-19
Refactor sound analysis pipeline


ffont:
When sounds are uploaded to Freesound, we *process* them to generate previews, display images and extract some metadata; and we also *analyze* them to extract audio featues using the [Freesound Essentia Extractor](https://essentia.upf.edu/freesound_extractor.html). This issue only deals with the **analysis** part.

It would be good to refactor the analysis pipeline so that, besides the analysis carried out with the Freesound Essentia Extractor, we can also run other analyzers like the [Audio Commons Extractor](https://github.com/AudioCommons/ac-audio-extractor), extact embeddings from pre-trained models, etc.

The features for the new analysis pipeline should include:
 
 * The analysis pipeline should be configurable (maybe via a JSON file) and a number of "analysis modules" defined.
 * Each analysis module is a process that analyzes a file and generates some data (e.g. the Freesound Essentia Extractor).
 * The analysis modules are containerized and have a standard input/output API. Also, the analysis modules are versioned.
 * When a sound in analyzed, the results should be stored in the Freesound DB (or in a file if these are huge like Essentia frames), together with an indicator of the analysis module version that was used to generate them. We should also store the analysis log (stdout of the process) and some sort of state to indicate whether the analysis succeeded, or failed, etc.*
 * There should be an "analysis orchestrator" process that will check if there are sounds that have not been analyzed by some of the analysis modules (or that are analyzed with older versions) and trigger analysis processes.

I believe instead of updating our current analysis code, we might want to create a new analysis task in Gearman which implements the new analysis pipline system, plus a new Django management command that implements the analysis orchestrator.

*To store that, we already have a model `sounds.SoundAnalysis` which fits this purpose quite well. A couple of fields should be added to it but it could do the job.

ffont:
Closing this issue as it has been addressed in this PR https://github.com/MTG/freesound/pull/1572 and in some other commits, and has already been merged.

PR #1567, opened 2021-07-22
Feature/analysis refactoring

gonznm:
**Issue(s)**
https://github.com/MTG/freesound/issues/1526

**Description**
First steps towards the analysis orchestrator, which checks if there are sounds that have not been analyzed by the analyzers defined in an Analysis Configuration json file. Some questions:

- Should the main file `freesound/sounds/management/commands/analysis_orchestrator.py` be named differently? "orchestrate_analysis"?
- I think that `freesound/sounds/management/commands/list_sound_analysis.py` command could be deleted now, as `analysis_orchestrator` with the `--status` flag covers that functions plus others... what do you think?
- As you can see, I have created a dictionary at the top of `analysis_orchestrator.py` to be able to 'decode' simple analyzer names and versions (e.g. "fs-essentia:1") into the actual docker images (e.g. "fs-essentia-extractor:20210525_9a3bd10"). On the other side, the Analysis Configuration json file that I have thought of would look like this: 

>     "analyzer:version": ["fs-essentia:1","audio-commons:1", "audioset-vggish:1"]

- I'm not sure if this is the best way to go. In this case both the version and the simple analyzer name should be sent to `analyze_v2` function and then to the analysis worker (this is not implemented, because maybe it doesn't make sense to you). Maybe this dictionary should be in `freesound/sounds/management/commands/gm_worker_analysis.py`, in a json file, or even in the same json file as the analyzers list.


sounds/management/commands/analysis_orchestrator.py
```diff
@@ -0,0 +1,96 @@
...
+import logging
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from sounds.models import Sound, SoundAnalysis
+
+console_logger = logging.getLogger("console")
+
+# Dictionary that links simple names and versions of the analyzers to their actual docker images
```
ffont, line 31: I guess this is what should be defined in the configuration file? Why do we want this mapping? we could probably simply use the name of the docker images

sounds/management/commands/analysis_orchestrator.py
```diff
@@ -0,0 +1,96 @@
...
+            n_sounds = len(Sound.objects.all())
+            n_analyzers = len(config_file['analyzer:version'])
+            total2analyze = n_sounds * n_analyzers
+            console_logger.info("{0} analysis performed. In total, there can be {1} ({2} sounds x {3} analyzers in config file).".format(n_analyzed, total2analyze, n_sounds, n_analyzers))
+            
+            # Count number of analysis to be performed
+            n_analysis = 0
+            for a in config_file['analyzer:version']:
+                # Check all sounds available
+                for s in Sound.objects.all():
```
ffont, line 73: We will iterate all sounds for every analyzer!
Why not simply using SoundAnalysis.objects.filter(...) for every analyzer to know how many sounds have already been analyzed (also those that failed or skipped) and the ones remaining?

I think what the status command should basically output is a table like:

```
analyzer name        | # ok   | # failed   | # skipped    | # missing
essentia:134678x     | 2459   | 123        | 45           | 364736   
...
```

Running the orchestrator will create jobs for the "missing" sounds for each analyzer. Maybe also add some flags to re-trigger "skipped" and/or "failed" and/or  "ok" sounds. Maybe the sounds to be analyzed could be determined with a flag `--include-skipped` , `--include-failed`, `--include-ok`, `--include-missing` for each column. The usual thing will be to call the orchestrator with `--include-missing` only, but if we want to re-process failed sounds, maybe we use `--include-failed` or a combination of other flags.

sounds/management/commands/analysis_orchestrator.py
```diff
@@ -0,0 +1,96 @@
...
+                    _, version = a.split(":")
+                    if not SoundAnalysis.objects.filter(sound=s, analyzer=ANALYZERS_DICT[a], analyzer_version=version).exists():
+                        n_analysis += 1
+            console_logger.info("{0} analysis to be performed.".format(n_analysis))
+
+            # List all saved analysis
+            console_logger.info("List of analyzed sounds:")
```
ffont, line 80: Don't do that, we could have millions of rows :)

sounds/management/commands/analysis_orchestrator.py
```diff
@@ -0,0 +1,96 @@
...
+            for a in SoundAnalysis.objects.all():
+                console_logger.info(json.dumps({'sound_id': a.sound.id, 'SoundAnalysis_id': a.id,
+                                                'analyzer': a.analyzer, 'analyzer_version': a.analyzer_version
+                                                }))
+            
+        else:
```
ffont, line 86: Maybe instead of "status", the special flag could be called `--dry-run` and the difference is that when `--dry-run` is set no jobs are actually scheduled, but the first part of the command (checking current status to know which sound should be analysed) should be the same. Therefore I'd not do a if/else here, just wrap the part in which jobs are created in `if not options['dry-run']:`

sounds/management/commands/analysis_orchestrator.py
```diff
@@ -0,0 +1,96 @@
...
+            console_logger.info("Analysis configuration file: {0}".format(config_file))
+            for a in config_file['analyzer:version']:
+                # Check all sounds available
+                for s in Sound.objects.all():
+                    # if the combination sound-analyzer-version does not exist, trigger analysis
+                    _, version = a.split(":")
+                    if not SoundAnalysis.objects.filter(sound=s, analyzer=ANALYZERS_DICT[a], analyzer_version=version).exists():
+                        console_logger.info(
```
ffont, line 94: Don't print one line per sound, it can potentially be thousands or even millions. I'd just print one line per analyzer and say "X sounds were send to analyze". It will take time to create all the jobs, so you probably want to show some sort of progress and show a percentage as new jobs are being created.

sounds/management/commands/analysis_orchestrator.py
```diff
@@ -0,0 +1,96 @@
...
+            for a in SoundAnalysis.objects.all():
+                console_logger.info(json.dumps({'sound_id': a.sound.id, 'SoundAnalysis_id': a.id,
+                                                'analyzer': a.analyzer, 'analyzer_version': a.analyzer_version
+                                                }))
+            
+        else:
+            console_logger.info("Analysis configuration file: {0}".format(config_file))
+            for a in config_file['analyzer:version']:
+                # Check all sounds available
+                for s in Sound.objects.all():
```
ffont, line 90: This will not be simply "all sounds", it will be only the sounds selected with the flags that I mentioned before. So this could be sounds which are "missing" (have no `SoundAnalysis` object for that analyzer/version), sounds that failed or skipped, etc...

sounds/management/commands/analysis_orchestrator.py
```diff
@@ -0,0 +1,96 @@
...
+    "audioset-vggish:1": "audioset-vggish-extractor:20210615_f44030e"
+}
+
+
+class Command(BaseCommand):
+
+    help = """Checks if there are sounds that have not been analyzed by the analyzers defined in the Analysis 
+    Configuration File (or that are analyzed with older versions) and send jobs to the analysis workers if needed"""
+
+    def add_arguments(self, parser):
```
ffont, line 45: Another useful argument to have would be something like `--max-per-analyzer` which will allow to set a maximum number of analysis jobs to create per analyzer.

sounds/management/commands/gm_worker_analysis.py
```diff
@@ -0,0 +1,145 @@
+#
```
ffont, line 1: All this funcionatilty will be moved to the celery workers so to the analyzer containers. There will be one worker per type of analyzer, but they should all have common funcionality for soudn conversion (you already added that) and also to do the other things done here like checking for free space.

One difference however is that the new analyzers won't have access to Freesound dababase as they won't have a django installation like these workers. We need to think the best way to create the `SoundAnalysis` objects. I'll discuss that with alastair. Maybe we should add some http endpoint in Freesound that given analysis results will create the `SoundAnalysis` objects.

sounds/migrations/0038_merge_20210301_1845.py
```diff
@@ -0,0 +1,16 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('sounds', '0037_soundanalysis_extractor_version'),
+        ('sounds', '0037_license_short_summary'),
```
ffont, line 12: Conflicting migration here, aren't you getting errors with that?

sounds/models.py
```diff
@@ -1102,6 +1102,21 @@ def analyze(self, force=False, high_priority=False):
             }), wait_until_complete=False, background=True, priority=gearman.PRIORITY_HIGH if high_priority else None)
             sounds_logger.info("Send sound with id %s to queue 'analyze'" % self.id)
 
+    def analyze_v2(self, force=False, high_priority=False, analyzer="fs-essentia:1"):
```
ffont, line 1105: This will need to be reimplemented with celery but it is fine to keep it like that until I get more details about how to do the celery thing

ffont (changes requested):
> - Should the main file `freesound/sounds/management/commands/analysis_orchestrator.py` be named differently? "orchestrate_analysis"?

Yeah, I think it makes sense to rename as you propose.

> - I think that `freesound/sounds/management/commands/list_sound_analysis.py` command could be deleted now, as `analysis_orchestrator` with the `--status` flag covers that functions plus others... what do you think?

Right, no need for that command anymore.

> - As you can see, I have created a dictionary at the top of `analysis_orchestrator.py` to be able to 'decode' simple analyzer names and versions (e.g. "fs-essentia:1") into the actual docker images (e.g. "fs-essentia-extractor:20210525_9a3bd10"). On the other side, the Analysis Configuration json file that I have thought of would look like this: 
>     "analyzer:version": ["fs-essentia:1","audio-commons:1", "audioset-vggish:1"]

See my comments in the code. I don't think we need this name mapping.

> - I'm not sure if this is the best way to go. In this case both the version and the simple analyzer name should be sent to `analyze_v2` function and then to the analysis worker (this is not implemented, because maybe it doesn't make sense to you). Maybe this dictionary should be in `freesound/sounds/management/commands/gm_worker_analysis.py`, in a json file, or even in the same json file as the analyzers list. 

Or as I said, maybe no need for any mapping. The docker image can be used as analyzer and version.

sounds/management/commands/analysis_orchestrator.py
gonznm: I thought we wanted to have simplified names and versions, so they were easily comparable – the versions specifically, as integers. I don't know why we wanted this for, though...

sounds/management/commands/analysis_orchestrator.py
gonznm: Because the sounds that haven't been analyzed don't have a SoundAnalysis object... 
If I understand well, the number of missing files for each analyzer would be `len(Sound.objects.all()) - len(SoundAnalysis.objects.filter(analyzer=analyzer, analyzer_version=version))`, do you mean that? 
In any case, right now we are missing a field called `state` in the SoundAnalysis model that is set to "OK", "SK" or "FA" when the analysis finishes, aren't we?

ffont: yes, we're missing that status field. also correct calculating number of missing sounds by subtracting the ones done from the total. however, the way you do it will take long. you should use counts: `Sound.objects.all().count() - SoundAnalysis.objects.filter(analyzer=analyzer, analyzer_version=version).count()` (and also avoid repeating queries, the first one you can keep it in a variable for sure so you don't count for every analyzer)

sounds/management/commands/analysis_orchestrator.py
ffont: yeah, we want simple names. I think we should modify the makefile for the analyzers build process to simply use the folder name as image name. then it would be nice to add a label to the docker image which includes the current date and commit (see example here: https://medium.com/@chamilad/lets-make-your-docker-image-better-than-90-of-existing-ones-8b1e5de950d, using LABEL and arguments should be easy). I found however that it is good that the docker image name is not changing all the time.

if the "version" is not in the folder name, then we're somehow loosing versioning in the docker image name. maybe we could re-add it by somehow asking analyzer developers to specify "version" in a file inside the folder (eg, place a file named "VERSION" with a tag like "v1"), and then from the makefile read that version file and add it to the image name (so we get analyzer_name:version). I think that'd work well.

PR #1571, opened 2021-08-09
Abstract search engine & add pysolr client

xavierfav:
This PR aims at abstracting the search engine functionalities from the web server application. This was originally discussed with @ffont. The main idea was to separate the code specific to the employed search engine backend, in order to facilitate changing the backend (and/or its client).

Additionally, this PR proposes a pysolr-based client as an alternative to the already existing custom client.
The structure of a search engine backend is as follows:
- a` client.py` file containing all the specific code needed for performing the basic search engine related operations.
- a `wrapper.py` file mapping the specific functions to abstracted methods of a `SearchEngine` class.

There are probably some things to improve, and maybe some more serious flaws, but I think it's a good time for a first review or discussion on the way the things were done.


ffont:
@xavierfav what is the status of this PR? I seem to remember @alastair was reviewing it but is that true?

xavierfav:
I think @alastair started to look at it.
If I remember correctly, this PR may need some little more work. At the time, I requested its review to get some feedback.
I could work on it if you find the time to review it.

ffont:
As discussed in our slack channel, I've been working hard on this PR and now it is ready to be merged. As a quick summary, I abstracted all search engine related methods* into a `SearchEngineBase` abstract class and defined a full high-level API for interacting with a search engine and returning query results (with a `SearchResults` object). The API of `SearchEngineBase` is agnostic of the search engine used and therefore I tried to remove as much as possible Solr-related concepts. I implemented a couple of backend options for `SearchEngineBase`, one using our own custom code to interface with Solr and one which uses some methods from `pysolr` but that still uses lots of our own custom code. I also added a management command to be able to test search engine backends, make sure that the `SearchEngineBase` API is implemented and compare results. Check the instructions in https://github.com/MTG/freesound/blob/e3ee5daf003cdd7c7d0e502b1039323815a51002/DEVELOPERS.md#search-engine-backends about running the test command for new or existing backends.

Both the [custom solr backend](https://github.com/MTG/freesound/blob/c559f7c9676645d61c81f0cf1822602899505a8d/utils/search/backends/solr451custom.py) and the [pysolr backend](https://github.com/MTG/freesound/blob/c559f7c9676645d61c81f0cf1822602899505a8d/utils/search/backends/solr451pysolr.py) implemented in this PR pass the tests and return exactly the same results (provided they connect to the same index, of course). The default backend used is still our custom one.


*this includes methods for indexing sounds/forum posts and removing them from the index, as well as searching and some other things like getting a random sound, etc.

PR #1595, opened 2021-12-23
Remove beautifulsoup3

alastair:
**Description**
bs3 isn't supported in python 3. It was only used in 1 place, and we already had bs4 installed anyway.
Add tests for message quoting and replace bs3 usage with bs4

### DIFF
--- a/messages/views.py
+++ b/messages/views.py
@@ -163,10 +163,7 @@ def new_message(request, username=None, message_id=None):
                     raise Http404
                 
                 body = message.body.body.replace("\r\n", "\n").replace("\r", "\n")
-                body = ''.join(BeautifulSoup(body).findAll(text=True))
-                body = "\n".join([(">" if line.startswith(">") else "> ") + "\n> ".join(wrap(line.strip(), 60))
-                                  for line in body.split("\n")])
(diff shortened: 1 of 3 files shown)


messages/views.py
```diff
@@ -181,6 +178,14 @@ def new_message(request, username=None, message_id=None):
     return render(request, 'messages/new.html', tvars)
 
 
+def quote_message_for_reply(body, username):
+    body = ''.join(BeautifulSoup(body).find_all(text=True))
```
alastair, line 182: bs4 also has a `get_text()` method, but it won't return the text inside a `<!--comment-->`, whereas `find_all(text=True)` will. Do we prefer any specific method?

ffont: I don’t think we have preference. If this works as expected we’re fine.

ffont (approved):

messages/views.py
alastair: OK, so let's leave it as-is for now.
I notice that when people write a message we actually send it through bleach (`utils/text.py::clean_html`). The only other place that we use bs4 is to filter http `` tags and turn them into an anchor (in order to prevent cross-protocol errors). We could probably use bleach/html5lib for this too, and remove another dependency.

PR #1614, opened 2022-09-20
Solr 9

alastair:
**Description**

Adds a new solr 9 core and removes the old solr 4 core

The new PointInteger field in solr doesn't support being used as
a unique id field, so we switched id to a string. This requires
casting it back to an int in a few places when using it to retrieve
sounds from the database

date fields can be a DatePoint or DateRange, the Range type is used
for filtering within a range (e.g. created:[from TO to]), but the
point field must be used for ordering, so we add the created field
with both types and choose one depending on what we want to do.

Still some experiments to be done on the search results and similarity types that we will use.

Checklist:

- [x] fix tests that use hardcoded solr4.5 backend
- [x] Add test in `test_search_engine_backend` command for [`get_user_tags`](https://github.com/MTG/freesound/blob/47df8313ecc919c84588195533390b9e20450979/utils/search/__init__.py#L331) method
- [x] Add test in `test_search_engine_backend` command for [`get_pack_tags`](https://github.com/MTG/freesound/blob/47df8313ecc919c84588195533390b9e20450979/utils/search/__init__.py#L343) method
- [x] Update `test_search_engine_backend` to create a new core for running tests against
- [ ] Add test for [`get_stream_sounds`](https://github.com/MTG/freesound/blob/47df8313ecc919c84588195533390b9e20450979/follow/follow_utils.py#L66) (which also performs a search in solr) and refactor some parts of it if needed
- [ ] add a test in `test_search_engine_backend` scripts to test multiple-word queries.
- [ ] Integrate alastair's local solr comparison script into search engine tests (combination of query + filter, including multiple filters and examples with OR on a filter)
- [ ] add a test in `test_search_engine_backend` for geo queries
- [ ] Add tests on date facet
- [ ] handle difference between datepoint and daterange (created)
- [x] review difference in results on 5 vs 9 based on scoring
- [x] Configure default query operator to be `AND` instead of `OR` to replicate the default behaviour of solr < 5
- [x] Move from specific queryhandler to just /select with defaults set
- [x] see if we should use `commitAfter` instead of always commit
- [x] Use BooleanSimilarity for matches (doesn't include ranking based on number of times a word occurs in the document)
- [x] simplify SolrQuery builder
- [x] Update search documentation to suggest new style of geo search
- [x] Ensure that we're using json for sending and retrieving data from solr


utils/search/backends/solr9pysolr.py
```diff
@@ -0,0 +1,107 @@
...
+            only_sounds_with_pack (bool, optional): whether to only include sounds that belong to a pack
+            only_sounds_within_ids (List[int], optional): restrict search results to sounds with these IDs
+
+        Returns:
+            str: processed filter query string.
+        """
+        # Add type suffix to human-readable audio analyzer descriptor names which is needed for solr dynamic fields
+        query_filter = solr555pysolr.add_solr_suffix_to_dynamic_fieldnames_in_filter(query_filter)
+
+        if 'created:' in query_filter:
```
alastair, line 84: add comment explaining why we do this

utils/search/solr9.0.0/docker-entrypoint/create-cores.sh
```diff
@@ -0,0 +1,2 @@
+precreate-core freesound /cores/freesound
+#precreate-core forum /cores/forum
```
alastair, line 2: todo: add forum core

Issue #1083, opened 2017-12-12
Python 3 migration


alastair:
> As a developer and maintainer of Freesound, I want Freesound codebase to run on Python 3 before the Python 2 end of life

Opening a ticket to put out some ideas about the process that we have to take in order to get freesound running on python3

* Check that all of our dependencies run on python3
* Slowly convert our code to be able to run on both python2 and python3
  * Use `__future__` imports to convert syntax and behaviour (print, division, absolute imports)
  * Use six for changed types/modules
* Revise data boundaries for bytes/string data types, use of `unicode()`, `.encode()`, `.decode()`, etc
* Once the code is python3 syntax compatible, start running tests on python3 to see if there are any failures
* Ensure that we have tests for as many pages and processing functions as possible
* At some stage, switch to running python3!

alastair:
Here's a list of current dependencies that don't install on python3:

 * BeautifulSoup (only used once in messages, we should replace with bs4)
 * python-memcached (upgrade available)
 * gearman (gearman3 should be a drop in replacement)
 * akismet (upgrade available)
 * pyyaml (doesn't compile on 3.7, upgrade available)
 * python-cjson (doesn't compile on 3.7, no upgrade available, used for solr, we should replace with newer client)

Bomme:
I propose to use to follow [this guide](https://docs.python.org/3/howto/pyporting.html) in upgrading.
Specifically, I think it would be best to use [futurize](https://python-future.org/automatic_conversion.html) to get a codebase that supports both Python 2.7 and Python 3.3+.
We can upgrade with futurize in two stages to limit the number of changes to review.

[Stage 1](https://python-future.org/futurize.html#stage-1-safe-fixes)

> This applies fixes that modernize Python 2 code without changing the effect of the code. With luck, this will not introduce any bugs into the code, or will at least be trivial to fix. The changes are those that bring the Python code up-to-date without breaking Py2 compatibility. 

[Stage 2](https://python-future.org/futurize.html#stage-2-py3-style-code-with-wrappers-for-py2): 

> The goal for stage 2 is to make further mostly safe changes to the Python 2 code to use Python 3-style code that then still runs on Python 2 with the help of the appropriate builtins and utilities in `future`.

Bomme:
I checked again which dependencies need to be updated for Py3 support, similar to @alastair (see above)

[...]

PR #1620, opened 2022-11-09
automatic conversion from futurize --stage1 with small fixes

Bomme:
**Issue(s)**
This is a first step for Py3 migration #1083 

**Description**
This is the result of applying _Stage 1: “safe” fixes_ of `futurize`.
Probably the most interesting change in this commit is the use of `absolute_import`.
This caused problems with the mutual imports of `apiv2_utils.py` and `exceptions.py` which I fixed by using an absolute import.

I left out `solr451custom.py` and `solr451pysolr.py` to not cause merge problems with #1614 

**Deployment steps**:
None

### DIFF
--- a/_docs/api/add_examples_to_api_docs_source_files.py
+++ b/_docs/api/add_examples_to_api_docs_source_files.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import sys
 import urllib
 sys.path.append("../../apiv2")
@@ -8,7 +9,7 @@ def get_formatted_examples_for_view(view_name):
     try:
         data = examples[view_name]
     except:
-        print 'Could not find examples for view %s' % view_name
+        print('Could not find examples for view %s' % view_name)
         return ''
 
     output = ''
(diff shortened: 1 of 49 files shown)


ffont:
Hi @Bomme,
Thanks for this PR!
I'd like to see @alastair's review, but to me it looks fine.
I'd also include the solr files so alastair can at some point merge master to his branch and fix the merging conflicts.

Is this supposed to be ready to merge or you want to do more stuff with it before merging?

ffont:
Thanks for the review @alastair. I think I misread the names of the solr files and I thought these were for solr5x. I'll go ahead and merge this PR then!

Bomme:
This PR is ready to be merged. The next step will be in a separate PR.

> How did you select for which files you used it?

`futurize` auto detects implicit relative imports and adds the `absolute_import`. So if no change is needed, nothing is changed.

ffont:
Hi, I tested the similarity server and tag recommendation server after your changes and I had to make some modifications in the import to make them work (I was having this error `ValueError: Attempted relative import in non-package`). You did not see these errors because these services are not tested (these are actually non django stuff and run in different containers). Not sure if what I did makes sense (see commit changes https://github.com/MTG/freesound/pull/1620/commits/fc80d89da452059716ac617628f62bac9f8e5914), but services work now. If that is fine, then I think the PR is ready to merge.

ffont (approved):

PR #1622, opened 2022-11-15
Upgrade akismet to 1.0.1

Bomme:
**Issue(s)**
Part of #1083 

**Description**
Before upgrading to Py3 we need to update the [`akismet` dependency to 1.0.1](https://akismet.readthedocs.io/en/1.0.1/upgrade.html)
As part of the upgrade a few tests failed if the akismet API key is not provided; fixed by catching APIKeyError ad ConfigurationError

**Deployment steps**:
None

### DIFF
--- a/accounts/tests/test_spam.py
+++ b/accounts/tests/test_spam.py
@@ -18,8 +18,6 @@
 #     See AUTHORS file.
 #
 
-import mock
-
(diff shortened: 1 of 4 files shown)


ffont:
Thanks @Bomme, that looks great. Merging it!

PR #1623, opened 2022-11-15
Remove unused dependencies

Bomme:
**Issue(s)**
Part of #1083 

**Description**
Remove dependencies that are likely very outdated and no longer used.

**Deployment steps**:
None


ffont:
Thanks!
The LICENSES file actually should be reworked because it is probably vary outdated. I wonder if there will be some tool that generates a list of licenses from a PIP file. We'll need something similar for the JS dependencies. But well, that is something for another PR, just mentioning it here for fun :)

PR #1624, opened 2022-11-19
replace recaptcha-client with django-recaptcha

Bomme:
**Issue(s)**
part of #1083 

**Description**
Replace the outdated `recaptcha-client` dependency with `django-recaptcha`.
The ReCaptcha fields can be tested with test keys that are set as defaults.

**Deployment steps**:
`SILENCED_SYSTEM_CHECKS += ['captcha.recaptcha_test_key_error']`
This line silences a system check that would guard against deployment without configured keys. `RECAPTCHA_PRIVATE_KEY` & `RECAPTCHA_PUBLIC_KEY` should be provided.


ffont:
Awesome @Bomme! I tested it and it works nice. I added a fix for recaptcha field positioning in new UI (BW) and I'll merge now. Thanks!

PR #1625, opened 2022-11-22
Remove unused python dependencies

Bomme:
**Issue(s)**
as part of #1083 

**Description**
Remove dependencies that are currently not used (and outdated) 

**Deployment steps**:
None

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -47,7 +47,6 @@ pysndfile==1.4.3
 python-cjson==1.1.0
 pytz==2020.1
 PyYAML==3.11
-raven==6.10.0
 redis==3.2.0
 sentry-sdk==0.14.4
 Sphinx==1.6.3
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -47,10 +45,8 @@ pyparsing==2.4.7
 pysolr==3.10.0b1
 pysndfile==1.4.3
 python-cjson==1.1.0
-python-memcached==1.53
 pytz==2020.1
 PyYAML==3.11
-raven==6.10.0
```
ffont, line 53: I think we need this for sentry reports right? @alastair

Bomme: Raven is deprecated in favor of Sentry-Python (`sentry-sdk`)

ffont: ah right, I see, I forgot that we made that change :)

ffont:
Thanks @Bomme, merged!

PR #1630, opened 2022-12-08
Return score as a valid field in search results

alastair:
When testing the API, I found that it was useful to have solr's `score` field to compare two search results and see how they differed.

Because we don't return the results from solr directly in the API, store the scores and look them up in the Sound Serializer.

This has been tested in the search endpoint, but I didn't test it yet in the sound view endpoint (in which case it should return no value).

I didn't add this to the solr PR, because I'm not sure how useful it is. Do we really want users to see a score? Does it make sense to expose this? Let me know what you think.


ffont:
I think this is a great idea, I'll review the PR and merge if all looks good.

apiv2/views.py
```diff
@@ -135,14 +135,16 @@ def get(self, request,  *args, **kwargs):
...
+        sound_ids = [ob[0] for ob in object_list]
+        get_analysis_data_for_queryset_or_sound_ids(self, sound_ids=sound_ids)
+        sounds_dict = Sound.objects.dict_ids(sound_ids=sound_ids)
 
         sounds = []
-        for i, sid in enumerate(ids):
+        for i, sid in enumerate(sound_ids):
             try:
-                sound = SoundListSerializer(sounds_dict[sid], context=self.get_serializer_context()).data
+                sound = SoundListSerializer(sounds_dict[sid], context=self.get_serializer_context(), score_map=id_score_map).data
```
ffont, line 147: This is a nice way of passing extra info to the serializer. Same technique should be used for passing analysis data for the `analysis` field. The way I did it until now was to save this data in the view (see ugly code in `get_analysis_data_for_queryset_or_sound_ids`) and then pass the context to the serializer so it can retrieve a reference to the view. However latetly I've been a bit suspicious about that code, thinking that it might lead to mysterious issues in the api when getting analysis data (one day I told you we should talk about these issues). So I think I should follow that same strategy you propose. After merging this PR, I can work on my changes.

PR #1632, opened 2022-12-08
Add search tests for get_user_tags and get_pack_tags

alastair:
With the change to solr 5, we didn't notice that the behaviour of get_pack_tags changed from AND to OR, and so the list of tags in a user's pack was incorrect, instead showing the most common tags that the user had applied over all of fs, not just within this pack. Explicitly use AND in the search query. We're doing a query against /select and not using dismax, so use set_query.

**Issue(s)**

**Description**

**Deployment steps**:


utils/search/backends/test_search_engine_backend.py
```diff
@@ -246,6 +249,44 @@ def sound_check_extra_queries(self):
...
+        so this could by chance not coincide with any of the tags on this sound.
+        Instead, get all tags by the user and check that the ones from solr are a subset of them.
+        """
+        user_tagged_items = TaggedItem.objects.filter(user=sound.user).select_related('tag').all()
+        all_user_tags = [ti.tag.name for ti in user_tagged_items]
+        tags_and_counts = self.search_engine.get_user_tags(sound.user.username)
+        solr_tags = [t[0] for t in tags_and_counts]
+
+        remaining_tags = set(solr_tags) - set(all_user_tags)
+        assert_and_continue(len(remaining_tags) == 0, "get_user_tags returned tags which the user hasn't tagged")
```
alastair, line 264: we don't write anything to the test execution file here, I'm open to ideas for suggestions about what we should write. Perhaps the value of these two sets? The sound id/name?

ffont: The sets of tags could be nice

ffont (approved):
That looks awesome @alastair, I did not merge as I understand you wanted to add something to be printed in the tests log file. Let me know when ready or merge it yourself!

alastair:
sure, I will update this test to add the tags to the output file and then we can merge it.

PR #1640, opened 2022-12-23
Apiv2 fixes

ffont:
Addresses two APIv2 issues:
* Issue in API root page due to wrong construction of dict
* Improves the way sound analysis data is passed to the sound serializers. This can potentially fix some memory leaks by now adding data directly to the view object

### DIFF
--- a/apiv2/apiv2_utils.py
+++ b/apiv2/apiv2_utils.py
@@ -565,18 +565,21 @@ def get_analysis_data_for_queryset_or_sound_ids(view, queryset=None, sound_ids=[
         # Required descriptors are indicated with the parameter 'descriptors'.
         # If 'descriptors' is empty, we return nothing
         descriptors = view.request.query_params.get('descriptors', [])
-        view.sound_analysis_data = {}
+        sound_analysis_data = {}
         if descriptors:
             try:
-                view.sound_analysis_data = get_sounds_descriptors(
+                sound_analysis_data = get_sounds_descriptors(
                     ids, descriptors.split(','), view.request.query_params.get('normalized', '0') == '1',
                     only_leaf_descriptors=True)
             except:
                 pass
         else:
             for id in ids:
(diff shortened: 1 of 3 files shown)


apiv2/serializers.py
```diff
@@ -57,6 +57,12 @@ class AbstractSoundSerializer(serializers.HyperlinkedModelSerializer):
     default_fields = None
 
     def __init__(self, *args, **kwargs):
+        if 'sound_analysis_data' in kwargs:
+            # A sound_analysis_data is a dictionary with sound analysis informartion for sound ids
+            self.sound_analysis_data = kwargs.get('sound_analysis_data')
+            del kwargs['sound_analysis_data']
+        else:
+            self.sound_analysis_data = {}
```
Bomme, line 65:
```suggestion
        self.sound_analysis_data = kwargs.pop('sound_analysis_data', {})
```

apiv2/apiv2_utils.py
```diff
@@ -256,6 +257,18 @@ def initial(self, request, *args, **kwargs):
         self.get_request_information(request)
         self.store_monitor_usage()
 
+    def get_serializer(self, *args, **kwargs):
+        # Overwrite DRF's get_serializer method so that we add sound analysis information when requested
+        serializer_class = self.get_serializer_class()
+        kwargs['context'] = self.get_serializer_context()
+        if 'SoundListSerializer' in str(serializer_class):
+            # If we are trying to serialize sounds, check if we should and sound analysis data to them and add it
+            if type(args[0] == list):
```
Bomme, line 266: `list` might be a bit restrictive. anything iterable should work.
maybe we can use 
`isinstance(args[0], Iterable)` instead?

Bomme (commented):
I left some small comments :)

apiv2/apiv2_utils.py
ffont: Thanks, I'll update that :)

PR #1641, opened 2023-01-05
replace home-made encryption with django cryptographic signing

Bomme:
**Issue(s)**
dealing with a mix of `bytes` and `str` is different in PY3 #1083 

**Description**
Use [Django cryptographic signing](https://docs.djangoproject.com/en/1.11/topics/signing/) to construct secure links that expire.

**Deployment steps**:
Theoretically, a link created before the deployment would not work after. Since it is only valid for 10 seconds, I don't think it is an issue.


ffont:
Thanks for the PR @Bomme, this looks great!

PR #1648, opened 2023-01-11
Drop solr451

Bomme:
**Issue(s)**
#1083 

**Description**
bring changes that remove Solr4 from #1614 to a separate PR.  

**Deployment steps**:


alastair:
The failing tests are because of a change in the exception that gets raised in the case of a connection error.

In current freesound, almost all operations that connect to solr have a try/except/pass block:
https://github.com/MTG/freesound/blob/417f6419e0bbc1791687cd06d5805d0e96f04973/forum/models.py#L180-L184

In fact, when running tests these exceptions have always been raised, as we have no way of disabling index-to-solr during tests. It was just luck that the solr 451 custom backend raised a `SearchEngineException` which allowed us to skip any potential exceptions while running tests.

It seems that tests were running with the solr451 backend, rather than the solr5 pysolr backend.

Now with pysolr, the `pysolr. SolrError` exception isn't getting caught and converted into a `SearchEngineException`.

The easiest immediate fix for this to get us back to where we were would be to correctly catch SolrError and re-throw it as a SearchEngineException. In the long-term we should decide if we want to run solr during tests or not.

docker-compose.yml
```diff
@@ -46,6 +45,7 @@ services:
             - db
             - rabbitmq
             - redis
+            - search
```
Bomme, line 48: @alastair should we keep this line then?

alastair: yes, this is still required, see the following lines:

```yaml
    search:
        image: solr:5.5.5
```

the service is still called `search` so we need to keep it in the depends_on for the website

alastair (approved):
Tests pass, and I was able to run this code and get search results. looks good, thanks!

PR #1650, opened 2023-01-16
assertContains to hand off decoding response content (bytes) to Django

Bomme:
**Issue(s)**
#1083 

**Description**
Response content is encoded in bytes. In Python3 it cannot be compared directly to strings.
Switch tests to `assertContains` as suggested in
https://docs.djangoproject.com/en/1.11/topics/python3/#httprequest-and-httpresponse-objects

**Deployment steps**:
None

### DIFF
--- a/accounts/tests/test_user.py
+++ b/accounts/tests/test_user.py
@@ -67,7 +67,7 @@ class UserRegistrationAndActivation(TestCase):
             u'email2': [u'example@email.com']
         })
         self.assertEqual(resp.status_code, 200)
-        self.assertIn('You must accept the terms of use', resp.content)
+        self.assertContains(resp, 'You must accept the terms of use')
         self.assertEqual(User.objects.filter(username=username).count(), 0)
         self.assertEqual(len(mail.outbox), 0)  # No email sent
 
@@ -80,7 +80,7 @@ class UserRegistrationAndActivation(TestCase):
             u'email2': [u'exampleemail.com']
         })
         self.assertEqual(resp.status_code, 200)
-        self.assertIn('Enter a valid email', resp.content)
+        self.assertContains(resp, 'Enter a valid email')
         self.assertEqual(User.objects.filter(username=username).count(), 0)
         self.assertEqual(len(mail.outbox), 0)  # No email sent
 
@@ -93,7 +93,7 @@ class UserRegistrationAndActivation(TestCase):
             u'email2': [u'example@email.com']
         })
         self.assertEqual(resp.status_code, 200)
-        self.assertIn('This field is required', resp.content)
+        self.assertContains(resp, 'This field is required')
         self.assertEqual(User.objects.filter(username=username).count(), 0)
         self.assertEqual(len(mail.outbox), 0)  # No email sent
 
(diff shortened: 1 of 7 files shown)


accounts/tests/test_user.py
```diff
@@ -370,7 +370,7 @@ def test_user_delete_keep_sounds_using_web_form(self, submit_job):
         self.assertRedirects(resp, reverse('front-page'))
 
         # Check loaded page contains message about user deletion
-        self.assertIn(resp.content, 'Your user account will be deleted in a few moments')
+        self.assertContains(resp, 'Your user account will be deleted in a few moments', status_code=302)
```
Bomme, line 373: I think the two tests with the redirect were broken before.
Debugging the test locally I found:
```
> resp.content
''
```
which would render the previous test as 
```
> '' in 'Your user account will be deleted in a few moments'
True
```

Bomme: Any suggestion for a meaningful test @ffont ?

ffont: Hmm I'm not sure I understand. This is a message that should be rendered in the html page after a user deletes her account. `resp.content` should have the contents of the redirected page. If this is not the case, maybe there is some extra change that makes following redirects not default? I guess if the tests were broken before, they'd be failing?

ffont (commented):
Thanks @Bomme. There are some tests which are not yer passing right? So I'll wait until further changes are made and re-review. Or did I miss something? :)

accounts/tests/test_user.py
alastair: looking at the source code, the message is added using the messages framework:
https://github.com/MTG/freesound/blob/master/accounts/views.py#L1410
and then we do a redirect to the home page

so at this time that the `resp` object is being checked, it's a 302 rediret, with no html content. I think @Bomme is correct that the test never actually worked as expected.

This test should be updated to 1) check that the resp is a redirect to the home page, 2) actually load the home page (or follow the redirect) and check that the string exists correctly.

I see that other cases of `self.assertIn` are like:

```py
self.assertIn('Incorrect password', resp.content)
```

and this test was 
```py
self.assertIn(resp.content, 'Your user account will be deleted in a few moments')
```

which has the parameters around the wrong way

Bomme: Thanks @alastair for explaining 😄 
I think I found a good way of fixing the test.

accounts/tests/test_user.py
```diff
@@ -342,7 +344,7 @@ def test_user_delete_include_sounds_using_web_form(self, submit_job):
         self.assertRedirects(resp, reverse('front-page'))
 
         # Check loaded page contains message about user deletion
-        self.assertIn(resp.content, 'Your user account will be deleted in a few moments')
+        self.assertContains(resp, 'Your user account will be deleted in a few moments')
```
alastair, line 347: https://docs.djangoproject.com/en/4.1/topics/testing/tools/#making-requests

> If you set follow to True the client will follow any redirects and a `redirect_chain` attribute will be set in the response object containing tuples of the intermediate urls and status codes.

One additional test we could do here is ensure that
* we are on the home page
* `resp.redirect_chain` includes the intermediate request to the delete page with the 302 redirect

Bomme: I believe the test in line 344 is exactly testing that.

alastair: oh, excellent! perfect

PR #1653, opened 2023-01-19
use django default_token_generator instead of self-made token for account activation

### LINKED ISSUE #1644: Replace utils.encryption.create_hash with default_token_generator
I want to suggest to replace all usages of `utils.encryption.create_hash` where `add_secret=True` with django's [`default_token_generator`](https://github.com/django/django/blob/stable/1.11.x/django/contrib/auth/tokens.py#L84).

This way we could leave all cryptographic things including handling of the `SECRET_KEY` to Django.

What do you think @ffont & @alastair?

### utils.encryption.create_hash:
```python
def create_hash(data, add_secret=True, limit=8):
    import hashlib
    m = hashlib.md5()
    if add_secret:
        m.update(str(data) + settings.SECRET_KEY)
    else:
        m.update(str(data))
    return m.hexdigest()[0:limit]
```

Bomme:
**Issue(s)**
Closes #1644 

**Description**
use django default_token_generator instead of self-made token

this changes the behaviour of the app since now a token will expire
while before a token was valid "forever"

**Deployment steps**:
This will impact users that started a sign up before the deployment and want to finish it in the new deployment.

### DIFF
--- a/accounts/tests/test_user.py
+++ b/accounts/tests/test_user.py
@@ -19,6 +19,7 @@
 #
 
 import json
+import re
 from unittest import skipIf
 
 import mock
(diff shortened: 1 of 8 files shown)


ffont:
Thanks @Bomme, I tested and works great!

PR #1654, opened 2023-01-19
remove unused open() that might lead to a file descriptor leak

Bomme:


ffont:
Good catch! I remember this function was refactored years ago and this might have been left there :)

PR #1657, opened 2023-01-23
Various fixes for Python 3 compatibility

Bomme:
**Issue(s)**
#1083 

**Description**
This PR is a collection of changes that I accumulated when testing with Python3. I tried to keep commits atomic. 
I suggest to do the review of this PR by commit to see related changes. 

**Deployment steps**:


ffont:
Thanks @Bomme! I looked at it and it seems good. I'll merge, deploy and we'll see if we get any new issues :)

PR #1665, opened 2023-02-02
Automatic rewrites

Bomme:
**Description**
Since we run on Python3 now we can remove a bulk of the compatibility code.
I used
- `future` to re-write language idioms and
- pyupgrade to remove code related to future/six/Python 2

I did my best to keep the commits atomic, so that it's easier to review by commit. Also, if  we don't like a set of changes we can undo the specific commit. 

**Deployment steps**:


ffont:
Thanks @Bomme, this is awesome!
I did some review commit by commit, but there are so many files and lies. We have to trust your review, our tests and the fact that these automatic changes should be harmless. I'll merge and deploy this morning!

PR #1666, opened 2023-02-02
More Python2/3 clean up

Bomme:


ffont:
Thanks again @Bomme, the code looks much better now ;)

Issue #1660, opened 2023-01-29
Upgrade to Django 2


Bomme:
After we reach compatibility with Python 3 (#1083), we can upgrade to the next major version of Django

The [Django upgrade guide](https://docs.djangoproject.com/en/2.0/howto/upgrade-version/) states:
> If you’re upgrading through more than one feature version (e.g. A.B to A.B+2), it’s usually easier to upgrade through each feature release incrementally (A.B to A.B+1 to A.B+2) rather than to make all the changes for each feature release at once. For each feature release, use the latest patch release (A.B.C).
>
>The same incremental upgrade approach is recommended when upgrading from one LTS to the next.

Before we attempt the upgrade we should make sure to:

- [x] Read the [release notes](https://docs.djangoproject.com/en/2.2/releases/) for each ‘final’ release from the one after your current Django version, up to and including the version to which you plan to upgrade.
- [x]  upgrade to the latest version Django-related dependencies 
- [x] resolve any deprecation warnings raised by the current version of Django
- [x] follow deployment steps suggested by Django: https://docs.djangoproject.com/en/2.2/howto/upgrade-version/

ffont:
Hi, upgrading Django is going to be awesome!
The most difficult part will most likely be the nightmare of compatibility between FS dependencies and versions of django. However, to help in this process I went through all FS django-related dependencies and checked with which versions of Django they are compatible. Below I copy a list of my conclusions. I indicate versions of the dependency that are compatible with django 1.11, django 2.2 and 3.2+. I guess first step would be to upgrade all dependencies to the latest 1.11 compatible version (which in most cases is "guaranteed" that it is also 2.2 compatible and in some other cases I'm sure it will also be). Then upgrade django to 2.2. Then upgrade dependencies to latests 2.2 compatible version (which will hopefully be also compatible with django 3.2+). Then upgrade django to 3.2, etc...

Anyway, here is the data I collected:

> django-amazon-ses==3.0.1  

Django compatibility not well specified, our current version will hopefully support Django 2.2
4.0.1 (latest) compatible with Django 2.2+

> django-cors-headers==2.5.3

3.2.1 is the last compatible with Django 1.11 (and is also compatible with 2.2)
3.11.0 is the last compatible with Django 2.2
3.13.0 (latest) compatible with Django 3.2+

> django-debug-toolbar==1.7

2.2.1 is the last compatible with Django 1.11 (and is also compatible with 2.2)
3.1.1 is the last compatible with Django 2.2
[...]

PR #1669, opened 2023-02-07
Upgrade to Django 2.2

ffont:
**Issue(s)**
https://github.com/MTG/freesound/issues/1660

**Description**
This PR upgrades Django to version 2.2. All the steps indicated in https://github.com/MTG/freesound/issues/1660 were followed and tests are passing.

Pending tasks:

- [x] Upgrade django dependencies to versions that will run both on 2.2 and 3.2
- [x] Use `path` instead of `url` in url configuration and use new path syntax where possible (not required but nice improvement) - only partially did that, many apps have not been updated yet
- [x] Read 2.2 release notes
- [ ] Deploy!

**Deployment steps**:
Apply migrations on production. Should not take long.


apiv2/oauth2_urls.py
```diff
@@ -46,6 +46,8 @@ def _wrapped_view_func(request, *args, **kwargs):
     return _wrapped_view_func
 
 
+app_name = 'oauth2_provider'
```
Bomme, line 49: Why is this required?

sounds/templatetags/display_sound.py
```diff
@@ -114,10 +114,10 @@ def sound_object_retrieved_using_bulk_query_id(sound):
             'request': request,
             'is_explicit': sound_obj.is_explicit and
                            (not request.user.is_authenticated or not request.user.profile.is_adult),
-            'is_authenticated': request.user.is_authenticated(),
+            'is_authenticated': request.user.is_authenticated == True,
```
Bomme, line 117: If this is a `bool` we do not need to compare to `True`

apiv2/oauth2_urls.py
ffont: This was to fix one of the deprecation warnings to update to 2.x. When URLs are included using `include`, the included URLs file now requires the `app_name` to be present. See some info: https://docs.djangoproject.com/en/4.1/topics/http/urls/#url-namespaces-and-included-urlconfs

sounds/templatetags/display_sound.py
ffont: This is quite interesting. Some cache-related tests were failing because of that, and what I saw is that is_authenticated was not really a `bool` but a cached bool, and that would make tests fail because `is_authenticated` is used to construct cache keys. So I need the comparison to make this a normal `bool`.

However, now I don't see these errors anymore, even if removing the comparison. It is quite strange because I swear I those tests were failing and it actually took quite a long time for me to track down the issue :S

PR #1670, opened 2023-02-09
Upgrade to Django 3.2

ffont:
**Description**
This PR upgrades migrations to Django 3.2.

- [x] Upgrade Django related dependencies
- [x] Read release notes for 3.0, 3.1 and 3.2
- [x] Modernize code using [django upgrade](https://github.com/adamchainz/django-upgrade)

**Deployment steps**:
Apply migrations on production.

**Things to test manually once deployed**:

- [ ] Make sure that stripe web-hooks are handled correctly as we're checking headers from `request.headers` instead of from `request.META`
- [ ] Make sure sound downloads correctly get range from `request.headers` to not duplicate download objects per range request
- [ ] Make sure client IPs are logged correctly as we now get them from `request.headers`


ffont:
Django 3.2 PR is now ready to deploy. Let's maybe do it tomorrow morning :)

PR #1671, opened 2023-02-10
BW edit and describe pages (part 2)

ffont:
**Description**
This PR continues the implementation of the new describe and edit sound pages in BW UI. Continues from https://github.com/MTG/freesound/pull/1655

**Deployment steps**:
Deploy static files
Also BW sound caches will be broken and should be cleared


sounds/forms.py
```diff
@@ -128,6 +130,49 @@ def __init__(self, pack_choices, *args, **kwargs):
         self.fields['pack'].queryset = pack_choices.extra(select={'lower_name': 'lower(name)'}).order_by('lower_name')
 
 
+def _pack_form_clean_pack_helper(cleaned_data):
+    if 'pack' not in cleaned_data or cleaned_data['pack'] == '' or int(cleaned_data['pack']) == -1:
+        # No pack selected
+        return None
+    elif int(cleaned_data['pack']) == 0:
```
Bomme, line 137: Maybe it's nicer to introduce constants for the magic numbers `-1` and `0`, since they are used quite widely in this PR

ffont: Yeah that is a great suggestion, I'll do it :)

PR #1672, opened 2023-03-01
Change missing is_authenticated check from function to property

alastair:
One missing change - most of these were done in 52668ee2724f4e49a758b3dbc8eac2977f5b7fbe but this one seems to have been missed


ffont:
Hi, thanks @alastair :)
I fixed this in another PR but because it might take time until I merge, I'll merge it here as well.

PR #1679, opened 2023-03-23
fix property access in HTMLElement dataset

Bomme:
dash-style to camelCase name conversion
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset


ffont (approved):
Thanks a lot @Bomme! 
Are you missing Freesound development? :)

PR #1687, opened 2023-06-29
remove a few calls to old_div

Bomme:
**Description**
I removed calls to `past.utils.old_div` and replaced them with the expected division operation (floor division or true division). 
`old_div` was introduced when we `futurize`d the code base.
Once all calls to `old_div` are removed we can drop the dependency on `future`.

'Old' behaviour:
```python
>>> old_div(1, 2)
0
>>> old_div(1.0, 2)
0.5
>>> old_div(1, 2.0)
0.5
>>> old_div(1.0, 2.0)
0.5
```
'New' behaviour:
```python
>>> 1 / 2
0.5
>>> 1 // 2
0
>>> 1.0 // 2
0.0
```

**Deployment steps**:

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -38,7 +38,7 @@ jobs:
     - name: Pull docker images
       run: docker-compose -f docker-compose.test.yml pull
 
-    - uses: satackey/action-docker-layer-caching@v0.0.11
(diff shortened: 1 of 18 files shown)


ffont:
oh wow, @Bomme is back!

Bomme:
> oh wow, @Bomme is back!

to finish what I started

PR #1692, opened 2023-09-19
Remove unused Favorites module

alastair:
**Description**
`Favorite` link in some items isn't used anywhere. Remove it entirely.

This meant that the only item on the SocialModel is tags, so just move tags directly to the items which inherit from SocialModel. Because these are generic foreign keys, there are no migrations

**Deployment steps**:
The table `favorites_favorite` still exists, this can be deleted manually or we can create a migration in general to remove it.

While looking at tables, we also have the django_flatpages, django_flatpages_sites, and south_migrationhistory, stopforumspam_cache, stopforumspam_log tables which we could also remove. I suspect the easiest way to remove these is to just delete them


ffont:
Thanks for this @Alastair! I think it would make sense to remove the unused tables with a migration? Also now there are conflicts that should be fixed. Would you like to do that?

Also mention that the `SocialModel` was used for a number of models, but tags are really only used for sounds. It is still fine to keep the `tags` field in the models other than `Sound` that previously inherited from `SocialModel`, but maybe we want to add a comment next to the field saying that we don't use them?

PR #1693, opened 2023-09-19
Reduce size of test db, anonymize unneeded fields

alastair:
**Description**
We want a new version of the development sample database because it's been a long time since the last one was made.

We decided to reduce the number of sounds and downloads that we keep, in order to make the db smaller.

Additionally, delete tables that contain private information that shouldn't be shared, and anonymise a few further columns.

After removing sounds/downloads, but before cleaning the additional tables, the dump was about 800mb (60 compressed). I noticed that most of the space was used by old silk logs and moderation tickets that were no longer connected to any sounds or users. After deleting this, the database was down to 60mb uncompressed/7mb compressed.

Because it's so small, we could probably get away with increasing the number of sounds a bit more if we want. We'd need to evaluate this against the size of the sound archive.

### DIFF
--- a/general/management/commands/prune_database.py
+++ b/general/management/commands/prune_database.py
@@ -19,19 +19,34 @@
 #
 
 
+import datetime
 import logging
+import os
 import random
 
-from django.contrib.auth.models import User
+from django.contrib.auth.models import User, Group
+from django.contrib.admin.models import LogEntry
+from django.contrib.sessions.models import Session
 from django.core.management.base import BaseCommand
+from django.core.management import call_command
 from django.db import connection
 from django.db.models.signals import post_delete, pre_delete, pre_save, post_save
+from silk.models import Request
+from accounts.models import DeletedUser, EmailBounce, GdprAcceptance, OldUsername, Profile, ResetEmailRequest, SameUser, UserDeletionRequest, UserEmailSetting, UserFlag
+from apiv2.models import ApiV2Client
+from donations.models import Donation
+from forum.models import Subscription
+from general.models import AkismetSpam
+from messages.models import MessageBody
+from oauth2_provider.models import AccessToken, Application, Grant, RefreshToken
 
 import comments
 import forum
 import ratings
 import sounds
+from sounds.models import BulkUploadProgress, DeletedSound, Flag
 import tickets
+from tickets.models import Ticket, TicketComment
 
 console_logger = logging.getLogger('console')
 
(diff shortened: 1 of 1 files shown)


general/management/commands/prune_database.py
```diff
@@ -95,23 +110,36 @@ def delete_some_users(self, userids):
...
+        numusers = len(users)
         console_logger.info(f'Number of uploaders: {numusers}')
-        percentage_remove = 1.0 - (pkeep / 100.0)
-        randusers = sorted(random.sample(userids, int(numusers*percentage_remove)))
-        ch = [c for c in chunks(randusers, 100)]
+
+        totalsounds = 0
+        userids = []
+        for u, numsounds in users:
+            if numsounds <= 1000:
```
alastair, line 133: should be a config option

general/management/commands/prune_database.py
```diff
@@ -130,17 +158,88 @@ def delete_downloaders(self, numkeep):
         random.shuffle(userids)
         # Keep `numkeep` users, and delete the rest
         randusers = sorted(userids[numkeep:])
-        ch = [c for c in chunks(randusers, 100000)]
+        ch = [c for c in chunks(randusers, 1000)]
```
alastair, line 161: can increase chunks to make the operation a bit faster

general/management/commands/prune_database.py
```diff
@@ -130,17 +158,88 @@ def delete_downloaders(self, numkeep):
...
+        for user in User.objects.all():
+            user.email = str(user.id) + '@freesound.org'
+            user.first_name = ''
+            user.last_name = ''
+            # this password is 'freesound'
+            user.password = 'pbkdf2_sha256$36000$PJRTmkaiwSEC$a8+HUj33133PZX7ToOuypT/CfLKNwMeJMXqBJ4QbQPg='
+            user.is_staff = False
+            user.is_superuser = False
+            users_to_update.append(user)
+        #User.objects.bulk_update(users_to_update, ['email', 'first_name', 'last_name', 'password', 'is_staff', 'is_superuser'])
```
alastair, line 178: uncomment - this was only for testing

general/management/commands/prune_database.py
```diff
@@ -130,17 +158,88 @@ def delete_downloaders(self, numkeep):
...
+            last_attempt_of_sending_stream_email=None,
+            is_adult=False,
+            last_donation_email_sent=None,
+            donations_reminder_email_sent=False,
+        )
+
+        # Bookmarks? Possibly set created to now()
+        # Messages (link between 2 users)
+
+        User.objects.all().update(last_login=datetime.datetime.now(), date_joined=datetime.datetime.now())
```
alastair, line 198: I added this because I didn't want to leak login/creation times. Perhaps there are some other date fields which we could include in this anonymisation too

general/management/commands/prune_database.py
```diff
@@ -130,17 +158,88 @@ def delete_downloaders(self, numkeep):
...
+        GdprAcceptance.objects.all().update(date_accepted=datetime.datetime.now())
+        
+        Profile.objects.all().update(
+            is_whitelisted=False,
+            not_shown_in_online_users_list=False,
+            last_stream_email_sent=None,
+            last_attempt_of_sending_stream_email=None,
+            is_adult=False,
+            last_donation_email_sent=None,
+            donations_reminder_email_sent=False,
```
alastair, line 192: The counts in the profile are out of date (num sounds, num downloads, etc). We could run the management command which updates this

general/management/commands/prune_database.py
alastair: after looking at this again, it doesn't seem important to have this an argument. I've moved it to an in-file constant

PR #1694, opened 2023-09-19
Add linting action

alastair:
**Description**
This adds a sample flake8 linting check to freesound. It's based on a generic linter setup that I have in some other projects.
I'm just interested in seeing what kind of errors are reported - it's possible that we might decide that all of these tests are terrible and not really useful. Let's see what happens and if the kind of reports are useful. We can remove checks as necessary if we don't like them, or consider [looking at other flak8 plugins](https://github.com/DmytroLitvinov/awesome-flake8-extensions) to see if we want to add more

I added a basic config file, but we should also consider moving it to a pyproject.toml

It would be nice to use github's built in pip cache to speed up installation.

I've previously had problems with this action, as there are no versions specified in the requirements file. So if a new version introduces new checks, suddenly an unrelated change will result in a bunch of failures. We should consider how we define our dependencies (in a separate linting file, or as part of the dev dependencies?), and decide how we want to specify versions.


alastair:
something I remembered here -
the flake8 linter also checks that the order of imports is "correct" - that they are grouped in stdlib/dependencies/application and that within these groups they're alphabetical.
We should use `isort` to automatically sort these (perhaps at the same step as yapf), so that we don't need to manually go and fix ordering after the test fails

PR #1695, opened 2023-09-19
Add yapf checker

alastair:
**Description**
We have discussed in the past that we want to add a code formatter to freesound in order to be consistent with everything!
This PR adds an action to check formatting with yapf. We had a small discussion about using yapf vs black and yapf edged out slightly.

I've not used this yapf-action before, previously I just installed yapf manually and ran it. We need to check if it works correctly with pyproject.toml

~~Obviously, no actual formatting in this PR (yet!), let's see what happens!~~

Might need to be careful with existing py2 code.


alastair:
it seems like black has won. We're going to switch to using ruff

PR #1696, opened 2023-09-19
Use docker/build-push-action to cache docker build

alastair:
**Description**
Now that we use docker buildx in the unit tests, use the `docker/build-push-action` to allow us to use docker's built-in layer cache when running tests.

This also builds the base image manually instead of using `make`, as the makefile now contains both the py2 and py3 base images and we only need the py3 one.

I manually build the worker image (to cache it), and then run docker-compose build again to correctly tag it. Theoretically this should use existing layers and then tag the image as expected.

I use 2 "cache images" in the github docker cache. This should make the caching operation quite fast because it stays within github instead of going to docker hub.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -12,6 +12,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - uses: actions/checkout@v3
+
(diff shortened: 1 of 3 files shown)


alastair:
giving up on this, we can probably make builds faster by just using uv instead

PR #1697, opened 2023-09-19
Only show tickets to the user who created it, or moderators

alastair:
**Description**
Tickets should only be visible by the user who created them or admins/moderators. Users should be signed in to view them.

Remove old view which I can't find any reference to in the rest of the code.


Bomme:
LGTM!

PR #1699, opened 2023-09-27
update to latest sentry-sdk

Bomme:
**Description**
I noticed that we are running a quite old `sentry-sdk` install. Since we recently upgraded the `sentry` version it might be good to keep up on the application side.

I didn't want to go to [`1.10.0`](https://github.com/getsentry/sentry-python/releases/tag/1.10.0) because of breaking changes in dashboards and discover queries and I wasn't sure if we run any of those.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -45,7 +45,7 @@ pysndfile==1.4.4
 pytz==2023.3
 PyYAML==6.0.1
 redis==3.2.0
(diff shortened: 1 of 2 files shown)


alastair:
we're not using any extra functionality in sentry, let's just go straight to the latest sdk version

PR #1702, opened 2023-10-16
Upgrade to celery 5

alastair:
**Description**
Celery 4 doesn't work in python 3.11, therefore we need to upgrade to celery 5

We were using the old `@task` api in celery 4, which was already deprecated, and removed in v5. Switch to `shared_task`

**Deployment steps**:
I tested that celery started, but I haven't tested that it works as expected


ffont:
I tested celery tasks and it seems to work fine, I'll merge this.

PR #1703, opened 2023-10-23
refactor images.py to use PIL ImageOps for much simpler code

Bomme:
I was following the tutorial in
https://pillow.readthedocs.io/en/stable/handbook/tutorial.html#relative-resizing

I started refactoring since I wanted to remove all calls to `old_div` but found it was easier just to rewrite large portions relying on PIL library code

### DIFF
--- a/utils/images.py
+++ b/utils/images.py
@@ -17,62 +17,45 @@
 # Authors:
 #     See AUTHORS file.
 #
+from PIL import Image, ImageOps
+from PIL.Image import Resampling
 
-from past.utils import old_div
-from PIL import Image
 
-class ImageProcessingError(Exception):
-    pass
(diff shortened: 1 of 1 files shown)


Bomme:
### Differences between old and new code


#### Original image (97x128)
97_128shape
---
Note: I added a one pixel border in black for demonstration purposes

 size | old | new
-------- | -------- | --------
50x50   |  50_50shape_old_border  | 50_50shape_new_border
100x100 | 100_100shape_old_border| 100_100shape_new_border
150x150 |  150_150shape_old_border  | 150_150shape_new_border

ffont (approved):
That looks great, thanks for the contribution!
The white added space will not look nice for some avatars (specially in dark mode), but then again users can fix that by uploading a properly squared avatar so this is fine.

PR #1706, opened 2023-10-31
enable sentry transaction sampling

Bomme:
https://docs.sentry.io/platforms/python/configuration/sampling/


ffont:
I merged this PR with the transaction sampling rate of 10%. I'll deploy and see how it goes. We ca further adjust this in the future.

PR #1711, opened 2023-11-11
let Django cleaning handle amount validation; show errors close to input

Bomme:
**Issue(s)**
FREESOUND-WEB-1QZ

**Description**
Currently, no max amount value is set on the donation form.
With this change I leave all form validation to Django and only adapt the display of the error in the JS code.

Django gives an error if the amount is not within the specified min and max value range. We'll show this error right below the input field. All other errors (none expected currently) will be shown in the `errorlist` as it was done before.

**Deployment steps**:
None

### DIFF
--- a/donations/forms.py
+++ b/donations/forms.py
@@ -1,26 +1,33 @@
-import json
 import base64
+import json
+
 from django import forms
 from django.utils.safestring import mark_safe
 
 from .models import DonationCampaign
 
+
 class DonateForm(forms.Form):
     RADIO_CHOICES = []
 
     donation_type = forms.ChoiceField(
-        widget=forms.RadioSelect(), choices=RADIO_CHOICES,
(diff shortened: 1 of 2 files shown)


ffont:
Thanks for this @Bomme! As I understand, you added the max value in the form so Django does the validation. Why do we need the JS bit however? Aren't the errors also rendered by Django like in other forms?

Bomme:
> you added the max value in the form so Django does the validation.
Yes, I felt this was the cleanest way to enforce the limits on the amount.

> Aren't the errors also rendered by Django like in other forms?
No, since the form submission is never *really* done. The buttons have their own logic in JS and make the requests to the payment providers directly in the frontend code. (see the eventListeners)

If we would handle the redirect to the payment provider sides in the backend we could do away with all this complicated error logic. But I didn't want to tackle this here since it's only about the upper limit.

PR #1739, opened 2023-12-19
enable password validation

Bomme:
**Description**
This PR introduces [password validation](https://docs.djangoproject.com/en/3.2/topics/auth/passwords/#enabling-password-validation) for user registration and password change.

Up to now, there was no password validation done and consequently no password requirements enforced. 

The settings are taken from the default Django `settings.py`.

If I'm not mistaken there should be no effect on currently stored passwords. Only, new (and changed) passwords will need to fulfill the requirements.

Conveniently, the default password change form will inform the user about the password requirements (see below).

**Before**:
Bildschirmfoto 2023-12-19 um 11 20 26

**After**:
Bildschirmfoto 2023-12-19 um 11 18 21

**Deployment steps**:
None

### DIFF
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -26,9 +26,11 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.forms import PasswordResetForm, AuthenticationForm, SetPasswordForm, PasswordChangeForm
 from django.contrib.auth.models import User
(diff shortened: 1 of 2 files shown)


ffont:
Thanks @Bomme! That unfortunately means I won't be able to use 12345678 as my test password anymore :S

PR #1748, opened 2024-01-04
remove a few more calls to old_div

Bomme:
**Issue(s)**
#1633 introduced many calls to `old_div` that are the last reason we have a dependency on the `future` package.
This PR removes all in the core code components. `tagrecommendation` and `clustering` are still todo.

**Description**

**Deployment steps**:


ffont:
Thanks @Bomme!

PR #1750, opened 2024-01-07
Nodejs install did not install npm

Bomme:
**Description**
give preference to the nodesource repo when installing nodejs
since debian bookworm nodejs was installed from the debian repo and not
from the nodesource repo. This package did not require npm by default.
Using the installer script solves the issue.

See: https://github.com/nodesource/distributions/issues/1725

**Deployment steps**:

### DIFF
--- a/docker/Dockerfile.base
+++ b/docker/Dockerfile.base
@@ -1,4 +1,3 @@
-ARG PYTHON
 FROM python:3.10
 
 LABEL org.opencontainers.image.authors="support@freesound.org"
--- a/docker/Dockerfile.workers_web
+++ b/docker/Dockerfile.workers_web
@@ -13,9 +13,7 @@ RUN make clean && make
 
 FROM freesound:2023-07
 
-RUN mkdir -p /etc/apt/keyrings/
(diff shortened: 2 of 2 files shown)


alastair:
hahaha, what? 
https://github.com/nodesource/distributions

> Installation Scripts: Back by popular demand, the installation scripts have returned and are better than ever. See the installation instructions below for details on how to use them.

So.. they decided to remove the script (https://github.com/MTG/freesound/pull/1691), but now it's back? I have no idea. Happy to merge this if they recommend it again. Let's count down the days until they remove it again

ffont:
Ahh, all node-related stuff is so frightening to me...
Did you try that with an image built with this you can install all static dependencies and build them?

Bomme:
yes, I could install and build.

The only thing that scared me was the `--force` in `docker compose run --rm web npm install --force` 😆

Bomme:
Alternatively we could remove the nodesource stuff and explicitly take the `nodejs` version from debian (currently 18.19.0) https://packages.debian.org/bookworm/nodejs

ffont:
I like to specify the version so let's keep it how it is (after your change). Thanks!

PR #1752, opened 2024-01-12
Simplify check in flag_user

Bomme:
**Issue(s)**
FREESOUND-WEB-1PD

**Description**
I found this because of a N+1 Query alert in sentry. When a sound has many comments from users that could be flagged it runs a new query for every comment that is displayed on the site.
While this PR does not solve this issue it should simplify and speed up the check.

There are the following changes:
1. I move the check if the commenter has sounds from the template (`{% if user_sounds < 1 %}`) to the function so that the logic is no longer split across two files
2. `flagged` was a list but finally only used as a bool. By using `exists()` we make it explicit and hopefully slightly faster.
3. remove `'done_text': "Marked as spam/offensive",  # Not used in BW`

**Deployment steps**:


ffont:
That looks great, thanks @Bomme!
There are I think 5 comments per page, so I guess the multiple queries is not too bad in that case. Also if I understood correctly, to improve this the only think we could do is to always show the report button, and then if user can't be flagged show a message if the request user is trying to flag her. But yeah this is for another PR.

PR #1760, opened 2024-05-02
SOLR schema: convert license to a docValues field for faster faceting

Bomme:
**Description**
Change the type of `license` to `string` similar to other fields to mark `docValues` as `true`. This should [help speed up faceting of search results.](https://solr.apache.org/guide/solr/9_2/indexing-guide/docvalues.html#using-docvalues)

The `fieldType` definition looks like this: 
```<fieldType name=“string” class=“solr.StrField” sortMissingLast=“true” docValues=“true”/>```

**Deployment steps**:
Re-indexing required!

### DIFF
--- a/freesound/settings.py
+++ b/freesound/settings.py
@@ -564,7 +564,7 @@ ANALYZERS_CONFIGURATION = {
 # -------------------------------------------------------------------------------
 # Search engine
 
-FCW_FILTER_VALUE = '("attribution" OR "creative commons 0")'
+FCW_FILTER_VALUE = '("Attribution" OR "Creative Commons 0")'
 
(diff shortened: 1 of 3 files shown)


ffont:
This is nice, thanks @Bomme !
Did you check if there are other fields used in faceting that might need similar treatment?

Also in beta search features we do faceting with dynamic fields as well, and I wonder how this would be affecting faceting performance. We'll have to investigate these issues before making some of these extra facets public.

ffont (approved):
Thanks for the PR @Bomme !
I added an extra change to make the "free cultural works" filter work with the now case sensitive `license` field. Now this should be ready to merge when Solr is reindexed :)

PR #1781, opened 2024-06-18
custom management command that sets an announcement banner on front page

Bomme:
needs more styling

**Issue(s)**

**Description**

**Deployment steps**:


general/management/commands/announcement_cache.py
```diff
@@ -0,0 +1,45 @@
...
+        parser.add_argument("action", type=str, help="Indicates whether to set or clear the cache")
+        parser.add_argument("title", type=str, nargs="?", help="Title of the announcement")
+        parser.add_argument("text", type=str, nargs="?", help="Text of the announcement")
+
+    def handle(self, *args, **kwargs):
+        action = kwargs["action"]
+        title = kwargs["title"]
+        text = kwargs["text"]
+
+        announcement_cache_key = "announcement_cache"
```
ffont, line 31: This could be moved to settings.py

general/management/commands/announcement_cache.py
```diff
@@ -0,0 +1,45 @@
...
+        if action == "set":
+            one_day = 60 * 60 * 24
```
ffont, line 33: The number of days could be a parameter of the command as well, and I'd say that if not specified, then it could have no expiration

ffont (commented):
Thanks @Bomme! I left a couple of comments and I'll work on the styling part. I also think there's some missing change in the view of the front page where the contents of `announcement_cache` are passed to the front page template. I guess you did not commit that file?

general/management/commands/announcement_cache.py
Bomme: Moved to settings.py

general/management/commands/announcement_cache.py
Bomme: is now an optional argument to the command

Bomme:
@ffont I addressed your comments and added the missing line!

ffont:
Cool! I added the styling. Also renamed the management command to `announcement_banner` which I think will be easier to remember. You can try if you want and think of a text for your own announcement. I would use no title, I think there's no need. After my updated, you can pass empty title and the message will be rendered fine.

Issue #1577, opened 2021-11-21
Allow renaming of bookmarks


Headphaze:
I just noticed that it's not possible to rename saved bookmark categories. This should definitely be allowed. 

Cheers
Sam

PR #1799, opened 2024-11-27
Enable a modal to edit bookmark categories

quimmrc:
**Issue(s)**
https://github.com/MTG/freesound/issues/1577

**Description**
addition of a basic modal to rename bookmark categories

**Deployment steps**:
1. new edit bookmark category modal
2. new method to validate data in BookmarkCategoryForm
3. new view funcion edit_bookmark_category
4. in modals.js - changes to handelGenericModalsWithForms, including new parameter dataReloadOnSuccess

### DIFF
--- a/bookmarks/forms.py
+++ b/bookmarks/forms.py
@@ -30,7 +30,15 @@ class BookmarkCategoryForm(forms.ModelForm):
         widgets = {
             'name': forms.TextInput(attrs={'class': 'category_name_widget'}),
         }
+    
+    def clean(self):
+        cleaned_data = super().clean()
+        name = self.cleaned_data.get("name", )
 
+        if BookmarkCategory.objects.filter(user=self.instance.user, name=name).exists():
+            raise forms.ValidationError("This name already exists for a bookmark category")
+        
+        return cleaned_data
 
 class BookmarkForm(forms.Form):
     category = forms.ChoiceField(
--- a/bookmarks/views.py
+++ b/bookmarks/views.py
@@ -27,7 +27,7 @@ from django.http import Http404, HttpResponseRedirect, JsonResponse
 from django.shortcuts import get_object_or_404, render
 from django.urls import reverse
 
-from bookmarks.forms import BookmarkForm
+from bookmarks.forms import BookmarkForm, BookmarkCategoryForm
 from bookmarks.models import Bookmark, BookmarkCategory
 from sounds.models import Sound
 from utils.pagination import paginate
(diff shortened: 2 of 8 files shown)


bookmarks/forms.py
```diff
@@ -30,7 +30,15 @@ class Meta:
         widgets = {
             'name': forms.TextInput(attrs={'class': 'category_name_widget'}),
         }
+    
+    def clean(self):
+        cleaned_data = super().clean()
+        name = self.cleaned_data.get("name", )
 
+        if BookmarkCategory.objects.filter(user=self.instance.user, name=name).exists():
+            raise forms.ValidationError("This name already exists for a bookmark category")
```
ffont, line 39: This error message could be improved. Maybe "You have already created a bookmark category with that name"

bookmarks/views.py
```diff
@@ -85,6 +85,33 @@ def delete_bookmark_category(request, category_id):
...
+def edit_bookmark_category(request, category_id):
+
+    if not request.GET.get('ajax'):
+        return HttpResponseRedirect(reverse("bookmarks-for-user", args=[request.user.username]))
+    
+    category = get_object_or_404(BookmarkCategory, id=category_id, user=request.user)
+
+    if request.method == "POST":
+        edit_form = BookmarkCategoryForm(request.POST, instance=category)
+        print(edit_form.is_bound)
```
ffont, line 98: :astonished:

bookmarks/views.py
```diff
@@ -85,6 +85,33 @@ def delete_bookmark_category(request, category_id):
...
+        if edit_form.is_valid():      
+            category.name = edit_form.cleaned_data["name"]
+            category.save()
+            return JsonResponse({"success":True})
+        if not edit_form.is_valid():
+            print(edit_form.errors.as_json())
```
ffont, line 104: 😲 😲 
I guess this whole `if` clause can be removed

freesound/static/bw-frontend/src/components/modal.js
```diff
@@ -195,7 +198,7 @@ const handleGenericModal = (fetchContentUrl, onLoadedCallback, onClosedCallback,
 };
 
 
-const handleGenericModalWithForm = (fetchContentUrl, onLoadedCallback, onClosedCallback, onFormSubmissionSucceeded, onFormSubmissionError, doRequestAsync, showLoadingToast, modalActivationParam) => {
+const handleGenericModalWithForm = (fetchContentUrl, onLoadedCallback, onClosedCallback, onFormSubmissionSucceeded, onFormSubmissionError, doRequestAsync, showLoadingToast, modalActivationParam, dataReloadOnSuccess) => {
```
ffont, line 201: `dataReloadOnSuccess` should probably be called `triggerPageReloadOnSuccess` for making it fully self-explainable

freesound/static/bw-frontend/src/components/modal.js
```diff
@@ -85,19 +85,22 @@ const bindConfirmationModalElements = (container) => {
...
-const handleDefaultModal = (modalUrl, modalActivationParam, atPage) => {
-  if ((atPage !== undefined) && modalUrl.indexOf('&page') == -1){
-    modalUrl += '&page=' + atPage;
-  }
+const handleDefaultModal = (modalUrl, modalActivationParam) => {
   handleGenericModal(modalUrl, undefined, undefined, true, true, modalActivationParam);
 }
 
+const handleDefaultModalWithForm = (modalUrl, modalActivationParam) => {
+  handleGenericModalWithForm(modalUrl, undefined, undefined, (req) => {showToast('Form submitted succesfully!')}, undefined, true, true, modalActivationParam, true);
```
ffont, line 93: You are enabling the reload on success by default, and also not providing a way to disable it unless `handleGenericModalWithForm` is used directly instead of an automatic binding. You should probably add a parameter to `handleDefaultModalWithForm` (e.g. `reloadOnSuccess`) which defaults to false, and then in the bind function, see if the element has a specific dataset property to decide if `reloadOnSuccess` should be true or false (e.g. `data-reload-on-success`)

ffont: Ok, it might be a good idea to simply pass `element` (also when binding), you can do it like that

Issue #1290, opened 2019-04-08
Download sounds of bookmark category in single zip file


ffont:
> As a user I want to be able to create lists of sounds bookmarked in categories and then download them all together in a single ZIP file similar to when downloading a pack

We can do that with relative ease re-using downloading code for the packs use case: https://github.com/MTG/freesound/blob/4616b3410cd171a7b3ec45ce431b11be878555b3/sounds/views.py#L370-L395

Nevertheless we have to decide how to store these downloads. I guess we should create `BookmarkDownload` and `BookmarkSoundDownload` tables like in the packs case. Or we could decide to not store these downloads in DB for now and wait until we refactor bookmarks into _Collections_. Food for thought...

ffont:
Hi,
This is not a script to download a zip of packs. This is currently not possible. What you could do is use the API to automate downloads. You can check documentation here http://freesound.org/docs/api/

cheers

aarora8:
Thank you, I am now able to download my bookmarked sounds using the freesound API and OAuth2 authentication.

PR #1800, opened 2024-12-02
bookmark category download implementation

quimmrc:
**Issue(s)**
https://github.com/MTG/freesound/issues/1290

**Description**
Implementation of new view for bookmarkcategory downloads, new method to get the sound attributions for a bookmark category and download_sounds adaptation to make it viable for both packs and bookmark categories

Errors: download_sounds redirects to a url with the text content of the download instead of downloading the desired zip file

Captura de pantalla de 2024-12-02 12-05-14

### DIFF
--- a/templates/bookmarks/bookmarks.html
+++ b/templates/bookmarks/bookmarks.html
@@ -35,7 +35,7 @@
             <div class="col-lg-8">
                 <h4><span class="text-light-grey">Category:</span> {% if category %}{{category.name}}{% else %}Uncategorized{% endif %} </h4>
                 <div class="v-spacing-top-3">
-                {% if page_bookmarks_and_sound_objects %}
+                {% if page.object_list %}
                     <div class="row">
                         {% for bookmark, sound in page_bookmarks_and_sound_objects %}
                             <div class="col-6 col-md-4">


bookmarks/views.py
```diff
@@ -83,6 +84,18 @@ def delete_bookmark_category(request, category_id):
         return HttpResponseRedirect(next)
     else:
         return HttpResponseRedirect(reverse("bookmarks-for-user", args=[request.user.username]))
+    
+@transaction.atomic()
+def download_bookmark_category(request, category_id):
+    category = get_object_or_404(BookmarkCategory, id=category_id)
+    licenses_url = (reverse('bookmark-category-licenses', args=[category_id]))
+    #missing: cache checking done in packdownload
```
ffont, line 92: I would suggest to improve this comment, something like: # NOTE: unlike pack downloads, here we are not doing any cache check to... 

Make it clear, so if we look at it in the future we understand it easily.

utils/downloads.py
```diff
@@ -24,26 +24,32 @@
...
 
+from bookmarks.models import Bookmark, BookmarkCategory
 from donations.models import DonationsModalSettings
-from sounds.models import Download, PackDownload
+from sounds.models import Download, Pack, PackDownload, Sound
 from utils.nginxsendfile import prepare_sendfile_arguments_for_sound_download
 
 
-def download_sounds(licenses_url, pack):
+def download_sounds(licenses_url, batch):
```
ffont, line 33: I kind of don't like the name `batch`. Currently it should be some type of object that implements a method called `get_attribution` which should return the same thing that is returned in the `licenses_url` argument. But this is not well explained. It wasn't well explained before either. But a good practice we try to follow is to take the opportunity to improve code documentation when we refactor things, so maybe it could be documented now.

I might consider having 3 arguments `licneses_file_url`, `licenses_file_content` and `sound_list`, and then add proper documentation for these arguments. Copilot should probably help you. We follow google style for python documentation, see for example [this function as an example](https://github.com/MTG/freesound/blob/09ad86db4b30540caf1bcc04c694224a8bea5ee1/utils/search/search_sounds.py#L78).

ffont (changes requested):
I left a couple of comments to make code a bit cleaner and better documented, but other than that the PR looks great. The "error" you mention (that clicking the download link redirects to a file list), is also happening with pack downloads right? This is because to turn this into a zip file we use some specific server tricks which are not available in the development server. If the behaviour is the same as for the pack downloads, then it should be working when deployed.

templates/bookmarks/bookmark_category_attribution.txt
```diff
@@ -0,0 +1,21 @@
+{% load absurl %}Bookmark Category downloaded from Freesound
```
ffont, line 1: I was thinking if it would make sense to reuse some template/method code between packs and bookmark categories. Did you think  about that?

ffont:
Oh, and please edit the title of the PR, don't say this is an "attempt". I guess everything that we do is just initially an attempt :)

templates/bookmarks/bookmark_category_attribution.txt
quimmrc: I am reusing the pack_attributions.txt file for bookmark categories as well, shall I modify the "pack" naming of the variable in the txt file to find something suitable for both kind of classes? I have made it work to return a proper .txt file in either case using "pack" naming but I don't know if it might be confusing in the future:

image

ffont: I did not realize you were reusing the same file. Then I might rename the file to `multiple_sounds_download_attribution.txt`, and pass `pack` or `bookmark_category` as needed, then check `type` to see if you should use `pack.name` or `bookmark_category.name`.

quimmrc: I was not reusing the same file in fact, I just tried after you suggested it xD

apiv2/views.py
```diff
@@ -702,7 +702,9 @@ def get(self, request,  *args, **kwargs):
                 msg='Sounds in pack %i have not yet been described or moderated' % int(pack_id), resource=self)
 
         licenses_url = (reverse('pack-licenses', args=[pack.user.username, pack.id]))
-        return download_sounds(licenses_url, pack)
+        licenses_content = pack.get_attribution()
+        sound_list = pack.sounds.filter(processing_state="OK", moderation_state="OK").select_related('user', 'license')
```
ffont, line 706: I imagine that inside `get_attribution` we already do this query `pack.sounds.filter(processing_state="OK", moderation_state="OK").select_related('user', 'license')` so we are doing it twice. We should improve that. Maybe `get_attribution` could get an optional argument `sound_qs` (`qs`stands for queryset), and when the argument is provided, it does not do the query inside `get_attribution` and it uses the provided queryset. Simlar logic should be applied for the bookmarks case. Do you think it makes sense?

bookmarks/views.py
```diff
@@ -84,20 +84,25 @@ def delete_bookmark_category(request, category_id):
         return HttpResponseRedirect(next)
     else:
         return HttpResponseRedirect(reverse("bookmarks-for-user", args=[request.user.username]))
-    
+
+@login_required
```
ffont, line 88: good catch!

bookmarks/views.py
```diff
@@ -84,20 +84,25 @@ def delete_bookmark_category(request, category_id):
...
 def download_bookmark_category(request, category_id):
     category = get_object_or_404(BookmarkCategory, id=category_id)
-    licenses_url = (reverse('bookmark-category-licenses', args=[category_id]))
-    #missing: cache checking done in packdownload
-    return download_sounds(licenses_url, category)
+    licenses_content = category.get_attribution()
+    licenses_url = (reverse('category-licenses', args=[category_id]))
+    
+    bookmarked_sounds = Bookmark.objects.filter(category_id=category.id).values("sound_id")
+    sounds_list = Sound.objects.filter(id__in=bookmarked_sounds, processing_state="OK", moderation_state="OK").select_related('user','license')
```
ffont, line 96: Same comment as before, try to do it in a way in which this query is only performed once.

sounds/models.py
```diff
@@ -1795,9 +1795,10 @@ def get_attribution(self):
         users = User.objects.filter(sounds__in=sounds_list).distinct()
         # Generate text file with license info
         licenses = License.objects.filter(sound__pack=self).distinct()
-        attribution = render_to_string("sounds/pack_attribution.txt",
-            dict(users=users,
-                pack=self,
+        attribution = render_to_string("sounds/multiple_sounds_attribution.txt",
+            dict(type=self.__class__.__name__,
```
ffont, line 1799: wow, this is ugly :S 
just hardcode the name, e.g. `type="pack"`, it is clearer.

utils/downloads.py
```diff
@@ -24,32 +24,29 @@
...
-    From a list of sounds generates the HttpResponse with the information of
-    the wav files of the sonds and a text file with the license. This response
+def download_sounds(licenses_file_url, licenses_file_content, sounds_list):
+    """From a list of sounds generates the HttpResponse with the information of
+    the wav files of the sounds and a text file with the license. This response
     is handled by mod_zipfile of nginx to generate a zip file with the content.
+
+    Args:
+        licenses_file_url (str): url to the sound Pack or BookmarkCategory licenses
+        licenses_file_content (django SafeString): attributions for the different sounds in the Pack or BookmarkCategory
```
ffont, line 39: These `django` things were suggested by copilot or some other AI, or did you find about that elsewhere?

bookmarks/models.py
```diff
@@ -38,9 +38,10 @@ def get_attribution(self):
         users = User.objects.filter(sounds__in=sounds_list).distinct()
         # Generate text file with license info
         licenses = License.objects.filter(sound__id__in = sounds_list).distinct()
-        attribution = render_to_string("bookmarks/bookmark_category_attribution.txt",
-            dict(users=users,
-                category=self,
+        attribution = render_to_string(("sounds/multiple_sounds_attribution.txt"),
+            dict(type=self.__class__.__name__,
```
ffont, line 42: `type="bookmark_category"` (or whatever you use to compare in the template). It will be clearer for our future selves this way.

ffont (changes requested):
Thanks @quimmrc! I left some other comments, this is almost there!

utils/downloads.py
quimmrc: It wasn't provided by neither copilot nor an AI tool, I can't remember exactly how I got that but probably by looking at Django official documentation or by looking into the type of the variable. I'll check it again and see what it really is.

quimmrc: definetly took it from the "type" of the variable "licenses_file_content", as it returns:

<class 'django.utils.safestring.SafeString'>

However I would agree that it's not a very friendly name,  comes from the "render_to_string" from a .txt file

ffont (approved):

PR #1802, opened 2024-12-11
uncategorized bookmarked sounds download

quimmrc:
**Issue(s)**
#1290 - uncategorized bookmarked sounds

**Description**
Make available the download of uncategorized bookmarked sounds for a user. 
New view function to handle these type of sounds, which since they do not correspond to a BookmarkCategory object they need slight changes in their handling.

**Deployment steps**
1. Download uncategorized bookmarks id set to 0 (URL handling).
2. download_bookmark_category view computes necessary information according to category_id
3. uncat_bookmark_data returns all necessary information to download uncategorized bookmarks. It only uses the "request" parameter because it needs to be called both from "download_bookmark_category" and "bookmark_category_licenses" views, and I did not know how to handle the latter with suitable parameters for the purpose.
4. bookmark_category_licenses view returns url according to category_id
5. .txt template contemplates receiving "not_an_object" for its rendering, since uncategorized bookmarks do not group in any object itself (I think).

### DIFF
--- a/bookmarks/tests.py
+++ b/bookmarks/tests.py
@@ -69,7 +69,7 @@ class BookmarksTest(TestCase):
         self.assertEqual(1, len(response.context['page'].object_list))  # 1 bookmark uncategorized
         self.assertEqual(1, len(response.context['bookmark_categories']))  # 1 bookmark cateogry
 
-        # Test bookmark cateogry page
+        # Test bookmark category page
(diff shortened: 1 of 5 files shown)


quimmrc:
Further work to do: 

- add unit tests for licenses views (one for packs and one for bookmark categories)

ffont:
Thanks for the PR! After our discussion at the office we decided that we would not merge this PR as it adds a bit of weirdness and we'll replace this with collections "soon" anyway.

Issue #1181, opened 2018-05-28
Uploading file with same file name overwrites description queue file without warning


qubodup:
1. Upload a.mp3 (the first file)
2. Upload a different a.mp3 (the second file)

Expected:
You have two "a.mp3" files in the description queue

Actual:
You have one "a.mp3" file in the description queue. It is the second file.

Users should be at least warned about this.

Note:
I was doing this to test the indev batch describer

ffont:
Good catch!
Adding it to the list of current issues.

PR #1806, opened 2024-12-16
Enable upload of different sound files with the same filename

quimmrc:
**Issue(s)**
#1181

**Description**
Sound files uploaded with the same file name got overwritten in the description queue.
In order to solve that, my proposal is to change the filenames adding an index as done in most of OS files handling (for example, for download), e.g.:

a.mp3
a(1).mp3

**Deployment steps**:
`upload()` creates a submitted files list, using this list:

1) Create a new list to store filenames
2) For each file name, check if the name already exists in the list and if so, give it an index for as many times as the name has already appeared.
3) Update the filename with the index appropiately

Note: all filenames will be added to the filenames list for indexing purposes (for the duplicated ones, addition must be done before filename modification).

### DIFF
--- a/accounts/views.py
+++ b/accounts/views.py
@@ -1284,7 +1284,18 @@ def upload(request, no_flash=False):
             form = UploadFileForm(request.POST, request.FILES)
             if form.is_valid():
                 submitted_files = request.FILES.getlist('files')
+                files_names = list()
                 for file_ in submitted_files:
+                    #check for duplicated names and add an identifier, otherwise, different files with the same
+                    #name will be overwritten in the description queue
+                    if file_.name in files_names:                        
+                        name_counter = files_names.count(file_.name) 
+                        files_names.append(file_.name) 
+                        name, extension = os.path.splitext(file_.name)
+                        file_.name = "%s(%d)%s" % (name, name_counter, extension) 
+                    else:
+                        files_names.append(file_.name)
+                    
                     if handle_uploaded_file(request.user.id, file_):
                         uploaded_file = file_
                         successes += 1


ffont:
We should add a unit test for that, you can probably add an extra test method next to this one: https://github.com/MTG/freesound/blob/8f29b2178fabc636abd01b0061153d17f3d7d184/accounts/tests/test_upload.py#L44

ffont:
Thanks @quimmrc! Once you have some time to add the unit test I'll test and merge this PR.

accounts/views.py
```diff
@@ -1284,7 +1284,21 @@ def upload(request, no_flash=False):
             form = UploadFileForm(request.POST, request.FILES)
             if form.is_valid():
                 submitted_files = request.FILES.getlist('files')
+                if not submitted_files: 
+                    #this allows upload_tests to work, requests use the dict key in singular
```
ffont, line 1288: Hmmmm, this comment is a bit strange :S Could this be because in line 73 of test_upload.py you set "file" in singular?

ffont:
Hi! @alastair just made me realise that this PR only addressed the case in which 2 files with the same name are uploaded in a single request, but does not address the case described in the issue, which is that a sound with that filename already exists in the user upload folder but it gets overwritten by a new one. 

A way to fix that would be to add a check for duplicates which not only includes other files uploaded in the same request but also other files in user upload folder (here you can see [how this folder path is defined](https://github.com/MTG/freesound/blob/07af74b23fde0946cd5da6ad434e51cc28d5e0cc/accounts/views.py#L1202)). If there is a duplicate, then rename in a similar way as you are already doing. However, we might also simply enforce a "no duplicates" policy, let user know. If there is a filename which already exists in users' directory, then skip processing that file and don't call `handle_file_upload`, adding a new item to `errors` (see code here: https://github.com/MTG/freesound/blob/3ba8a17e4bf10a26ace445b40ea6fd28d1dd782d/accounts/views.py#L1302-L1306)

Issue #1312, opened 2019-05-14
tag browse isn't case sensitive but appears to be


alastair:
If you view a tag browse url with a lower-case tag then it will perform a search and highlight that tag in the cloud:
https://freesound.org/browse/tags/thunder/
However, if you use an upper-case tag:
https://freesound.org/browse/tags/Thunder/
then it will perform the same search (because solr is case insensitive), but it will not highlight the lowercase 'thunder' as a selected tag. These tags should be selected

Additionally, perhaps we could remove multiple instances of the same tag in a url, because the search results are the same:
https://freesound.org/browse/tags/thunder/rain/Thunder = https://freesound.org/browse/tags/thunder/rain/thunder = https://freesound.org/browse/tags/thunder/rain/ 

Be careful if some users are subscribed to the same tag more than once with or without differences in case. Make sure that they can still unsubscribe if we change the URL.

PR #1807, opened 2024-12-17
make tag browsing case-insensitive and unique

quimmrc:
**Issue(s)**
#1312 

**Description**
In the `tag` view, make the _multiple_tags_ list elements to be lowercased and unique, in order to get a proper _search_filter_ 

Note: I do not really understand why uppercase tags exist, other than by directly writing them into the url I have not found a way to use them in the Freesound website. However I guess this is still useful.


tags/views.py
```diff
@@ -34,12 +34,13 @@
...
 def tags(request, multiple_tags=None):
-
+    
     if multiple_tags:
         multiple_tags = multiple_tags.split('/')
     else:
         multiple_tags = []
-    multiple_tags = sorted([x for x in multiple_tags if x])
+    #Make all tags lower-cased and unique to get a case-insensitive search filter and shortened browse url
+    multiple_tags = sorted(set([x.lower() for x in multiple_tags if x]))
```
Bomme, line 43: No need to for the list comprehension if we turn it into a set anyways
```suggestion
    multiple_tags = sorted({x.lower() for x in multiple_tags if x})
```

search/views.py
```diff
@@ -152,7 +152,7 @@ def search_view_helper(request):
             'tags_mode': sqp.tags_mode_active(),
             'query_time': results.q_time 
         }))
-
+
```
Bomme, line 155: Nit pick: Consider setting your editor to trim all trailing whitespace automatically or use a code formatter

Bomme:
> Note: I do not really understand why uppercase tags exist, other than by directly writing them into the url I have not found a way to use them in the Freesound website. However I guess this is still useful.

I think it's still possible to add both lower-case and upper-case tags when uploading a Sound or when changing a sounds description.

quimmrc:
Solved it, though in terms of code formatting I have Prettier set up and I thought it managed this automatically already. I'll dive into this whitespace trimming

Issue #1029, opened 2017-11-17
Collections


ffont:
Freesound currently implements the concept of `Packs` which are groups of sounds by the same uploader. Packs can have a textual description, have a dedicated page and all sounds can be downloaded in a single zip. Complementarily, Freesound also implements the concept of `Bookmark categories` which are groups of sounds that can come from various uploaders. `Bookmark categories` are created by a single Freesound user and are public to all other users. Nevertheless, `Bookmark categories` can't be collaboratively created, can't have textual descriptions and the sounds can't be downloaded all at once.

We have the idea to evolve the concept of bookmarks into `Collections`. A collection would be like a `Bookmark category` but it could be created collaboratively, have a dedicated page with relevant information (description, maybe even an image) and be downloaded all at once. It would be some hybrid between packs and bookmarks categories but on steroids. This should be useful to promote specific collections of sounds, to get users to create and group sounds in Freesound and to attract new collections to be uploaded to Freesound.

After that Freesound would support `Packs` and `Collections`, but bookmark categories would disappear and be replaced by collections.

naught101:
This would be super useful. It would effectively make freesound properly curatable, which would make navigating and searching for appropriate samples a LOT easier.

Allowing collaborative collections would make for some *excellent* sample packs. Especially if there is some kind of voting mechanism as well - not just sample ratings, but "how well does this sample suit this collection" ratings.

I was also going to request some kind of downloadable pack for bookmarks, but private collections would effectively solve that problem too.

ffont:
- [ ] item 1
- [ ] item 2

ffont:
After merging #1810 we have a first implementation of that feature, although it is not yet public. We want to make more improvements before we make it public. I took some notes about things that need to be looked after when continuing working on this issue. These should be combined with the original notes of this issue, there might be some overlap:

* Optimize DB queries in `display_collection`
* Implement collections "featured sounds" so, in the collection edit page, users can select a number of sounds to be "featured"
    * There is a field already created for that in the data model, but it needs UI work
    * We would like to implement that in a way that is reusable for packs. Maybe the current data model will not allow that... Also we could decide not to make it a feature of packs as well ;)
* We should add more unit tests to collections, make sure that we are testing if a collection can't be deleted by a user that is not its creator? (or maintainer, whatever we decided the logic should be)?
[...]

PR #1810, opened 2025-01-16
Collections

quimmrc:
**Issue(s)**
https://github.com/MTG/freesound/issues/1029

**Description**
App creation and first implementations of basic functionalitites for collections.


fscollections/models.py
```diff
@@ -0,0 +1,46 @@
...
+
+from django.contrib.auth.models import User
+from django.db import models
+
+from sounds.models import Sound, License
+
+class Collection(models.Model):
+
+    author = models.ForeignKey(User, on_delete=models.CASCADE) 
+    name = models.CharField(max_length=128, default="BookmarkCollection") #add restrictions
```
ffont, line 29: No need to specify a default

freesound/settings.py
```diff
@@ -927,6 +928,8 @@
 # -------------------------------------------------------------------------------
 # Extra Freesound settings
```
ffont, line 930: Maybe rename this to `ENABLE COLLECTIONS`?

fscollections/models.py
```diff
@@ -0,0 +1,60 @@
...
+from django.contrib.auth.models import User
+from django.db import models
+
+from sounds.models import Sound, License
+
+class Collection(models.Model):
+
+    user = models.ForeignKey(User, on_delete=models.CASCADE) 
+    name = models.CharField(max_length=255) #max_length as in Packs (128 for Bookmarks)
+    created = models.DateTimeField(db_index=True, auto_now_add=True)
```
ffont, line 30: add `modified` field

fscollections/models.py
```diff
@@ -0,0 +1,60 @@
...
+class CollectionSound(models.Model):
+   #this model relates collections and sounds
+   #it might be worth adding a name field composed of the sound ID and the collection name for
+   # for the sake of queries understanding
+   user = models.ForeignKey(User, on_delete=models.CASCADE) #not sure bout this
+   sound = models.ForeignKey(Sound, on_delete=models.CASCADE)
+   collection = models.ForeignKey(Collection, related_name='collectionsound', on_delete=models.CASCADE)
+   created = models.DateTimeField(db_index=True, auto_now_add=True)
+   
+   STATUS_CHOICES = (
```
ffont, line 54: Options:

- Accepted
- Pending
- Refused

fscollections/views.py
```diff
@@ -0,0 +1,226 @@
...
+             'form': form,
+             'collections_with_sound': collections_already_containing_sound
+             }
+    
+    return render(request, 'collections/modal_collect_sound.html', tvars)
+
+def delete_collection(request, collection_id):
+    collection = get_object_or_404(Collection, id=collection_id)
+
+    if request.user==collection.user:
```
ffont, line 138: Consider else case, return 404 or reload edit page but with a message saying that user is not allowed

freesound/static/bw-frontend/src/components/addSoundsModal.js
```diff
@@ -73,11 +88,14 @@ const prepareAddSoundsModalAndFields = (container) => {
                 const newSoundIds = serializedIdListToIntList(selectedSoundIds);
                 const combinedIds = combineIdsLists(currentSoundIds, newSoundIds);
                 selectedSoundsHiddenInput.value = combinedIds.join(',');
+                if(selectedSoundsHiddenInput.value.split(',').length >= 4 && selectedSoundsHiddenInput.id === "collection_sounds"){
```
quimmrc, line 91: there's an error here, this 4 should take the value of settings.MAX_SOUNDS_PER_COLLECTION

freesound/static/bw-frontend/src/components/collections.js
```diff
@@ -0,0 +1,170 @@
...
+        if (element.dataset.alreadyBinded !== undefined){
+            return;
+        }
+        element.dataset.alreadyBinded = true;
+        element.addEventListener('click', (evt) => {
+            evt.preventDefault();   
+            const modalUrlSplitted = element.dataset.modalContentUrl.split('/')
+            const soundId = parseInt(modalUrlSplitted[modalUrlSplitted.length - 3], 10)
+            if (!evt.altKey) {
+                handleGenericModalWithForm(element.dataset.modalContentUrl, () => {
```
quimmrc, line 57: in here, bookmarks used a custom saveCollection function, however I found more intuitive and standard to use a handleGenericModalWithForm to properly handle the submission of the SelectCollectionOrNewCollection form, which is the one used to add sounds to Collections from Sound URLS.

freesound/static/bw-frontend/src/components/collections.js
```diff
@@ -0,0 +1,170 @@
...
+        /*
+        USEFUL NOTES AFTER DEBUGGING:
+        
+        The function updateObjectSelectorDataProperties in charge of defining the selectedIds and unselectedIds for each
+        selector does not work for maintainers. Instead it is only triggered once the user interacts with a maintainer checkbox.
+        Data might get overwritten as this works simultaneously with addSoundsModal.js file. The thing is that in
+        updateObjectSelectorDataProperties, the "call stack" states that originally, the function is called from here (line 175).
+        Therefore, the definition of selectedMaintainersDestinationElement somehow is wrong or confused with the selectedSoundsDestinationElement.
+        I tried to filter this using the above queryselector but it does not work. I guess this might have something to do with the file
+        collectionEdit.js.
```
quimmrc, line 123: This piece of comment might be interesting but the issue prompted by this is already solved. Basically the query for users does not exclude any, and instead "validation" for users who are added as maintaienrs is done from server-side.

fscollections/forms.py
```diff
@@ -0,0 +1,358 @@
...
+            'description': Textarea(attrs={'rows': 5, 'cols': 50}),
+            'public': forms.RadioSelect(choices=[(True, 'Public'), (False, 'Private')], attrs={'class': 'bw-radio'})
+        }
+
+class CollectionEditFormAsMaintainer(CollectionEditForm):
+    class Meta(CollectionEditForm.Meta):
+        fields = CollectionEditForm.Meta.fields + ['collection_sounds'] + ['maintainers']
+    
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
quimmrc, line 289: this might need a good review, it was a topic of discussion in our last meeting and I think that my proposal differs a bit from your main idea. In here this forms inherits everything so that it can be displayed in the edit url but all fields get disabled as soon as the form is initalized.

fscollections/models.py
```diff
@@ -0,0 +1,144 @@
...
+        ("PE", 'Pending'),
+        ("OK", 'Accepted'),
+        ("RE", 'Refused'),
+    )
+   status = models.CharField(db_index=True, max_length=2, choices=STATUS_CHOICES, default="PE")
+   #sound won't be added to collection until maintainers approve the sound
+
+@receiver(post_save, sender=CollectionSound)
```
quimmrc, line 108: num_sounds are now updated using CollectionSounds triggers, not only in the save collection method

fscollections/views.py
```diff
@@ -0,0 +1,277 @@
...
+    user_collections = Collection.objects.filter(user=user).order_by('-modified')
+    maintainer_collections = Collection.objects.filter(maintainers__id=user.id).order_by('-modified')
+    tvars = {'user_collections': user_collections,
+             'maintainer_collections': maintainer_collections
+             }
+    # one URL needed to display all collections and one URL to display ONE collection at a time
+    # the collections_for_user can be reused to display ONE collection so give it a thought on full collections display
+    return render(request, 'collections/your_collections.html', tvars)
+
+def collection_stats_section(request, collection_id):
```
quimmrc, line 83: In here I tried to imitate the pack stats behaviour but I was not much familiar with it and how the cache is handled so this is not properly updated (needs further review by me)

fscollections/views.py
```diff
@@ -0,0 +1,277 @@
...
+            form = CollectionEditForm(request.POST, instance=collection, label_suffix='', is_owner=is_owner, is_maintainer=is_maintainer)
+        elif is_maintainer:
+            form = CollectionEditFormAsMaintainer(request.POST, instance=collection, label_suffix='', is_owner=is_owner, is_maintainer=is_maintainer)
+        else:
+            return HttpResponseRedirect(reverse('collection',collection_id))
+        
+        if form.is_valid():
+            form.save(user_adding_sound=request.user)
+            return HttpResponseRedirect(reverse('collection', args=[collection.id]))
+        else:
```
quimmrc, line 181: this part might need a good review since I did something unusual here, but I still believe it's solid enough

templates/collections/modal_collect_sound.html
```diff
@@ -0,0 +1,56 @@
+{% extends "molecules/modal_base.html" %}
+{% load static %}
+{% load bw_templatetags %}
+
+{% block id %}collectSoundModal{% endblock %}
```
quimmrc, line 5: This should be named "AddSoundToCollectionModal" or similar, I forgot to change this, my bad

freesound/static/bw-frontend/src/components/addSoundsModal.js
quimmrc, line 1: this allows to differentiate the sound selector in the edit collections page from the users selector (for maintainers)

fscollections/migrations/0011_collection_sounds.py
```diff
@@ -0,0 +1,19 @@
...
+        ('sounds', '0055_remove_sound_geotag'),
+        ('fscollections', '0010_remove_collection_max_num_sounds'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='collection',
+            name='sounds',
+            field=models.ManyToManyField(blank=True, related_name='collections', through='fscollections.CollectionSound', to='sounds.sound'),
+        ),
```
ffont, line 18: I think we should probably consolidate all these migrations into a single one. I can try and do that when testing this PR.

fscollections/templatetags/display_collections.py
```diff
@@ -0,0 +1,42 @@
...
+
+register = template.Library()
+
+
+@register.inclusion_tag('collections/display_collection.html', takes_context=True)
+def display_collection(context, collection_id):
+    collection = get_object_or_404(Collection, id=collection_id)
+    request = context.get('request')
+    try: 
+        sound = Sound.objects.get(collections=collection, collectionsound__featured_sound=True)
```
ffont, line 36: This is fine for now, but we'll need to optimise number of DB queries, etc, when displaying collections. We'll have to discuss with alastair the best way to do that, but let's wait until the rest of the first prototype is working as expected.

fscollections/tests.py
```diff
@@ -0,0 +1,244 @@
...
+        # User logged in, check given collection id works
+        resp = self.client.get(reverse('collection', args=[self.collection.id]))
+        self.assertEqual(resp.status_code, 200)
+
+        # Create collection view    
+        resp = self.client.post(reverse('create-collection')+'?ajax=1', {'name': 'tbdcollection'})
+        self.assertEqual(resp.status_code, 200) 
+        delete_collection = Collection.objects.get(name='tbdcollection')
+
+        # Delete collection view
```
ffont, line 48: Are we also testing if a collection can't be deleted by a user that is not its creator? (or maintainer, whatever we decided the logic should be)

fscollections/views.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+@login_required
+def collection(request, collection_id):
+    user = request.user
+    is_owner = False
+    is_maintainer = False
+    maintainers = []
+    # if no collection id is provided for this URL, render the oldest collection
+    # only show the collections for which you're the user(owner)
```
ffont, line 45: I guess these 2 comments are old/wrong?

fscollections/views.py
```diff
@@ -0,0 +1,279 @@
...
+    if not collection_id:
+        # this could probably be a JsonResponse message indicating an error
+        return HttpResponseRedirect(reverse('your-collections'))
```
ffont, line 48: Why this redirect? I think we should make `collection_id` mandatory and that's it. In fact, isn't that already mandatory? I think this `if` is not needed?

fscollections/views.py
ffont: I'll have a look when testing the PR. This async loading of stats has to do with caching computation of stats, but maybe if stats computation is fast enough, then this might not be needed. We added some recent changes to compute some of the stats in DB rather than in python with good results, so we might apply this here and not do the stats section async.

fscollections/views.py
```diff
@@ -0,0 +1,279 @@
...
+    else:
+        form = CreateCollectionForm(user=request.user)
+    tvars = {'form': form}
+    return render(request, 'collections/modal_create_collection.html', tvars)
+    
+def delete_collection(request, collection_id):
+    collection = get_object_or_404(Collection, id=collection_id)
+
+    if request.method=="POST" and request.user==collection.user:
+        collection.delete()
```
ffont, line 152: Add a message like "collection successfully deleted"? See how django messages app works, we do it in many other places

fscollections/views.py
```diff
@@ -0,0 +1,279 @@
...
+        return HttpResponseRedirect(reverse('your-collections'))
+    else:
+        return HttpResponseRedirect(reverse('collection', args=[collection.id]))
```
ffont, line 155: Maybe we should return "not allowed" response. Did you check how similar things are handled elsewhere in Freesound?

fscollections/views.py
ffont: I'll have a detailed look into this when I can test the PR. At first sight I don't think we need to anything unusual here.

sounds/models.py
```diff
@@ -456,6 +456,18 @@ def bulk_sounds_for_pack(self, pack_id, limit=None, include_analyzers_output=Fal
             qs = qs[:limit]
         return qs
 
+    def bulk_sounds_for_collection(self, collection_id, limit=None, include_analyzers_output=False):
+        qs = self.bulk_query(include_analyzers_output=include_analyzers_output)
+        qs = qs.filter(
+            moderation_state="OK",
+            processing_state="OK",
+            collections=collection_id
```
ffont, line 464: Does this work?? I guess it is some feature of the `through` model, but I'm a bit surprised here... I'd like to look into this with @alastair

freesound/context_processor.py
```diff
@@ -76,7 +76,9 @@ def context_extra(request):
             'next_path': request.GET.get('next', request.get_full_path()),
             'login_form': FsAuthenticationForm(),
             'problems_logging_in_form': ProblemsLoggingInForm(),
-            'system_prefers_dark_theme': request.COOKIES.get('systemPrefersDarkTheme', 'no') == 'yes'  # Determine the user's system preference for dark/light theme (for non authenticated users, always use light theme)
+            'system_prefers_dark_theme': request.COOKIES.get('systemPrefersDarkTheme', 'no') == 'yes',  # Determine the user's system preference for dark/light theme (for non authenticated users, always use light theme)
+            'enable_collections': settings.ENABLE_COLLECTIONS,
+            'max_sounds_per_collection': settings.MAX_SOUNDS_PER_COLLECTION,
```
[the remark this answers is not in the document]
quimmrc: I agree for `max_sounds_per_collection`, however, for the enable_collections case it is used in a couple of html templates which are the navigation bar and the sound url ones, which have a more global behavior rather than collections-related-pages, do you think we could leave `enable_collections` as a context processor?

freesound/static/bw-frontend/src/components/addSoundsModal.js
```diff
@@ -92,9 +98,12 @@ const prepareAddSoundsModalAndFields = (container) => {
                 const newSoundIds = serializedIdListToIntList(selectedSoundIds);
                 const combinedIds = combineIdsLists(currentSoundIds, newSoundIds);
                 selectedSoundsHiddenInput.value = combinedIds.join(',');
-                if(selectedSoundsHiddenInput.value.split(',').length >= 4 && selectedSoundsHiddenInput.id === "collection_sounds"){
+                if(maxSounds !== "None" && selectedSoundsHiddenInput.value.split(',').length >= maxSounds){
                     addSoundsButton.disabled = true;
+                    maxSoundsHelpText.style.display = 'block';
                 }
+                if (soundsLabel){
+                    soundsLabel.innerHTML = "Sounds in collection (" + selectedSoundsDestinationElement.children.length + ")"}
```
quimmrc, line 106: A part from removing the hard-coded part and retrieving the number of elements from the sound selector, I automated the display of the help text for this field (informing of maximum number of sounds permitted) and the label of this field (which shows the number of sounds in collection).

sounds/templatetags/display_sound.py
```diff
@@ -148,35 +149,35 @@ def display_sound_small_with_timesince(context, sound):
...
+    return display_sound(context, sound, player_size='big_no_info', show_collections=False, show_bookmark=False, show_similar_sounds=False, show_remix=False)
 
 @register.inclusion_tag('sounds/display_sound.html', takes_context=True)
 def display_sound_small_no_info(context, sound):
     return display_sound(context, sound, player_size='small_no_info', show_rate_widget=True)
 
 @register.inclusion_tag('sounds/display_sound.html', takes_context=True)
 def display_sound_small_no_info_no_buttons(context, sound):
-    return display_sound(context, sound, player_size='small_no_info', show_rate_widget=False, show_bookmark=False, show_similar_sounds=False, show_remix=False)
+    return display_sound(context, sound, player_size='small_no_info', show_rate_widget=False, show_collections=False, show_bookmark=False, show_similar_sounds=False, show_remix=False)
```
alastair, line 180: its not directly related to this PR, but I see that `show_rate_widget` comes _after_ the bookmark/collections params. It'd be good for these named parameters to be in the same order as they are defined in the method

templates/accounts/display_user_selectable.html
```diff
@@ -0,0 +1,10 @@
+<div class="bw-selectable-object between {% if selected %}selected{% endif %}">
+    <div class="middle spacing-2">
+        <label>
+        <input type="checkbox" class="bw-checkbox" data-object-id="{{ user.id }}" {% if selected %}checked{% endif %}>
+        </label>
```
alastair, line 5: can you select this input by clicking on the display_user? If not, then maybe we don't need the label

templates/collections/collection.html
```diff
@@ -0,0 +1,71 @@
...
+        </div>    
+        <div class="col-lg-8">
+            {% if collection %}
+            <div class="middle bw-sound-page__user v-spacing-top-3">
+                <div class="h-spacing-1 ellipsis">
+                    {% bw_user_avatar collection.user.profile.locations.avatar.M.url collection.user.username 40 %}
+                </div>
+                <div class="h-spacing-left-1">
+                    <a href="{% url 'account' collection.user.username %}">{{ collection.user.username | truncate_string:15 }}</a>
+                    <p class="text-grey">{{ collection.created|date:"F jS, Y" }}</p>
```
alastair, line 34: not part of this PR, but I see that this `date` format is used in lots of places in freesound. would be great to make our own custom filter that has the format built-in

templates/collections/collection.html
```diff
@@ -0,0 +1,71 @@
...
+            <div>{% if collection.public %} {% bw_icon 'eye' %} Public {% else %} {% bw_icon 'eye-blocked' %} Private {% endif %}collection</div>
+            <span>Maintainers: 
+                {% if maintainers %}{% for u in maintainers %}
+                    {% if not forloop.last%}
+                        <a href="{% url 'account' u.username%}">{{u.username}}</a>, 
+                    {% else %} 
+                    <a href="{% url 'account' u.username%}">{{u.username}}</a>
+                    {% endif %}
+                {% endfor %}
+                {% else %}There are no maintainers in this collection yet.{% endif %}
```
alastair, line 47: personally I'd skip the maintainers section if there are no maintainers, especially in the public view

templates/collections/collection.html
```diff
@@ -0,0 +1,71 @@
...
+                    {% endfor %}
+                </div>
+            {% else %} 
+            There aren't any sounds in this collection yet 
+            {% endif %}
+            {% bw_paginator paginator page current_page request "collection" %}
+            </div>
+            {% else %}
+            <div class="v-spacing-top-3">
+                <span class=text-light-grey> You don't have any collection yet...</span>
```
alastair, line 65:
```suggestion
                <span class=text-light-grey> You don't have any collections yet...</span>
```

templates/collections/collection_stats_section.html
```diff
@@ -0,0 +1,18 @@
+{% load bw_templatetags %}
+{% load util %}
+{% load cache %}
+<!--{% cache 3600 bw_collection_stats collection.id %}-->
```
alastair, line 4: double-check - not sure this comment actually does anything in the context of django caches

templates/collections/modal_add_sound_to_collection.html
```diff
@@ -0,0 +1,56 @@
...
+            <div class="v-spacing-2">
+            {% bw_icon 'atom' %}The following collections have reached the maximum number of sounds allowed ({{max_sounds_per_collection}}):
+            {% for col in full_collections %}
+            <span class="text-black font-weight-bold">{{col.name}}</span>{% if not forloop.last%},{% endif %}
+            {% endfor %}
+            </div>
+        {% endif %}
+        <form method="post" action="{% url 'add-sound-to-collection' sound.id %}?ajax=1." class = "disable-on-submit">{% csrf_token %}
+            {{ form }}
+            <button class = "btn-primary v-spacing-top-2">Collect Sound</button>
```
alastair, line 50: it's not normal to use "Collect" as a verb in this context. You'll need to use "Add sound to collection"

templates/collections/your_collections.html
```diff
@@ -0,0 +1,73 @@
...
+<div class="v-spacing-top-5">
+    <div class="row no-gutters">
+        <div class="col-lg-2 offset-lg-1 order-lg-last v-spacing-3">
+            <h4>Your collections</h4>
+            <div class="v-spacing-top-3">
+                <ul class="list-style-type-none">
+                    <li><a class="cursor-pointer" data-toggle="modal-default-with-form" data-modal-content-url="{% url 'create-collection' %}?ajax=1" data-reload-on-success= "true" title="Create new collection" aria-label="Create new collection">
+                        Create a new collection <span style="margin-left: 5px">{% bw_icon 'bookmark' %}</span></a></li>
+                    {% if user_collections %}
+                        <li><span> Collections you own: {{user_collections|length}} </span></li>
```
alastair, line 20:
```suggestion
                        <li><span> Your Collections: {{user_collections|length}} </span></li>
```

templates/collections/your_collections.html
```diff
@@ -0,0 +1,73 @@
...
+                        {% for collection in user_collections %}
+                            <div class="col-md-6 col-lg-4 v-spacing-top-3">
+                                {% display_collection collection.id %}
+                            </div>
+                        {% endfor %}
+                    </div>
+                </div>
+                {% else %}
+                <div class="v-spacing-top-3">
+                    <span class=text-light-grey> You don't have any collection yet...</span>
```
alastair, line 45:
```suggestion
                    <span class=text-light-grey> You don't have any collections yet...</span>
```

templates/collections/your_collections.html
```diff
@@ -0,0 +1,73 @@
...
+                        {% for collection in maintainer_collections %}
+                            <div class="col-md-6 col-lg-4 v-spacing-top-3">
+                                {% display_collection collection.id %}
+                            </div>
+                        {% endfor %}
+                    </div>
+                </div>
+                {% else %}
+                <div class="v-spacing-top-3">
+                    <span class=text-light-grey> You're not a maintainer in any collection yet...</span>
```
alastair, line 67:
```suggestion
                    <span class=text-light-grey> You're not a maintainer of any collection yet...</span>
```

fscollections/models.py
```diff
@@ -0,0 +1,144 @@
...
+def update_collection_num_sounds(**kwargs):
```
alastair, line 109: It'd be better to have the explicit list of parameters to these signals: https://docs.djangoproject.com/en/5.2/ref/signals/#django.db.models.signals.pre_save
```suggestion
def update_collection_num_sounds(sender, instance, raw, using, update_fields):
```

Then you don't need to set isntance to False in the next line (this will already be None if you're in this situation)

fscollections/models.py
```diff
@@ -0,0 +1,144 @@
...
+    collectionsound = kwargs.pop('instance', False)
+    if collectionsound:
+        Collection.objects.filter(collectionsound=collectionsound).update(num_sounds=Greatest(F('num_sounds') + 1, 0))
+
+@receiver(post_delete, sender=CollectionSound)
+def update_collection_num_sounds_sound_removal(**kwargs):
```
alastair, line 115: Note that django doesn't send post_save/post_delete signals when you directly update a ManyToMany field: see https://docs.djangoproject.com/en/4.2/ref/signals/#m2m-changed
I see that you create these objects manually anyway (this is necessary if you have a through model with extra data, so this is fine), but we may be in a situation where we bulk remove sounds from a collection and this signal isn't triggered. Review if this might happen in any place

fscollections/forms.py
```diff
@@ -0,0 +1,357 @@
...
+        for snd in new_sounds:
+            if snd not in current_sounds:
+                sound = Sound.objects.get(id=snd)
+                CollectionSound.objects.create(user=user_adding_sound, sound=sound, collection=collection, status='OK')
+              
+            else:
+                current_sounds.remove(snd)
+        for snd in current_sounds:
+            sound = Sound.objects.get(id=snd)
+            CollectionSound.objects.get(collection=collection, sound=sound).delete()
```
alastair, line 255: this makes 3 queries per item to delete. better to do either:
```py
sounds = Sound.objects.filter(id__in=current_sounds)
collection.sounds.remove(*sounds)
```
(but careful with this option, because it won't trigger a delete signal, because it modifies the m2m field)

or maybe
```py
sounds = Sound.objects.filter(id__in=current_sounds)
CollectionSound.objects.filter(collection=collection, sound__in=sounds).delete()
```
because the filter/delete will do _only_ a delete query and no SELECT.

fscollections/models.py
```diff
@@ -0,0 +1,144 @@
...
+class Collection(models.Model):
+
+    user = models.ForeignKey(User, on_delete=models.CASCADE) 
+    name = models.CharField(max_length=255) #max_length as in Packs (128 for Bookmarks)
+    created = models.DateTimeField(db_index=True, auto_now_add=True)
+    modified = models.DateTimeField(db_index=True, auto_now=True)
+    # TODO: description should be required (check how to display it in edit form + collectionsound form)
+    description = models.TextField(blank=True)
+    maintainers = models.ManyToManyField(User, related_name="collection_maintainer", blank=True)
+    sounds = models.ManyToManyField(Sound, through="CollectionSound", related_name='collections', blank=True)
```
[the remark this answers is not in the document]
quimmrc: I think it is useful for form validation purposes (e.g.: removing all maintainers from a collection -> empty field). I took the idea from the [django references](https://docs.djangoproject.com/en/3.2/ref/models/fields/#blank), does this add up?

fscollections/tests.py
```diff
@@ -0,0 +1,254 @@
...
+        maintainer_test_collection = Collection.objects.create(name="maintainer_test_collection", user=self.user)
+        form_data = {'collection': maintainer_test_collection.id, 'new_collection_name': '', 'use_last_collection': False}
+        resp = self.client.post(reverse('add-sound-to-collection', args=[self.sound.id]) + '?ajax=1', form_data)
+        self.assertEqual(200, resp.status_code)
+        maintainer_test_collection.refresh_from_db()
+        self.assertEqual(0, maintainer_test_collection.num_sounds)
+
+        # Test deleting a collection where request user is maintainer (not valid)
+        resp = self.client.post(reverse('delete-collection',args=[self.collection.id]))
+        self.assertEqual(resp.status_code, 302)
```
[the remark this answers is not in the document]
quimmrc: When the user is invalid to delete a collection (is either a maintainer or an external user) is redirected to the collection url and a message is displayed informing the user about the lack of permission to delete the collection (hence a 302), what do you think of this proposal? Regarding the 200 for the add sound I think I might have made a mistake there, I'll further look into it.

fscollections/urls.py
```diff
@@ -0,0 +1,16 @@
+from django.urls import path
+from . import views
+
+urlpatterns = [
+    path('',views.collections_for_user, name='your-collections'),
+    path('<int:collection_id>/', views.collection, name='collection'),
```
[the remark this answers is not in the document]
quimmrc: This is something I'd like to dig into because even though the "shared" approach collections have, I think it would be interesting (or at least it would follow the common methodology across the freesound software). What do you think about this?

fscollections/urls.py
```diff
@@ -0,0 +1,16 @@
...
+    path('<int:sound_id>/add/', views.add_sound_to_collection, name='add-sound-to-collection'),
+    path('create/', views.create_collection, name='create-collection'),
+    path('<int:collection_id>/edit', views.edit_collection, name="edit-collection"),
```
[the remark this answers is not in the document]
quimmrc: Following my last comment, putting the collection name as a slug in the url could be a good fit. I like it better than putting the username in the url, so I'll check how to do that.

fscollections/migrations/0001_initial.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 3.2.23 on 2025-01-07 12:44
```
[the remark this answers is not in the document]
alastair: easiest way is to just delete them all and then run makemigrations again.

You'll need to undo all of your migrations locally so that you can apply it again (with `manage.py migrate fscollections zero`), or delete your database and reimport

fscollections/tests.py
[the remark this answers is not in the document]
alastair: You can check messages in a test by using `resp.context["messages"]`: https://docs.djangoproject.com/en/5.2/topics/testing/tools/#testing-responses

fscollections/urls.py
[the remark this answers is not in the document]
alastair: I also discussed this with Frederic and he confirmed. Don't worry about it for now

templates/collections/collection_stats_section.html
quimmrc: As of now the collections stats section is not computed as in Packs because of my lack of knowledge on how to handle caches, that's why I guess I commented that part.

fscollections/urls.py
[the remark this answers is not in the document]
ffont: I'll add this as a TODO item, but leave it as it is for the current merge

PR #1831, opened 2025-01-31
Upgrade redis and set specific version

### LINKED ISSUE #1745: Pin redis image version in docker compose file
Currently we use the `redis:alpine` image tag which might resolve to a different image (with a "new" redis version on pull)
We should instead use a `redis:<version>-alpine` tag.
See the [list of supported versions/tags](https://hub.docker.com/_/redis/)

alastair:
**Issue(s)**
Fixes #1745

**Description**
Different versions of the redis-py library support different versions of redis. Use an explicit version in docker-compose.yml and update python dependency to a version which supports it.

**Deployment steps**:
Pending freesound-deploy PR to do the same thing on prod

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -61,7 +61,7 @@ services:
 
     # Redis (caches)
     redis:
-        image: redis:alpine
+        image: redis:7.4.2
         command: redis-server --port 6379
 
     # Celery broker (queue manager)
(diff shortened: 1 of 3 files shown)


ffont:
I'll merge as the redis client should also support the redis version we use in prod. Then we could upgrade redis in prod as a separate step.

PR #1844, opened 2025-02-13
pyupgrade & django-upgrade changes

Bomme:
I ran [`pyupgrade`](https://github.com/asottile/pyupgrade) and [`django-upgrade`](https://github.com/adamchainz/django-upgrade) after upgrading to Django 4.2
* `django-upgrade --target-version 4.2`
* `pyupgrade --py310-plus`


alastair (approved):
looks good to me!

PR #1848, opened 2025-02-17
Clean up pack objects manager usage

alastair:
Because we always use a filter on the queryset, this view will never raise DoesNotExist, so remove it
We never filter deleted packs on the view page, so the deleted templated doesn't make sense


alastair:
todo: test to ensure that pack with an id that doesn't exist and a deleted pack correctly give a 404 error

Issue #1840, opened 2025-02-12
Upgrade to Django 5.2


ffont:
We should upgrade to the next LTS version of Django (5.2)

Steps:

- [ ] Clear Django 5 warnings (run tests with warnings option enabled to see them). This will require at least upgrading `SESSION_SERIALIZER` to `django.contrib.sessions.serializers.JSONSerializer`, which causes problems because we sometimes serialize `Sound` and `File` objects in the session. We need to reimplement the way we do that (or extend the serializer so it can serialize/unserialize these data types?)

- [ ] Make sure Django dependencies are fine:

* dj-database-url > should be fine with 5.x
* django-admin-sortable > unsure about 5.x. We could look at django-admin-sortable2 ifneedbe.
* django-amazon-ses > unsure about 5.x
* django-cors-headers >  should be fine with 5.x
* django-debug-toolbar >  should be fine with 5.x
* django-extensions >  should be fine with 5.x
* django-multiupload >  should be fine with 5.x
* django-oauth-toolkit >  should be fine with 5.x
* django-object-actions > no notes about django compatibility, but latest version works with 4.2
* django-ratelimit > no notes about django compatibility, but latest version works with 4.2
* django-recaptcha >  unsure about 5.x
* django-redis >  unsure about 5.x
* django-silk > should be fine with 5.x
* djangorestframework > should be fine with 5.x
* sentry-sdk[django] > should be fine with 5.x

- [ ] Update default checkbox render behaviour so that checkbox INPUT elements are rendered inside LABEL elements. Then we can remove the ugly hack in `checkbox.js` which moves checkbox elements inside labels. In Django 5.x, this change should be easily achieved, see this: https://docs.djangoproject.com/en/5.1/topics/forms/#reusable-field-group-templates

Potential follow-up steps:

* replace django-redis with [django.core.cache.backends.redis.RedisCache](https://docs.djangoproject.com/en/4.2/releases/4.0/#redis-cache-backend)? See https://github.com/jazzband/django-redis/issues/585 for a discussion
* Switch to Psycopg 3 https://docs.djangoproject.com/en/4.2/releases/4.2/#psycopg-3-support

PR #1857, opened 2025-02-21
Switch from psycopg2 to psycopg3

alastair:
**Issue(s)**
Part of https://github.com/MTG/freesound/issues/1840

**Description**
According to django and psycopg release notes, this shouldn't have a major effect. We have a few manually written queries, so let's see if it causes any issues when running tests


ffont:
This is ready now right?

alastair:
yes, from what I understand this should be a drop-in replacement.
There was the one query change that I had to make in the bulk query method - this is documented as a limitation in the new library which needs to be changed. The behaviour is the same. I can't find any other methods in FS that used this, and all tests pass: https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html#you-cannot-use-in-s-with-a-tuple

Issue #1818, opened 2025-01-30
Show user/pack sound stats in minutes if total duration is less than an hour


alastair:
![Image](https://github.com/user-attachments/assets/4e1c7bdd-f7b8-412c-8aca-7fd51b84a254)

![Image](https://github.com/user-attachments/assets/857218fc-ba90-447b-bea9-1074cdc86ab1)

For packs or users that have a short total duration of sounds, the stats don't really show a lot of useful information. In the case that the total sum is less than about an hour, we could show minutes instead

PR #1861, opened 2025-02-21
Show user and pack durations as minutes if less than an hour 

alastair:
**Issue(s)**
Fixes https://github.com/MTG/freesound/issues/1818

**Description**
'0:03 hours' isn't very useful, better to show '3:45 minutes'
Screenshot 2025-02-21 at 9 23 15 PM

The code for computing/selecting hours vs minutes is duplicated in two places, for now I'm not worried about this but we could combine it in a helper if needed. The `|duration_hours` templatetag is still used elsewhere, so don't remove it.

I updated the test settings to disable the cache, so that I could read values for the test multiple times. I think it generally makes sense to disable the cache during tests? although some tests are now failing.
If we want to keep it then I can work around it for just this test.

### DIFF
--- a/accounts/tests/test_profile.py
+++ b/accounts/tests/test_profile.py
@@ -32,6 +32,7 @@ from django.core.management import call_command
 from django.test import TestCase
 from django.urls import reverse
 from django.utils.http import int_to_base36
+import pytest
 
 import accounts.models
 from accounts.management.commands.process_email_bounces import process_message, decode_idna_email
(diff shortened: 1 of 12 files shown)


ffont:
Maybe we should re-write that into a template filter and then have a test for the filter function only so we don't depend on the cache?

Cache is tested in different tests, I guess it makes sense to have it. But you should also be able to disable cache for a specific test using `@override_settings`? In any case I think it might be better to not test the template but some util function.

alastair:
Updated to use a templatetag that shows both the time and units, added a specific test for the templatetag and also general tests for the section templates.
I'm using some new pytest functionality for modifying settings, take a look!

ffont:
Thanks @alastair!
I noticed the `@pytest.mark.django_db` thingy. I guess this is useful to group tests for running them?
Also the fixture for disabling caches, which is passes as an argument in the test function.

I'll merge that one!

PR #1864, opened 2025-03-03
wip: use ORM for bulk_query methods

alastair:
ORMs can do subqueries since 2017, which is like 2 years after we wrote the first version of this method. lol.

**Issue(s)**

**Description**

**Deployment steps**:


apiv2/serializers.py
```diff
@@ -46,23 +46,12 @@
 DEFAULT_FIELDS_IN_PACK_DETAIL = None  # Separated by commas (None = all)
 
 
-def get_sound_analyzers_output_helper(sound, fallback_to_db=True):
+def get_sound_analyzers_output_helper(sound):
     analyzers_output = {}
+    sound_analysis_dict = {an.analyzer: an.analysis_data for an in sound.analyses.all()}
```
alastair, line 51: at this point in time, `sound.analyses.all()` has been preloaded. We don't want to do `sound.analyses.filter(xxxx)` because that will trigger a new DB request. Instead turn it into a dict based on key and then get these key items based on the settings value.

apiv2/serializers.py
```diff
@@ -46,23 +46,12 @@
...
-            analysis_data = None
-            if hasattr(sound, query_select_name):
-                analysis_data = getattr(sound, query_select_name)
-                if type(analysis_data) == str:
-                    analysis_data = json.loads(analysis_data)
-            else:
-                if fallback_to_db:
-                    # Retrieve the analysis data from the db
-                    try:
-                        analysis_data = sound.analyses.get(analyzer=analyzer_name, analysis_status="OK").analysis_data
```
alastair, line 63: If there are no items in `sound.analyses.all()` then we'll just get `None` from the dictionary. This is different to the previous bulk method which had an attr on the sound which could have been None. In this case it doesn't make sense to get the item from the database "again". 
If we didn't do a prefetch, then the above call to `sound.analyses.all()` will get them anyway.

There's a slight difference here where we we might get items with a non-OK status. We could fix this by modifying the dict comprehension

apiv2/tests.py
```diff
@@ -362,21 +362,22 @@ class TestSoundSerializer(TestCase):
...
         ContentType.objects.get_for_model(Sound)
         Site.objects.get_current()
 
-        # Test that the serialized sound instance includes all fields in the serializer and does not perform any
-        # extra query. Because in this test we get sound info using Sound.objects.bulk_query_id, the serializer
-        # should perform no extra queries to render the data
-        with self.assertNumQueries(0):
-            dummy_request = self.factory.get(reverse('apiv2-sound-instance', args=[self.sound.id]) + '?fields=*')
-            serialized_sound = SoundSerializer(self.sound, context={'request': dummy_request}).data
+        with self.assertNumQueries(2):
```
alastair, line 374: for some reason I was unable to force this to execute both queries (the sound get and the prefetch) outside of this `with` statement. so just move the query inside it and expect to find both of them

apiv2/views.py
```diff
@@ -388,6 +388,9 @@ def post(self, request,  *args, **kwargs):
 # SOUND VIEWS
 #############
 
+def get_needs_analyzers_output(fields):
+    return 'analyzers_output' in fields or 'ac_analysis' in fields or '*' in fields
```
alastair, line 392: this logic was used in many places but missed `'*'` from the checks. Move it to a helper function

apiv2/views.py
```diff
@@ -348,8 +348,8 @@ def get(self, request,  *args, **kwargs):
         ids = results
         sound_analysis_data = get_analysis_data_for_sound_ids(request, sound_ids=ids)
         # In search queries, only include audio analyers's output if requested through the fields parameter
-        needs_analyzers_ouptut = 'analyzers_output' in search_form.cleaned_data.get('fields', '') or 'ac_analysis' in search_form.cleaned_data.get('fields', '')
```
alastair, line 351: spelling mistake in the variable name! At least it was consistent throughout the codebase...

sounds/models.py
```diff
@@ -478,106 +468,86 @@ def bulk_query_solr(self, sound_ids):
...
+                tag_array=ArraySubquery(tags_subquery),
+                analysis_state_essentia_exists=Exists(analysis_subquery),
+                search_engine_similarity_state=Exists(search_engine_similarity_subquery)
+            )
+
+        if include_analyzers_output:
+            analyzer_names = list(settings.ANALYZERS_CONFIGURATION.keys())
+            print(analyzer_names)
+            qs = qs.prefetch_related(
+                Prefetch('analyses', queryset=SoundAnalysis.objects.filter(analyzer__in=analyzer_names))
```
alastair, line 503: TODO: We can filter by status="OK" here

sounds/models.py
```diff
@@ -478,106 +468,86 @@ def bulk_query_solr(self, sound_ids):
...
+            sound=OuterRef("id"), analyzer=settings.FREESOUND_ESSENTIA_EXTRACTOR_NAME, analysis_status="OK"
+        )
+        search_engine_similarity_subquery = SoundAnalysis.objects.filter(
+            sound=OuterRef("id"), analyzer=settings.SEARCH_ENGINE_DEFAULT_SIMILARITY_ANALYZER, analysis_status="OK"
+        )
+
+        qs = self.select_related(
+            'user', 'user__profile', 'license', 'ticket', 'pack', 'geotag'
+            ).annotate(
+                username=F("user__username"),
```
alastair, line 484: these annotate values are the values that the old bulk query returned. This makes it easy to run the code as-is without changing it to reference the related objects. We can slowly move them out

sounds/tests/test_manager.py
```diff
@@ -55,11 +43,11 @@ def setUp(self):
 
     def test_bulk_query_id_num_queries(self):
 
-        # Check that all fields for each sound are retrieved with one query
-        with self.assertNumQueries(1):
+        # Check that all fields for each sound are retrieved with one query + one for the analyzers
+        with self.assertNumQueries(2):
             for sound in Sound.objects.bulk_query_id(sound_ids=self.sound_ids, include_analyzers_output=True):
                 for field in self.fields_to_check_bulk_query_id:
```
alastair, line 49: This test is now missing a bit now that we've removed all of the "extra fields". There's no way to check that accessing a related item will not perform another query. We should extend it with a list of foreign keys and many-to-many relations that we should access to ensure that we stay with the same number of queries

sounds/models.py
alastair: removed most of them!

PR #1865, opened 2025-03-04
Make old uname redirect decorator not raise 404

ffont:
**Description**
Our very appropriately named `redirect_if_old_username_or_404` decorator is used in all views that are triggered by URL patterns containing usernames (e.g., /people/UNAME/sounds/1234). This is incredibly nice, but:
* The decorator can raise exceptions if username URL part is malicious (easily solved by catching exceptions) 
* The decorator might be hiding the logic of some of our tests: if we expect a test to return 404 for whatever reason which is not a non-existing username, and yet we pass a non-existing username, then we would think the view logic correctly returns a 404 but it only does it for different reasons. Solution is to not make our decorator raise a 404 (and rename it), and pass this job to the views. That would be less magic and easier to understand. We will need to add a method like `get_parameter_user(request)` that raises `Http404` if `request.parameter_user is None` (which can happen if the decorator sets it this way because it did not find a suitable user).

This PR currently only modified the decorator to not raise 404, but the solutions to the two points above are still pending, and also a rename.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -86,14 +86,14 @@ account a number of things:
 * In case a user object exists, is the user marked as having been deleted? (check `user.profile.is_anonymized_user`) 
 * In case there's no `User` object with such username, is there any `OldUsername` object which maps the username in the
   URL with a `User` object in DB?
-  
-To deal with these checks, we use the `utils.username.raise_404_if_user_is_deleted` and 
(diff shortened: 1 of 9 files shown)


alastair:
I renamed this method and fixed another bug in a related decorator. All of the tests still pass, so I think we can just go ahead and merge it

Issue #1870, opened 2025-03-07
Not possible to use negation operator in search filter


ffont:
It is currently not possible to use negation operator in search filters because our code in `search_filter_make_intersection` will add `+` signs to the filter terms and that results in a syntax like `-+tag:guitar` which is not valid for Solr. We should update the method to allow for this type of filtering and also add tests for that. This is specially relevant in the API, where some users try to do more advanced filters.

ffont:
Hi Trevor, thanks for your message and for offering to contribute. I think this issue probably has a rather easy fix, but it requires some rather advanced knowledge about how Freesound works, how it communicates to Solr and how search engine tests should be run. There is some information in our DEVELOPERS file. The summary is that we pre-process search filters before sending them to Solr, and some time ago we introduced a change in which we would add "+" sign before each filter term (because Solr changed some default behavior in a backwards incompatible way). This code we made does not work well when a filter term already has the "-" sign on it. In that case we should not add the "+". You could start setting up the local Freesound installation and trying to run the search engine tests. Then add a test which uses a filter with a  negation operator. That test should fail. It might be a big hard to follow the code to find where these "+" signs are added, let me know if you need assistance. Thanks!

trevorspecht:
Thank you Frederic, and I appreciate the offer for help. I'll let you know if I reach a point where I need it.

One thing I found when going through the readme development setup instructions is that one file it says to download from Google is missing: [Freesound development similarity index](https://drive.google.com/file/d/1ydJUUXbQZbHrva4UZd3C05wDcOXI7v1m/view?usp=sharing)

I don't know how important that is, but I'll keep going without it for now.

Bomme:
Hi @trevorspecht 
Did the following step run successfully for you?

> 7. Build the base Freesound Docker image
> 
>        make -C docker

trevorspecht:
@Bomme yes that step worked for me.

Bomme:
I think I found the source of confusion and just made a fix on `master`.

Explanation:
On executing `make -C docker` it would only build the `py2` image because that was chosen by Make if no target is explicitly defined. Before `all` was chosen as the target which would build both `py2` and `py3`. With the latest commit it should work again.

trevorspecht:
@Bomme thank you, this fixed the problem!

trevorspecht:
@Bomme sorry to keep tagging you. I am able to run everything locally now and I have made a local branch and added a failing test that will stop failing once I fix the query filter negation issue.

[...]

PR #1895, opened 2025-03-26
Not possible to use negation operator in search filter

trevorspecht:
**Issue(s)**
https://github.com/MTG/freesound/issues/1870

**Description**
Bug fix and tests to allow negation operator in queries
Adds an additional regex to remove the extraneous '+' when the user includes a '-' for negation filtering
Adds two tests for single and multiple query filters

**Deployment steps**:
None - just make sure all tests pass, specifically `freesound/utils/search/backends/tests/test_solr555pysolr.py`

### DIFF
--- a/utils/search/backends/solr555pysolr.py
+++ b/utils/search/backends/solr555pysolr.py
@@ -408,7 +408,8 @@ class Solr555PySolrSearchEngine(SearchEngineBase):
         # that allows us to define our own filter syntax and then represent filters as some intermediate structure that can later
         # be converted to valid lucene/dismax syntax.
         query_filter = re.sub(r'\b([a-zA-Z_]+:)', r'+\1', query_filter)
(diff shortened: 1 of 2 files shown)


ffont:
Thanks a lot for the PR, we'll have a look soon. Is it ready for review or are you planning to add something else (just mentioning that because of it being marked as "draft"). Thanks again!

trevorspecht:
@ffont this is just a draft, I have only added one test so far, and that test fails because of the bug. I still need to fix the bug so the test passes and possibly add more tests. 

I’ll try to not post draft PRs in the future so I don’t grab your attention before everything is complete and tested.

trevorspecht:
Final changes included, all tests including newly added tests are passing.
@ffont ready for review!

ffont:
Thanks for the contribution, fix works great. I'll merge now and deploy later today.

PR #1907, opened 2025-05-08
Minimal Ruff lint

Bomme:
This pull request introduces a minimal Python linting workflow using Ruff.

### DIFF
--- a/.github/workflows/python-lint.yml
+++ b/.github/workflows/python-lint.yml
@@ -1,0 +1,16 @@
+name: Python Lint
+
+on:
+  push:
+    branches: ["master"]
+  pull_request:
+    branches: ["master"]
+
+jobs:
+  lint:
+    name: Run ruff check
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/ruff-action@v3
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,19 @@
+[project]
+requires-python = ">=3.10"
+
+[tool.ruff]
+line-length = 120
+exclude = ["**/migrations", "_docs/**/*"]
+
+[tool.ruff.lint]
+select = [
+    "E",  # Pycodestyle Errors
+    "F6",  # Pyflakes
+    "F7",
+    "PLE",  # Pylint Error
+]
+ignore = ["E501", "E7", "E4"]
+
 [tool.pytest.ini_options]
 DJANGO_SETTINGS_MODULE = "freesound.test_settings"
 python_files = "test*.py"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [project]
-requires-python = ">=3.10"
+requires-python = "==3.10"
 
 [tool.ruff]
 line-length = 120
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -1,3 +1,19 @@
+[project]
+requires-python = ">=3.10"
```
alastair, line 2: add a `<` selector too, to prevent this from eagerly selecting a version of python that we don't currently support

pyproject.toml
```diff
@@ -1,3 +1,19 @@
...
+exclude = ["**/migrations", "_docs/**/*"]
+
+[tool.ruff.lint]
+select = [
+    "E",  # Pycodestyle Errors
+    "F6",  # Pyflakes
+    "F7",
+    "PLE",  # Pylint Error
+]
+ignore = ["E501", "E7", "E4"]
```
alastair, line 15: wow, pretty amazing that you can get to 0 errors with just a few ignores. Good start!

alastair: fyi, I recently migrated a flake8 django project to ruff and it uses these rules:
```
# Rules: https://docs.astral.sh/ruff/rules
# I: import order
# S: flake8-bandit
# B: flake8-bugbear
# A: flake8-builtins
# T10: flake8-debugger
# DJ: flake8-django
# ERA: eradicate
# G: flake8-logging-format
# PIE: flake8-pie
# T20: flake8-print
select = ["A", "B", "E4", "E9", "F", "I", "S", "T10", "DJ", "ERA", "G", "PIE", "T20"]
ignore = ["DJ012", "T201", "B904", "G004", "S311", "S308"]
```

.github/workflows/python-lint.yml
```diff
@@ -0,0 +1,16 @@
...
+    branches: ["master"]
+
+jobs:
+  lint:
+    name: Run ruff check
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/ruff-action@v3
```
alastair, line 16: just leaving this here: Here is how I check linting and formatting in a single action:

```
    steps:
      - uses: actions/checkout@v4
      - uses: astral-sh/ruff-action@v3
        with:
          args: "--version"
      - run: ruff check
      - run: ruff format --check --diff
```

Note that you need to set the dummy `args: --version` to make the action do a noop at setup before it runs your 2 checks

alastair (commented):
looks good, it's got 0 warnings so let's just go ahead and add it!

In fs-deploy we already have ruff set up on pre-commit, it'd be good to do the same here for the linter: https://github.com/MTG/freesound-deploy/blob/master/.pre-commit-config.yaml

Note that the id will be `ruff` not `ruff-format` - see the docs at https://github.com/astral-sh/ruff-pre-commit

pyproject.toml
Bomme: ah, good point, I mostly added it because it also serves as a hint to ruff to guide it in some of the linters.

Any preferences on how strict we want the requirement to be? [Django's supported versions](https://docs.djangoproject.com/en/5.2/faq/install/#what-python-version-can-i-use-with-django), i.e. `<3.13`? `<4.0`?

pyproject.toml
Bomme: I really also want flake8-django
https://docs.astral.sh/ruff/rules/#flake8-django-dj

PR #1911, opened 2025-05-23
Add AI preference options

ffont:
**Description**
Adds the "AI preference" option in the user profile so users can express their AI preferences. This is only displayed for users that have sounds. That  preference can also be included as a sound serializer field in the API, and as a user serializer field in the API.

Also adds some help text and functions in the sound describe form to add GenAI tag for sounds generated with generative AI.

**Deployment steps**:
* Migrate DB. 
* Update FAQ section with the meaning of the AI preference options
* Reindex!
* Email uploaders (?)


accounts/forms.py
```diff
@@ -396,6 +410,11 @@ def __init__(self, request, *args, **kwargs):
         self.fields['sound_signature'].widget.attrs['class'] = 'unsecure-image-check'
         self.fields['is_adult'].widget.attrs['class'] = 'bw-checkbox'
         self.fields['not_shown_in_online_users_list'].widget = forms.HiddenInput()
+        self.fields['ui_theme_preference'].label = "User Interface theme preference:"
+
+        # If user has no sounds, do not show the AI sound usage preference field
+        if request.user.profile.num_sounds == 0:
+            self.fields.pop('ai_sound_usage_preference')
```
alastair, line 417: does this mean that the process for a user to change their preference means that they first need to upload a sound and then go back to their profile? Can we make it clearer somehow?

accounts/models.py
```diff
@@ -641,7 +641,19 @@ def get_stats_for_profile_page(self):
...
-
+    
+    def get_ai_preference(self):
+        try:
+            return self.user.ai_preference.preference
+        except AIPreference.DoesNotExist:
+            # If no preference is set, return the default one
+            return AIPreference.DEFAULT_AI_PREFERENCE
+        
+    def set_ai_preference(self, preference_value):
```
alastair, line 652: isn't this just `create_or_update`?
I don't think the `.update` is correct, you need to filter on the user first

accounts/models.py
```diff
@@ -657,6 +669,18 @@ class GdprAcceptance(models.Model):
     date_accepted = models.DateTimeField(auto_now_add=True)
 
 
+class AIPreference(models.Model):
+    user = models.OneToOneField(User, on_delete=models.CASCADE, related_name="ai_preference")
+    date_updated = models.DateTimeField(auto_now=True)
+    AI_PREFERENCE_CHOICES = (
+        ("fr", "My sounds are used following Freesound's recommendations for interpreting Creative Commons licenses in a generative AI training context"),
```
alastair, line 676: a new way that we can do this is models.TextChoices: https://docs.djangoproject.com/en/5.2/ref/models/fields/#enumeration-types

Let's use descriptive names too, instead of these small 1-2 letter codes

accounts/forms.py
ffont: Aaaaand we could email uploaders about that. It will be ~38k emails, so should be doable...

accounts/models.py
alastair: note that TextChoices isn't a new field type, it's just a data type that makes it a bit easier to handle choices for a CharField, much more similar to a python enum.

Additionally, I was suggesting to not have a short column. In postgres there's no difference between a length 2, length 100, or unlimited length text field. Especially in our case where we only ever read the value. Let's make things easier on ourselves and make it readable when it's in the database.

There is a library which allows us to use a postgres `enum` in this column, which we could use if you want. I'm not too worried about it: https://github.com/django-commons/django-enum

PR #1913, opened 2025-05-29
Fix faceting for multi-valued solr fields

ffont:
**Description**
The fix that was released yesterday for tag faceting included some issues which are addressed in this PR. The whole thing is a bit confusing because we define a Freesound name for the field "tags", which is eventually translated to the specific field name used in our solr schema (which is "tag"), but that when used for faceting, it should be named "tagfacet". In this PR, I added support for "custom facet fieldnames", and it works but the resulting code is a bit complicated with all these name changes. The previous fix was simply renaming the solr field name from "tag" to "tagfacet", but this might cause issues in places where we actually need to use the name "tag" (like filtering or specifying query field weights). I'm not 100% about this, so let's discuss that once we have some time and maybe try to find a better solution.

A remaining issue is still faceting for multi-valued dynamic fields. We have to investigate and make that work.

### DIFF
--- a/tags/views.py
+++ b/tags/views.py
@@ -31,7 +31,6 @@ from search.views import search_view_helper
 from tags.models import Tag, FS1Tag
 from utils.search import SearchEngineException
 from utils.search.search_sounds import perform_search_engine_query
-from utils.search.backends.solr555pysolr import get_solr_fieldname, get_solr_fieldname_for_facet
 
 search_logger = logging.getLogger("search")
 
@@ -69,11 +68,10 @@ def tag_cloud(request):
             sentry_sdk.capture_exception(e)
             tvars.update({"error_text": "The search server could not be reached, please try again later."})
         else:
-            solr_tags_facet_fieldname = get_solr_fieldname_for_facet(get_solr_fieldname(settings.SEARCH_SOUNDS_FIELD_TAGS))
-            if solr_tags_facet_fieldname in results.facets:
+            if settings.SEARCH_SOUNDS_FIELD_TAGS in results.facets:
                 initial_tagcloud = [
                     dict(name=f[0], count=f[1], browse_url=reverse("tags", args=[f[0]]))
-                    for f in results.facets.get(solr_tags_facet_fieldname, [])
+                    for f in results.facets.get(settings.SEARCH_SOUNDS_FIELD_TAGS, [])
                 ]
                 cache.set("initial_tagcloud", initial_tagcloud, 60 * 60 * 12)  # cache for 12 hours
                 tvars.update({"initial_tagcloud": initial_tagcloud})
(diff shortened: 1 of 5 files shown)


tags/views.py
```diff
@@ -68,10 +69,11 @@ def tag_cloud(request):
             sentry_sdk.capture_exception(e)
             tvars.update({"error_text": "The search server could not be reached, please try again later."})
         else:
-            if settings.SEARCH_SOUNDS_FIELD_TAGS in results.facets:
+            solr_tags_facet_fieldname = get_solr_fieldname_for_facet(get_solr_fieldname(settings.SEARCH_SOUNDS_FIELD_TAGS))
```
ffont, line 72: this needs to be reworked!
the idea is that outside the solr-specific code, we always interact with search functions using the Freesound field names, not the solr names. In this case this would be "tags", just like some lines above the facets are defining using `settings.SEARCH_SOUNDS_FIELD_TAGS`. I think there's a bug here, because `results.facets` should be using `settings.SEARCH_SOUNDS_FIELD_TAGS` as key instead of "tagfacet". I have to investigate.

ffont:
This PR could be merged to fix the current tag search/faceting issues, but still is not addressing faceting for multi-valued dynamic fields. I guess the way to go is to define another type of dynamic field suitable for faceting, and then, for `_ls` (_list of strings_) dynamic fields that are included in faceting config, also create the complementary `_lsf` (_list of strings faceting_) when indexing. It also means that some logic should be added, similar to `FACET_FIELD_NAMES_MAP` but for the dynamic fields, so that a hypothetical field named `fsdsinet_tags_ls` would use `fsdsinet_tags_lsf` when faceting.

utils/search/backends/solr555pysolr.py
```diff
@@ -246,6 +266,10 @@ def convert_sound_to_search_engine_document(self, sound, include_analyzer_output
                         suffix = SOLR_DYNAMIC_FIELDS_SUFFIX_MAP.get(type(value), None)
                         if suffix:
                             document[f'{key}{suffix}'] = value
+                            if suffix == '_ls':
+                                # For dynamic fields of type "list of strings", we also need to set an extra field that
+                                # will be used for faceting
+                                document[f'{key}{suffix}{SOLR_DYNAMIC_FIELD_FACET_EXTRA_SUFFIX}'] = value
```
alastair, line 272: instead of "special casing" this, we could instead make `SOLR_DYNAMIC_FIELDS_SUFFIX_MAP` be a map of type -> list of suffixes. So for the case of list we just have the _ls and _ls_f suffixes.

ffont: Thanks for the suggestion. I think this would complicate the logic in other parts of the code where `SOLR_DYNAMIC_FIELDS_SUFFIX_MAP`. I think the way it is now is not that bad with the comment.

PR #1914, opened 2025-06-02
Remove unused FILE_UPLOAD_TEMP_DIR setting

alastair:
**Description**
This setting isn't used anywhere, so remove it and also remove any code which still uses it


ffont:
Hey, but we do use `FILE_UPLOAD_TEMP_DIR`, this is a Django setting. Files go there while uploading, and are then moved to /uploads/UID/xx when finished uploading. At least this is how it used to work unless something changed in Django...

Issue #1714, opened 2023-12-13
Reimplement similarity search using Solr


ffont:
We've been discussing for a long time how could we get rid of our custom similarity server (based on [gaia](https://github.com/MTG/gaia)) and move its current functionality to another service. Since last versions of Solr support nearest-neighbours (NN) search, we decided to move the similarity functionality to Solr. For this to happen, there are a number of considerations to take into account and steps to follow, some have already been done. 

We mainly use gaia for similarity search, but we also use it for some advanced functionalities of the API through which users can implement complex search filters based on low-level audio descriptors (e.g., filter by `pitch variance` or `bpm`), and also can specify a number of descriptors which establish a custom similarity metric which only takes into account these descriptors to sort the results of a query. When moving to Solr, some of these features will be lost, and also users will need to change their app implementations to achieve similar results. Therefore, some actions will need to be taken to notify users about these changes and also provide alternative ways to achieve similar results.

Our current implementation of similarity search also uses gaia to apply some transformations to the audio descriptors that we calculate and convert them into a 100-dimension normalised vector used for the NN queries. If we get rid of gaia, we will need to move this functionality somewhere else so that we can continue generating these sound vectors. Also, we want to take this opportunity to introduce audio embeddings computed using pre-trained deep learning models to our similarity system, so we can test state of the art approaches to similarity search.

Here are some steps to carry out for the reimplementation of the similarity service:

- [ ] Deprecate `content search` and `combined search` functionality from the API. Add a note in the API documentation and provide examples of how to implement an alternative solution. Send email to the API mailing list and also specifically notify users who use these API resources.
- [ ] Choose some audio descriptors that previously were only indexed in gaia to also add them in our current Solr index so that complex queries involving audio descriptors will still be possible. Currently we already index some descriptors generated by the AudioCommons analyzer, so basically we just need to add some configuration to list which descriptors from the FreesoundExtractor analyzer we want to get indexed in Solr. A full list of descriptors is here: https://freesound.org/apiv2/descriptors/ (note we should only choose one-dimensional descriptors).
- [ ] Update the current FreesoundExtractor audio analyzer so that it generates a 100-dimension vector compatible with the current ones being generated by gaia. We need to "extract" the current PCA and normalisation rules from our gaia dataset and reimplement them in the extractor so we can project points in the same space.
- [ ] Add new freesound analyzers that calculate embeddings and define a way to configure which ones to be loaed and ready for similarity search.
[...]

PR #1998, opened 2025-11-11
Refactor handling of audio descriptor in Solr, API and similarity

allholy:
**Issue(s)**
https://github.com/MTG/freesound/issues/1714

**Description**
This PR refactors the way audio descriptors (analyzer's output) is handled in Freesound, and provides new similarity features strictly based on Solr. It addressed many of the (previously) missing items of https://github.com/MTG/freesound/issues/1714, and moves us one step closer to completely removing the old similarity server (see below).

Before this PR, the output of different analyzers was stored as json/yaml files in disk, and some parts were also loaded to the DB through the `analysis_data` json field of `SoundAnalysis` objects. In `settings.ANALYZERS_CONFIGURATION` it was possible to specify which analyzer outputs should be loaded in the DB and also which ones should be indexed in Solr. This PR changes this behaviour a little bit, as now we will no longer load individual analyzer's ouput in the DB nor index them in Solr. Instead, we now define a `settings.CONSOLIDATED_AUDIO_DESCRIPTORS` setting that defines all the audio descriptors that we want available both in the DB and in Solr. It specifies from which analyzers the descriptors need to be loaded, and also allows to define some conditions under which a descriptor should be loaded (e.g. only for sounds in FX category) and data transformations. The `Sound` model now has a `Sound.consolidate_analysis` method that will create a new "meta" `SoundAnalysis` object of analyzer name `"consolidated"` which will store all descriptor's data following `settings.CONSOLIDATED_AUDIO_DESCRIPTORS`. These descriptors become available in Solr for filtering (there is also a way to indicate that a descriptor should not be in Solr if necessary), and also in the DB for returning their values in API calls (using the 'fields' parameter). All these descriptors are now treated as "first-class citizens" in the sound serializer and query filter syntax, so they can be used just like any other standard metadata field.

This PR also updates the way in which similarity works, with the definition of a `settings.SIMILARITY_SPACES` dictionary which allows to indicate which similarity spaces should be created and loaded in Solr, using similarity vectors created by audio analyzers.

These changes above have a particular effect on the API, as now the "text search" endpoint becomes a superpowered resource with which complex queries mixing metadata, audio descriptors and similarity distance can be carried out. This updated endpoint (renamed to simply "search", although the "old" search/text URL is still functional) is a replacement for the deprecated "content" and "combined" search methods. These are still functional after the deployment of this PR (using old Gaia similarity server), but will be removed in the near future. This PR also includes changes in the API to reflect all these updated features. 

After this commit, traditional gaia-based similarity is only used in the deptecated content search and combined search API endopints. A new PR should remove all no longer needed similarity code once we decide to finally remove these endpoints.

**Deployment steps**:
Build docs
Migrate DB
Clean analysis_data field of all `SoundAnalysis` objects
Run `consolidate_analysis` for all Sounds
Run `load_similarity_vectors` for all Sounds
Reindex-Solr

### DIFF
--- a/sounds/management/commands/export_audio_descriptor_stats.py
+++ b/sounds/management/commands/export_audio_descriptor_stats.py
@@ -1,0 +1,85 @@
+#
+# Freesound is (c) MUSIC TECHNOLOGY GROUP, UNIVERSITAT POMPEU FABRA
+#
+# Freesound is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Freesound is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#     See AUTHORS file.
+#
+
+import json
+import logging
+import os
+
+from django.conf import settings
+from django.utils import timezone
+
+from freesound.celery import get_queues_task_counts
+from sounds.models import SoundAnalysis
+from utils.management_commands import LoggingBaseCommand
+
+
+console_logger = logging.getLogger("console")
+
+
+class Command(LoggingBaseCommand):
+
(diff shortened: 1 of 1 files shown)


ffont:
I merged the PR for now. After deployment, we can make a new PR with the final updates to the API docs.
Thinks to be added to the next PR:
* API docs updates
* Add a section with hints for porting old Content Based/Combined Search code to new Search code?
* Document how the whole CONSOLIDATED_AUDIO_DESCRIPTORS and SIMILARITY_SPACES thing works in DEVELOPERS.md

PR #2006, opened 2025-11-26
Add some more lint checks

alastair:
Fix failing checks with few instances, but just bulk ignore major changes for now

This does have some code changes, so we should check tests but also maybe manually check these places

### DIFF
--- a/sounds/models.py
+++ b/sounds/models.py
@@ -2430,7 +2430,7 @@ class SoundAnalysis(models.Model):
             pass
         try:
             with open(self.analysis_filepath_base + ".yaml") as f:
-                return yaml.safe_load(f, Loader=yaml.cyaml.CSafeLoader)
+                return yaml.load(f, Loader=yaml.CSafeLoader)
         except Exception:
             pass
         return {}
@@ -2450,7 +2450,7 @@ class SoundAnalysis(models.Model):
             pass
         try:
             with open(analysis_filepath_base + ".yaml") as f:
-                return yaml.safe_load(f, Loader=yaml.cyaml.CSafeLoader)
+                return yaml.load(f, Loader=yaml.CSafeLoader)
         except Exception:
             pass
         return {}


sounds/models.py
```diff
@@ -2433,7 +2430,7 @@ def get_analysis_data_from_file(self):
             pass
         try:
             with open(self.analysis_filepath_base + ".yaml") as f:
-                return yaml.load(f, Loader=yaml.cyaml.CSafeLoader)
+                return yaml.safe_load(f, Loader=yaml.cyaml.CSafeLoader)
```
alastair: I did this because the linter said to, but haven't checked if it's actually valid - I guess our yaml files are data only, so we shouldn't have any problem switching between load or safe load. Note that we use load with a specific safe loader anyway, so is this actually an issue??

utils/search/backends/solr555pysolr.py
```diff
@@ -259,13 +259,13 @@ def convert_sound_to_search_engine_document(self, sound):
             document["type"] = "wav"
         else:
             document["type"] = sound.type
-        document["name"] = remove_control_chars(getattr(sound, "original_filename"))
-        document["description"] = remove_control_chars(getattr(sound, "description"))
-        document["tag"] = list({t.lower() for t in getattr(sound, "tag_array")})
+        document["name"] = remove_control_chars(sound.original_filename)
+        document["description"] = remove_control_chars(sound.description)
+        document["tag"] = list({t.lower() for t in sound.tag_array})
```
alastair: I'm unsure about these changes, as I mentioned in a previous comment django does magic with getattr vs model.x when it's a OneToOne, however these fields all look like actual fields in our sound object. I guess the original idea here was to return None if it didn't exist? We should try and do a full reindex to check that it works as expected

freesound/middleware.py
```diff
@@ -75,9 +72,7 @@ def __init__(self, get_response):
 
     def __call__(self, request):
         if request.user.is_authenticated and dont_redirect(request.get_full_path()):
-            try:
-                request.user.gdpracceptance
-            except GdprAcceptance.DoesNotExist:
```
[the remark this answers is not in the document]
ffont: so we switch to the `hasattr` syntax? there was also this syntax change for the user profile case in this PR

utils/search/backends/solr555pysolr.py
ffont: I did a reindex and it seems to work, but only with the 5000 local files. I think in the past we supported creating these documents from different types of objects, but since we added `bulk_query_solr` we know the type of object that we get and we could be sure that it has such fields. I would keep the changes as the code becomes more readable.

freesound/middleware.py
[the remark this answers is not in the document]
alastair: yes, I switched to hasattr because it prevents the linter warning "statement has no effect": while from a python perspectve this is kind of true, from a django perspective this statement (`request.user.gdpracceptance`) absolutely does something. we could have added a noqa comment but I chose to rewrite it to something that is identical in behaviour but has no warnings

sounds/models.py
alastair: yaml was historically very insecure, and there are many workarounds to make it safer. We can review these arguments in more detail to determine the best fix, I'll do this

apiv2/management/commands/basic_api_tests.py
```diff
@@ -53,9 +53,9 @@ def api_request(full_url, type="GET", post_data=None, auth="token", token=None):
         headers = {"Authorization": f"Token {token}"}
 
     if type == "GET":
-        r = requests.get(url, params=params, headers=headers)
+        r = requests.get(url, params=params, headers=headers, timeout=10)
     elif type == "POST":
-        r = requests.post(url, params=params, data=data, headers=headers)
+        r = requests.post(url, params=params, data=data, headers=headers, timeout=10)
```
Bomme: Was this intentionally added?

PR #2014, opened 2025-12-18
Add ty for type checking

alastair:
Experimenting with https://astral.sh/blog/ty

Merge #2024 first, which sets up uv.

Run checks by performing

    uv sync --dev
    uv run ty check

Note that our docker setup doesn't install dev dependencies, so I'm just doing it locally. We could make some updates to have a dev-specific environment if we want to be able to run the checks there.

Initially I found about 600-700 errors, but applied some easy fixes and have it down to about 500 now. I've not added any pre-commit or actions to check this yet as the number of warnings is still quite high.

We could consider enabling it only for a few packages (e.g. sounds or forums or accounts), or just slowly work down the list until we're happy with it.

Eventually it'd be good to also move our types from our docstrings to the function definitions, we were using docstrings because we were still using python 2.7 at the time that we started.

Some quick fixes surfaced by the ty type checker and django-stubs
* For a FormField, label should either be a string or None, not False
* help_text should always be a string, set to an empty string (the default) to show no help text

### DIFF
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -227,8 +227,8 @@ class RegistrationForm(forms.Form):
         super().__init__(*args, **kwargs)
 
         # Customize some placeholders and classes, remove labels and help texts
-        self.fields["username"].label = False
-        self.fields["username"].help_text = False
+        self.fields["username"].label = None
+        self.fields["username"].help_text = ""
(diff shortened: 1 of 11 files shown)


alastair:
Failing test is fixed in https://github.com/MTG/freesound/pull/2033

PR #2025, opened 2025-12-28
handle search query processor errors in map browse page

alastair:
**Issue(s)**
Fixes https://logserver.mtg.upf.edu/organizations/sentry/issues/3955/
Fixes https://logserver.mtg.upf.edu/organizations/sentry/issues/4730/

**Description**
We rewrite old query parameters of the form `duration:[1 TO 3]` to the new search interface. If the value of duration is invalid (e.g. `duration:1 TO 3` (no braces) or even `duration:[1 to 3]` (lower-case to) then the parser treats this as a string not a range. Only convert this parameter if it is valid, otherwise ignore it.

In the case that an invalid parameter is passed (`grouping_pack`), show an error message on the map page, as we do on the search page

Failing tests are fixed in https://github.com/MTG/freesound/pull/2033

### DIFF
--- a/geotags/views.py
+++ b/geotags/views.py
@@ -296,7 +296,7 @@ def for_pack(request, username, pack_id):
 def for_query(request):
     tvars = _get_geotags_query_params(request)
     request_parameters_string = request.get_full_path().split("?")[-1]
-    sqp = SearchQueryProcessor(request, request_params=request.GET)
+    sqp = SearchQueryProcessor(request)
     if sqp.errors:
         tvars.update(
             {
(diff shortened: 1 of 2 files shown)


alastair:
tested with
https://freesound.org/search/?q=test&f=duration:[1%20to%203]&s=Automatic+by+relevance&si_tags=0&si_name=0&si_description=0&si_packname=0&si_sound_id=0&si_username=0&d0=0&d1=*&ig=1&r=0&g=1&dp=0&cm=0&mm=0&ss=laion_clap&st=&cc=0&eap=0&se=0#

and it fails as expected (`&f=duration:[1 to 3]`) has lower-case to instead of TO

Issue #1988, opened 2025-08-27
Small forum moderation enhancement: account creation age / time ago & comment count & download count


qubodup:
Currently https://freesound.org/forum/moderate/ displays age of the post and number of sounds and posts. Adding account age (and comment count and number of sounds the user downloaded) would make it easier to make a quick informed decision.

Current layout:

> <img width="392" height="287" alt="Image" src="https://github.com/user-attachments/assets/76717986-a332-4356-af00-55d36b5b599f" />

Mockup:

> <img width="393" height="267" alt="Image" src="https://github.com/user-attachments/assets/55945c76-db47-4e53-a0cc-13cf3b035b6b" />

Issue #1850, opened 2025-02-18
Display 'last edited' timestamp on forum posts


Bomme:
Show a 'last edited' timestamps to inform users of post updates. This would help transparency in developing discussions.
Edit timestamps are already in the database.

PR #2029, opened 2025-12-28
Forum extra context

alastair:
**Issue(s)**
Fixes https://github.com/MTG/freesound/issues/1988
Fixes https://github.com/MTG/freesound/issues/1850

**Description**
On the forum moderation page, show additional information about the user who wrote this post

Screenshot 2025-12-28 at 4 23 15 PM

This data is only shown on the moderation page, but it's generic so we could show it to an admin/moderator in any location that we show a post. thoughts?

----

On forum posts, if the post was edited show the time that it was edited

Screenshot 2025-12-28 at 4 23 07 PM

### DIFF
--- a/forum/templatetags/display_forum_objects.py
+++ b/forum/templatetags/display_forum_objects.py
@@ -104,6 +104,8 @@ def display_post(
     edited_diff = None
     if post.modified and post.created:
         edited_diff = abs((post.modified - post.created).total_seconds())
+    current_user = context["request"].user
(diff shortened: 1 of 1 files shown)


alastair:
I've removed moderator info and replaced "downloaded sounds". I didn't change the "ago", because this is just using a standard freesound feature. However I've seen you ask for compact representations in another view - so we can probably make a compact version. When we get to that I'll try to remember, otherwise please open another ticket to remind us

Issue #1874, opened 2025-03-10
Introduce sound upload/describe session model


alastair:
We started looking into switching the session serializer to JSONSerializer, as the pickle serializer is no longer available in Django 5 (part of #1840)

In the sound describe workflow, we sometimes store Pack, License, and Sound objects in the session. This could be "easily" resolved by switching to use ids in the session instead of the object (and perhaps is what we should have done in the first place anyway)

However thinking about this in more detail, we have a "Session" linked to a bunch of related information, and we pass a session id around in the URL sometimes. Perhaps at this point we should introduce some models to represent the process of uploading and describing the sounds. We think that this could resolve a number of issues:

* Uploaded sounds could have a model. This allows users to upload multiple sounds with the same name, and makes it easier for us to do processing before the description stage.
* The describe process could have a model. This allows users to more easily store in-process descriptions and come back to it later. It should also be able to be shared over different computers, and we can introduce more access controls to ensure that only the uploader of a sound describes it.

PR #2048, opened 2026-01-15
WIP: Create PendingSound and PendingSoundAnalysis models for upload process

alastair:
**Issue(s)**
#1874

**Description**
Instead of using filenames, create a specific model for a user uploading a file. this makes it easier to process files and keep track of them, in addition allowing a single user to upload multiple files with the same name


sounds/models.py
```diff
@@ -77,6 +77,52 @@
...
+    sound = models.OneToOneField(
+        "Sound", null=True, blank=True, on_delete=models.SET_NULL, related_name="pending_sound"
+    )
+    status = models.CharField(max_length=2, choices=Status.choices, default=Status.UPLOADED)
+
+    def __str__(self):
+        return f"{self.filename} uploaded by {self.user} at {self.created} (Status={self.status})"
+
+
+class PendingSoundAnalysis(models.Model):
```
alastair: this is a "processing", not analysis.

See if we can take as much of this pre-processed data and reuse it instead of re-processing after sound description

PR #2055, opened 2026-02-02
Move frontend build to a separate build stage

alastair:
Allows us to build in parallel, speeding up build process Removes node from our main runtime image


ffont:
That looks really nice @alastair ! Although the build is failing with something uv related?

alastair:
yeah, this is a common build issue with pysndfile, I keep running into it. let's see if we can fix it once and for all

alastair:
this is building now, I upgraded to the latest pysndfile.
There's a bug in their build script, but I submitted a patch, hopefully we won't need our workaround in the future. We can wait until next week to release this, in case some unexpected issue comes up with processing due to the upgrade.

PR #2082, opened 2026-04-09
Replace CDN Redis map with nginx secure_link signed URLs

alastair:
We used to use a random uuid for a download filename. this was to hide the original filename on disk and prevent people from bulk downloading via the cdn (need to use the download url to resolve a sound to its random uuid). However this required the use of the redis map, which added extra complexity when moving redis from place to place (we couldn't easily recreate the map)

CDN downloads now use time-limited HMAC-signed URLs validated by nginx's secure_link module
Still keep symlinks to hide the soundid_userid.ext filename that we store sounds as, but create them immediately on sound create instead of having a separate upload command to do it.
Include a backfill management command for existing sounds.

**Issue(s)**

**Description**

**Deployment steps**:


alastair:
local_settings.py:

```py
CDN_SECURE_LINK_SECRET = "shared secret"
USE_CDN_FOR_DOWNLOADS = True
```

  /etc/nginx/cdn_secret.conf:
```
  set $cdn_secret "THE_SHARED_SECRET_MATCHING_DJANGO";
```

cdn server block
```
  include /etc/nginx/cdn_secret.conf;

  location /sounds/ {
      secure_link $arg_md5,$arg_expires;
      secure_link_md5 "$secure_link_expires$uri$cdn_secret";

      if ($secure_link = "") {
          return 403;
      }
      if ($secure_link = "0") {
          return 410;
      }

      alias /data/fsweb/cdn_sounds/;
      add_header Content-Disposition 'attachment; filename="$arg_filename"';
      add_header Content-Type application/octet-stream;

      limit_rate_after 100m;
      limit_rate 800k;
  }
```

sounds/models.py
```diff
@@ -1865,6 +1865,10 @@ def on_delete_sound(sender, instance, **kwargs):
     instance.delete_from_indexes()
     instance.unlink_moderation_ticket()
 
+    from utils.cdn import delete_cdn_symlink
+
+    delete_cdn_symlink(instance)
```
ffont, line 1870: You (or claude) even thought of that, amazing!

alastair: not me!  (but we should move the import to the top of the file)

PR #2100, opened 2026-05-04
Skip avatar img in embed iframe when user has no avatar

alastair:
Prevent rendering `` in embed html


ffont:
Looks good to me :) Do you want to add something more to it?

Issue #1575, opened 2021-10-29
"Last post" property of threads not correctly re-set after spammers get deleted


ffont:
Visiting the forums page, it can be seen that some threads/forums do not show "last post" information. This is most likely because the last post was from a spammer that got deleted and then the last post is not re-set as it should.

quimmrc:
I have been looking into this issue and found out that not only the `last_post` attribute is missing in some threads but also the number of replies does not match the actual number of replies shown within the thread (usually: forum page gives _x_ replies for a thread, and thread has _x+1_ replies ). 

Since replies are computed directly using `num_posts`, I guess that this attribute is not properly updated when certain posts get deleted (sth similar probably happens with last post).

To inspect this issue properly, I have listed the ways in which an existing post can be deleted and I'd like to know if there are any missing:

1. Author deletes his/her own post.
2. Moderator deletes a post (which had initially been assigned a moderation_state = OK by default).
3. Moderator deletes user. In here two things can happen. User is anonymized (its posts keep existing with an anonymized username), or user gets erased entirely from the DB (I think this case might be the one giving problems).

Am I forgetting any other way in which a post can be deleted? The two first options seem to work fine.

quimmrc:
So, when a moderator deletes a user it's completely removed from the DB, and so do its posts due to being deleted on cascade according to the _author_ attribute in the Post class definition. 

I tried to modify the post query in `set_last_post` method adding to select only posts with existing authors, but it does not seem to do anything since deleted user posts' no longer exist.

In all the listed cases above, the post deletion is done with the same Django signal `update_thread_on_post_delete` (which works fine). In conclusion: I am a bit lost.

ffont:
I think the issue is that the `update_thread_on_post_delete` signal might not be called when deleting a user with `user.delete()` because these signal are not called when objects are batch deleted in cascade?

Maybe the best thing for this is to start making a unit test to reproduce the issue?

PR #2154, opened 2026-07-06
Fix Pagination + filtering in search page can return unexpected results

### LINKED ISSUE #2137: Pagination + filtering in search page can return unexpected results
If user is browsing page 2 of the results of a query and then decides to add a facet filter, the new query loads at page 2 as well, which can result in empty results if the new results set only has one page. We should think in which occasions the page number should be reset to 1 (and do it!). I imagine this should be every time the set of results is changed because of filters or other query parameters.

kashyap-harshit:
**Issue(s)**
https://github.com/MTG/freesound/issues/2137

**Description**
Fixes #2137 

Before: 

https://github.com/user-attachments/assets/058882d1-465d-410b-9191-4215a473da3c

After:

https://github.com/user-attachments/assets/57db522a-8d68-454a-8ff3-2c52beddae3a

Details:

### Problem

If a user is browsing page 2 of a search result set and then adds a facet filter (or removes an existing one, or changes the selected cluster), the new query loads while still on page 2. The filtered result set often has fewer pages, so the user can land on an out-of-range page that shows no results, even though matching sounds exist.

### Root cause

Facet and filter links are generated by `SearchQueryProcessor.get_url()`. That method rebuilds the URL from all currently active search options, and `page` is one of those options. When you are on page 2, `page=2` is a non-default, in-request value, so it gets copied into every generated URL; this includes the "add filter" and "remove filter" links. Following such a link re-runs the modified query but keeps the stale page number.

Pagination links themselves are built by a separate mechanism (`bw_paginator`, using the request's GET params), and normal form submissions like sort, query, and advanced options already reset the page because the search form has no hidden `page` field. So the bug was isolated to `get_url()`.

### Fix

`get_url()` now skips the `page` option when rebuilding the URL. Every caller of it (facet add-filter links, active-filter remove links, and cluster-change links) produces a different result set where the old page number is no longer meaningful, so the generated URL always points to page 1. The current query itselrams()` still applies `current_page`, and paginationis untouched.

### Changes
- `utils/search/search_query_processor.py`: skip the `page` option in the `get_url()` option loop.
- `utils/tests/test_search_query_processor.py`: update two assertions that expected `page` to be echoed back into `get_url()`, and add coverage that adding or removing a filter from page

### Testing

`utils/tests/test_search_query_processor.py` passes (8/8). Example of the added assertion:

```python
sqp, _ = self.run_fake_search_query_processor(params={"q": "test", "page": "2"})
self.assertEqual(sqp.get_url(add_filters=['tag:"tag1"']), "/search/?q=test&f=tag%3A%22tag1%22")
```

Note that `page=2` is gone from the output while `q=test` is preserved.

**Deployment steps**:
None

### DIFF
--- a/utils/search/search_query_processor.py
+++ b/utils/search/search_query_processor.py
@@ -796,8 +796,13 @@ class SearchQueryProcessor:
             base_url = reverse("sounds-search")
 
         # Add query parameters from search options
+        # Skipping the page option while iterating
+
+
         parameters_to_add = {}
-        for option in self.options.values():
+        for option_name, option in self.options.items():
+            if option_name == "page":
+                continue
             if option.set_in_request and not option.is_default_value:
                 params_for_url = option.as_URL_params()
                 if params_for_url is not None:
--- a/utils/tests/test_search_query_processor.py
+++ b/utils/tests/test_search_query_processor.py
@@ -103,10 +103,11 @@ class SearchQueryProcessorTests(TestCase):
         )
         self.assertGetUrlAsExpected(sqp, url)
 
-        # With page number specified
+        # With page number specified.
+        # It should now reset the page number if having a filter (see get_url)
         sqp, url = self.run_fake_search_query_processor(params={"page": "3"})
         self.assertExpectedParams(sqp.as_query_params(), {"current_page": 3})
-        self.assertGetUrlAsExpected(sqp, url)
+        self.assertGetUrlAsExpected(sqp, url.replace("page=3", ""))
 
         # With "search in" options specified
         sqp, url = self.run_fake_search_query_processor(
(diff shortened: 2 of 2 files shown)


kashyap-harshit:
@ffont Please let me know if I need to change or fix something on the PR or Code.
Thanks :)

ffont:
Hi @kashyap-harshit , thanks for the contribution!
This looks like a possible solution, although this PR just came it at the very same time that a more general rework of pagination was happening, which also solves that issue (among other things). See the newer PR here https://github.com/MTG/freesound/pull/2156. Maybe you could test that pagination works as expected with that newer PR (you should test search page, forum search and API?) instead, and if you assess it works as expected then I'll merge the other one. Thanks!

kashyap-harshit:
Hey, @ffont No worries. As I can see the other PR is merged, let me know if any other issues are there. Happy to help. 
Thanks :)

ffont:
Ups, looks like @alastair merged the PR, but I was planning to keep it open until you tested and made sure it works as expected. Let's close this one anyway. You could look at this one: https://github.com/MTG/freesound/issues/1575
Also, if interested, I have another more DSP-related task, that would involve the re-implementation of a couple of Python descriptors using MTG's Essentia library. Let me know if interested.

kashyap-harshit:
Hey, 
Sure, I will try to recreate the #1575 issue and let you know. 
And about the DSP related task, yes I would be happy to help :) it sounds interesting as well.

PR #2159, opened 2026-07-10
Templatetags kwargs

alastair:
**Description**
I noticed that it was getting difficult to correctly update all template tags every time we changed default values or added a new parameter.

Now, for any of our template tags that have a default parameter, require that they are explicitly defined when calling from the template


ffont (approved):
I'm both excited and frightened by this PR. Looks good on paper anyway :)

PR #2160, opened 2026-07-13
enforce a daily download limit

alastair:
**Description**
Very infrequently we see people try to download all of Freesound via the download buttons on each sound. Stop this from happening.

Count the number of downloads that a user makes over a day. Align all users to UTC for simplicity. Use Redis to store the count. Count Sounds, Packs, Collections, and Bookmark Category downloads as 1 download each time.
Once a user goes over the limit, don't show a download link, and prevent the direct download link from providing the file (HTTP 429).

Set an "in progress" download key when the user starts a download, and don't count again if they access the URL when the download sentinel is active - this prevents download managers that request different ranges from counting multiple times.

Add the same guard to Collection downloads (it was missing)

The In progress download key in redis was previously set, but we also checked the http Range header to see if this was a "new" or "continued" download. Remove the Range header entirely, because the redis check performs pretty much the same check and we were using it anyway.

There was one bug where a user could maliciously request `Range: 0-` in order to circumvent the check, although we expect that this probably never happened.
The new check does mean that if there is a large file (takes longer than 5 minutes to download) and the client makes multiple requests to download it, that we'll count this twice. We'll check this in a few weeks but don't expect that it'll cause an issue either.

**Deployment steps**:
Requires the "abuse" cache to be added in prod. We'll add it to volatile.

### DIFF
--- a/templates/collections/display_collection.html
+++ b/templates/collections/display_collection.html
@@ -34,7 +34,7 @@
             </div>
             {% if collection.num_downloads %}
             <div class="h-spacing-left-1" title="{{ collection.num_downloads|bw_intcomma }} download{{ collection.num_downloads|pluralize }}">
-                <a href="{{ collection.download_url }}" class="bw-link--grey-light">
+                <a href="javascript:void(0)" data-toggle="modal-default" data-modal-content-url="{% url 'collection-downloaders' collection.id collection.url_kwargs.collection_name %}?ajax=1" class="bw-link--grey-light">
                     <span class="bw-icon-download" style="font-size:90%;"></span> {{ collection.num_downloads|formatnumber }}
                 </a>
             </div>
--- a/utils/download_limit.py
+++ b/utils/download_limit.py
@@ -1,0 +1,123 @@
+import datetime
+import enum
+import logging
+
+from django.conf import settings
+from django.core.cache import cache, caches
+from django.db import transaction
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import render
+
+logger = logging.getLogger("web")
+
+DOWNLOAD_LIMIT_CACHE_ALIAS = "abuse"
+DOWNLOAD_LIMIT_KEY_PREFIX = "downloadlimit"
+DOWNLOAD_LIMIT_KEY_TTL = 60 * 60 * 48  # 48 hours
+
+# TTL of the per-download "in progress" sentinel (see count_download_and_set_sentinel). Long enough to
+# span a single (possibly multi-part / slow) download; short enough that it doesn't persist
+# between genuinely separate downloads of the same object.
+DOWNLOAD_LIMIT_SENTINEL_TTL = 60 * 5  # 5 minutes
+
+
+def _get_redis_client():
+    # Native redis-py client from the cache backend. We use the raw client so a single
+    # pipelined INCR+EXPIRE creates-and-bumps in one round-trip (the backend's incr()
+    # would raise on a missing key).
+    return caches[DOWNLOAD_LIMIT_CACHE_ALIAS]._cache.get_client(write=True)
+
+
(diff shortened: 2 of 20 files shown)


utils/download_limit.py
```diff
@@ -0,0 +1,123 @@
...
+
+def new_download_blocked(request: HttpRequest, download_type: DownloadType, object_id: int) -> bool:
+    """Check if the user is allowed to download this item.
+
+    The download is blocked if it's a new download and if the user is over their daily limit.
+    see ``register_download`` for a description of what new download means.
+    """
+    is_new_download = cache.get(_sentinel_key(download_type, object_id, request.user.id), None) is None
+    return is_new_download and download_limit_reached(request.user.id)
+
```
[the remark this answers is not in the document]
alastair: not in prod

utils/download_limit.py
```diff
@@ -0,0 +1,123 @@
...
+        count, _ = pipe.execute()
+        return int(count)
+    except Exception:  # noqa
+        # Intentionally fail open on redis error
+        logger.warning("Could not increment daily download count for user %s", user_id, exc_info=True)
+        return 0
+
+
+def download_limit_reached(user_id):
+    return get_daily_download_count(user_id) >= settings.MAX_DOWNLOADS_PER_DAY
```
ffont, line 64: Shouldn't we skip download limits for sound uploaders and possibly some other cases? Although you already mentioned the threshold is set well below any "human" download activity

fscollections/templatetags/display_collections.py
```diff
@@ -46,5 +46,6 @@ def display_collection(context, collection):
         "header_sounds": header_sounds,
         "request": request,
         "show_visibility": show_visibility,
+        "download_limit_reached": download_limit_reached,
```
ffont: I don't think we need here as one should not be able to download collection from the "display collection" template, only from the collection page (just like you can't do it for sounds and packs). If the download functionality is currently available from such template, then this is something that I oversaw when reviewing collections PR(s). Let's remove it.

templates/collections/display_collection.html
```diff
@@ -34,9 +34,13 @@ <h5 class="ellipsis">
             </div>
             {% if collection.num_downloads %}
             <div class="h-spacing-left-1" title="{{ collection.num_downloads|bw_intcomma }} download{{ collection.num_downloads|pluralize }}">
+                {% if download_limit_reached %}
```
ffont: Just like Sounds and Packs, this button should open a modal to show who downloaded the collection, not trigger the actual download. Maybe disable the functionality and open an issue for that? (or implement it if easy, I guess it should be :))

ffont (approved):
New logic looks much cleaner in code. I added a couple of relevant comments though :)

fscollections/templatetags/display_collections.py
alastair: this is "needed" because every collection card shows a download button. We pass it through from this top level view so that we don't need to re-compute it (redis lookup) in each card's template tag code

templates/collections/display_collection.html
alastair: yes good idea!

utils/download_limit.py
alastair: I'd probably suggest that we leave this one for now and add another issue for it, because as we said it should never happen (99% of downloads in the last year were < 50/user/month and our limit is 200/day)

fscollections/templatetags/display_collections.py
alastair: OK, I understand the issue now. This was a bug during implementation - the logic was supposed to be "show list of downloaders", and in fact it only shows if num_downloads > 0, but the link goes to "download". it should link to the downloaders modal. will fix as part of this

PR #2161, opened 2026-07-13
Validate search query values before sending to solr

alastair:
**Description**
We found that scrapers/attackers construct search urls incorrectly
(they double-escape spaces) so we end up with filters like
`samplerate:22050+tag:"x"` which parse as `samplerate:22050+tag`
and causes solr to raise an error.

Validate filter strings and values before sending the result to solr.
If it looks invalid then directly return an error page instead of
trying the search.

Check for errors on search, tag, map bytearray, and clustering views

**Deployment steps**:
This should have a "real" effect on search results, I expect at least the following:
 - real number of queries to solr will go down if we drop the requests - we don't count the number of invalid responses but we could probably simulate it based on (requests)-(queries to solr)
 - we may be able to handle "more concurrent search requests", simply because we return a response quickly instead sending to solr: The nginx limit is based on "concurrent requests" so invalid searches will allow us to move on to handling the next search request faster
 - slightly possible that real queries will hit this and get an error page. Not sure the best way to track this - maybe we can look for "searches that have a real referrer and have an error", but I don't think there's a way that we can identify this case in our stats. we might want to make a prometheus counter.

### DIFF
--- a/utils/search/filter_validation.py
+++ b/utils/search/filter_validation.py
@@ -1,0 +1,89 @@
+import luqum.tree
+from luqum.parser import parser
+
+FIELD_TYPES_MAP = {
+    "samplerate": int,
+    "bitrate": int,
+    "bitdepth": int,
+    "channels": int,
+    "duration": float,
+}
+
+COMPLEX_EXPR_TYPES = (
+    luqum.tree.Range,
+    luqum.tree.OpenRange,
+    luqum.tree.From,
+    luqum.tree.To,
+    luqum.tree.FieldGroup,
+    luqum.tree.Boost,
+)
+
+
+def parse_filter(filter_string):
+    """Parse a Lucene-style filter string into its list of top-level nodes.
+    Empty string -> []. Raises luqum.exceptions.ParseError on malformed input.
+    """
+    if not filter_string:
+        return []
+    tree = parser.parse(filter_string)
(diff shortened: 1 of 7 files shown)


utils/search/filter_validation.py
```diff
@@ -0,0 +1,89 @@
+import luqum.tree
+from luqum.parser import parser
+
+FIELD_TYPES_MAP = {
```
ffont, line 4: We are starting to "hardcode" search fields information in several places (see https://github.com/MTG/freesound/blob/176ce1e6fbd21a80ae113bc6155022d71529694b/freesound/settings.py#L546). In some cases at the Solr level, in other cases at a more abstract "could be applied to any search backend" level. I wonder if we should start thinking of an alternative way of doing that. Probably one single dict in settings which also includes the types... For now maybe simply add a comment here.

ffont (approved):
I made a comment with potential changes but maybe for another PR. Also, wouldn't this be useful to avoid using our rather hacky code that compensates for the AND/OR filter terms default change that happened in solr some time ago? Remember we wrote some ugly code to add the + signs? Or did we already get rid of that? :S

utils/search/filter_validation.py
alastair: hmm, yeah. we could have a dict of dataclasses or something - with field name, type, default weights, facet values, etc. it should end up consolidating all search field settings into a single location

PR #2163, opened 2026-07-14
Don't store objects in the session when describing sounds

alastair:
Use only values which can be serialized by django's JSONSerializer session serializer (dict + basic types)

**Issue(s)**

**Description**

**Deployment steps**:


ffont:
I tried it and it seems to work fine, thanks @alastair !

PR #2166, opened 2026-07-20
 Fix 'Last post' property of threads not correctly re-set after spammers get deleted

kashyap-harshit:
**Issue(s)**
#1575 

**Description**
### Problem
When a spammer is deleted from the DB, their posts are cascade-deleted along
with their account. This left threads and forums with a blank "last post" and
an inflated reply count on the forum listing pages.

**Root cause:** the `post_delete` signal handler that repairs these cached
fields tried to update the spammer's own profile first. Since the profile was
already deleted in the cascade, that line raised `Profile.DoesNotExist`, which
was silently caught, so the code never reached the part that re-points
`last_post` and fixes `num_posts`.

### Changes

**1. Fix the signal handler (`forum/models.py`)**
- Update the author's profile count with a queryset `update()` so it's a no-op
  (instead of an exception) when the profile is already gone. The thread/forum
  repairs now always run.
- Clamp all `num_posts` decrements with `Greatest(..., 0)` so deletions can't
  crash on counts that are already out of sync.
- Remove the now-unreachable `except Profile.DoesNotExist` branch.
- Add regression tests covering both deletion paths (`user.delete()` and
  `Profile.delete_user(delete_user_object_from_db=True)`).

**2. Repair already-broken data (`update_last_thread_forum_posts` command)**
The fix above only prevents *new* corruption; rows already broken on production
stay broken. This extends the existing repair command to also recompute
`num_posts` (previously it only fixed `last_post`), and to scan all threads and
forums rather than only those with a null `last_post`. Run it once after
deploying to heal the historical data.

### Testing
- New tests fail on the old code and pass with the fix.
- Full `forum/` and `accounts/tests/test_user.py` suites pass.
- Verified the repair command end-to-end: manually corrupted a thread/forum,
  ran the command, confirmed `last_post` and `num_posts` were restored.

**Deployment steps**:
After deploying, run the forum repair command once to fix threads/forums whose
`last_post`/`num_posts` were already corrupted by this bug before the fix:

    python manage.py update_last_thread_forum_posts

(Run with `--dry` first to see how many rows would change without writing.)

### DIFF
--- a/forum/management/commands/update_last_thread_forum_posts.py
+++ b/forum/management/commands/update_last_thread_forum_posts.py
@@ -20,15 +20,15 @@
 
 import logging
 
-from forum.models import Forum, Thread
+from forum.models import Forum, Post, Thread
 from utils.management_commands import LoggingBaseCommand
 
 console_logger = logging.getLogger("console")
 
 
 class Command(LoggingBaseCommand):
-    help = """Updates last_post property of all Thread and Forum objects to make sure these
-    do not get out of sync."""
+    help = """Updates last_post and num_posts properties of all Thread and Forum objects to make
+    sure these do not get out of sync."""
 
     def add_arguments(self, parser):
         parser.add_argument(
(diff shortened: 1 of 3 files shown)


kashyap-harshit:
@ffont  This is ready for review. it fixes #1575 by correcting the signal handler and also adds a repair command for the already corrupted production data. Tests cover both deletion paths. Let me know if you'd like any changes. Thanks :)

ffont:
Thanks a lot for the PR, I reviewed and looks good. I'll merge!

kashyap-harshit:
Thank you so much! 
Also I am still interested on the DSP-related task. I would love to hear more about it :)

PR #2172, opened 2026-07-22
Experiment app and BST category validation 

allholy:
**Issue(s)**
None.

**Description**
Add experiment app and first experiment (category vallidation) for Freesound.

**Deployment steps**:
wip


templates/user_feedback/inline_category_validation.html
```diff
@@ -0,0 +1,89 @@
...
+<style>
+    /* Drop opt-out link in really small screens, where it may overlap with other buttons. */
+    @media (max-width: 800px) {
+        #categoryValidationBox form { display: none; }
+    }
+    /* Underline on hover to signal it is clickable. */
+    #categoryValidationBox form button:hover { text-decoration: underline; }
+</style>
+<div id="categoryValidationBox" class="padding-3 border-grey-light border-radius v-spacing-top-2 text-14 text-grey" style="position: relative; padding-bottom: 24px; padding-right: 34px;">
+    {# Info sheet: circled "i" pinned top-right; opens the self-contained modal below (sound.js -> activateModal). #}
```
allholy, line 14: **Information sheet** for the experiment: circled-i button top-right opens a self-contained modal `#categoryValidationInfoModal` (`:36`) via `activateModal` (`sound.js:64`). Review if we want it that way and content.

user_feedback/experiments.py
```diff
@@ -0,0 +1,168 @@
...
+        return {}
+
+    def is_throttled(self, request, **kwargs):
+        """True if we should NOT show it because of 'do not nag' rules.
+        Default: don't show again once the user has opted out or answered."""
+        if self.has_opted_out(request.user):
+            return True
+        return UserFeedback.objects.filter(user=request.user, experiment_id=self.experiment_id).exists()
+
+    def should_show(self, request, **kwargs):
```
allholy, line 65: We ensure it is only logged in users like that, plus all views are `@login_required` (`views.py:10,41`). Make sure we want this behavior and is implemented correctly

templates/user_feedback/inline_category_validation.html
```diff
@@ -0,0 +1,89 @@
+{% load bw_templatetags %}
+{# Inline "category validation" box; shown by sounds.views.sound only when the category_validation experiment is on. #}
+{# Neither answer saves on click: both open a modal where Send confirms it, so a misclick costs nothing. #}
+{# Bound in pages/sound.js (not the declarative data-toggle) because the modal needs an onLoaded hook to initialise the category field. #}
+<style>
+    /* Drop opt-out link in really small screens, where it may overlap with other buttons. */
+    @media (max-width: 800px) {
```
allholy, line 7: Generally, I hide opt-out for small screens, since it seemed more straightforward. I like it like that but let's review this decision. Also any comments on the size and position of the button? I would go smaller to avoid confusion but it is also nice like that.

freesound/settings.py
```diff
@@ -470,10 +462,19 @@ def load_broad_sound_taxonomy_from_csv(path):
...
     (key, value["name"]) for key, value in BROAD_SOUND_TAXONOMY.items() if "-" in key
 ]  # Second-level categories
 
+
+# -------------------------------------------------------------------------------
+# User feedback experiments (user_feedback app)
+
+# Fraction of eligible users shown each experiment: 0.0 = off, 1.0 = everyone.
+# local_settings.py overrides these for local testing.
+FEEDBACK_SAMPLE_RATES = {"category_validation": 0.0, "category_filter_feedback": 0.0}
```
allholy, line 485: You can turn on each experiment like that. 0 would not show anything to nobody,1 to every sound page visit of a user.
Btw it shows to every user only if logged in AND not throttled (already answered / opted out).
We can start with 0.1 to test reactions when deployed.

allholy:
Notes for review (27/07/26) : 

@ffont you review this one right? 

I left in-line comments for some things, e.g. information sheet, logged-in users, dont ask again button, and how to activate the experiment. 

Also one commit has the base of the second experiment, I stopped there since we have to yet decide details for this one, plus it is better to review the first one now. I will continue when everything is reviewed and addressed.

sounds/tests/test_sound.py
```diff
@@ -616,6 +616,7 @@ def setUp(self):
     def _get_sound_view_cache_keys(self):
         return [
             get_template_cache_key("bw_sound_page", self.sound.id),
+            get_template_cache_key("bw_sound_page_details", self.sound.id),
```
ffont, line 619: I guess this change (and the following one) correspond to an unrelated bug fix?

freesound/settings.py
ffont: This is interesting. I'll see at the end of the review, but I don't like that there are experiment names hardcoded here. I wonder if we should have a "Experiment" DB object...

freesound/static/bw-frontend/styles/molecules/modal.scss
```diff
@@ -149,3 +149,20 @@
     white-space: nowrap;
   }
 }
+
+#categoryValidationModal {
```
ffont, line 153: I don't like having experiment name hardcoded here. We should make the modals as generic as possible, and if some needs something special, turn this into a generic css class and use that class in the actual experiment modal template, but most do not add experiment-specific code where it is not strictly needed.

freesound/static/bw-frontend/src/pages/sound.js
```diff
@@ -17,6 +23,50 @@ const urlParams = new URLSearchParams(window.location.search);
...
+        const box = form.closest('#categoryValidationBox');
+        if (box) box.remove();
+        showToast("Got it, we won't ask you this again.");
+      },
+      () => showToast('Something went wrong, please try again.')
+    );
+  });
+});
+
+// Study info sheet: the button shows the self-contained modal already in the DOM.
```
ffont, line 63: These comments are strange :)
What is this doing?

freesound/static/bw-frontend/src/pages/sound.js
```diff
@@ -17,6 +23,50 @@ const urlParams = new URLSearchParams(window.location.search);
 
 prepareAfterDownloadSoundModals();
 
+// Bound here instead of the declarative data-toggle: the category field inside the
```
ffont, line 26: This experiment-related JS code should not be in the `sound.js` file. This code is binding experiment modals, we should have a generic experiment JS code similar to what we do for other elements like similarity modals. So in the `components` JS directory, there should be a `experiments.js` file which does the binding of modals in a similar way, finding existing elements with a particular tag, and using `data` properties if extra info needs to be passed like the experiment ID or any URL. So basically the same thing you are doing but make it fully experiment agnostic, and move it to `experiments.js` file.

sounds/views.py
```diff
@@ -302,6 +312,8 @@ def sound(request, username, sound_id):
         "is_explicit": is_explicit,  # if the sound should be shown blurred, already checks for adult profile
         "sizes": settings.IFRAME_PLAYER_SIZE,
         "min_num_ratings": settings.MIN_NUMBER_RATINGS,
+        "show_category_validation": show_category_validation,
+        "category_validation_form": category_validation_form,
```
ffont, line 321: I guess this is the form for the experiment right? I'm unsure about that as well. Is there a better way to implement this and further separate sound page from experiments? I need to test it and do more review before answering, but that's what concerns me.

user_feedback/experiments.py
ffont: I think we said we don't need to restrict to logged in users. Why do we only want to ask them? Even if that was the purpose for an experiment, maybe we want to be able to to it in some other...

user_feedback/experiments.py
```diff
@@ -0,0 +1,168 @@
...
+        return UserFeedback.objects.filter(
+            user=request.user, experiment_id=self.experiment_id, data__category=category
+        ).exists()
+
+
+# The registry: the single place experiments are listed. Add class + entry for a new experiment.
+EXPERIMENTS = {
+    CategoryValidation.experiment_id: CategoryValidation(),
+    CategoryFilterFeedback.experiment_id: CategoryFilterFeedback(),
+}
```
ffont, line 164: Maybe this list should go to settings.py, together with the sampling options

ffont (commented):
Thanks for the PR!
I think this is work in the good direction! I left some comments, but I want to do more testing and review before proposing a "final" solution for some of the issues I mention, specially about separation between individual pages and experiment code. I expect to be able ton provide more feedback in the next days.

freesound/static/bw-frontend/src/pages/sound.js
allholy: confirms that opt out happened (or not)

freesound/static/bw-frontend/styles/molecules/modal.scss
allholy: i could move this in the template, makes sense

freesound/settings.py
allholy: i agree in comments above to not hardcode things, but it seemed like this is the place to actually do it, if somewhere, it should be here? on the other hand an experiment object may be ok..

sounds/tests/test_sound.py
allholy: I will recheck later.

sounds/views.py
allholy: i think there was a way of separating them more. It seemed less straightforward, and settled here. I can rethink that.

user_feedback/experiments.py
allholy: let's see if we'll keep the settings one (and not replace with db object?), if so, I agree

freesound/static/bw-frontend/styles/molecules/modal.scss
ffont: Just to clarify, I don't mean moving the css, I mean using generic classes that do not depend on an experiment name.

freesound/settings.py
ffont: If things need to be hardcoded, this is the place. For now I think we can live with having all experiments config here, maybe using a more complex dict structure that also replaces this: https://github.com/MTG/freesound/pull/2172/changes#diff-86942cbe559863992c0b395f5ee74b3e61a5fc16eaedbd21c327f5861dde977fR161

Note that experiment classes will probably need to be defined with strings for the class names.

sounds/tests/test_sound.py
ffont: I think I now understand you need that because otherwise the experiment template code would fall in the middle of the cached part of the page. I don't like it but also I don't see a better way of doing this.

sounds/views.py
```diff
@@ -293,6 +294,15 @@ def sound(request, username, sound_id):
...
 
+    # Category-validation experiment
+    # The experiment object owns all the rules (auth, sampling, throttling).
+    # Wiring is kept explicit on purpose for now.
+    category_validation = get_experiment("category_validation")
+    show_category_validation = category_validation.should_show(request, sound=sound)
+    category_validation_form = (
+        category_validation.form_class(initial={"sound_id": sound.id}) if show_category_validation else None
+    )
+
```
ffont, line 305: As I said in the main comment, user feedback saving should happen through AJAX and in a different view (the submit view?) so we don't need to have any form here. We should only load the experiment object (like in `experiment = get_experiment("category_validation")`), and then pass this to the template, where we can call some `experiment.render_inline_html` method or something like that.

user_feedback/experiments.py
```diff
@@ -0,0 +1,168 @@
...
+            return True
+        if not category:
+            return True
+        return UserFeedback.objects.filter(
+            user=request.user, experiment_id=self.experiment_id, data__category=category
+        ).exists()
+
+
+# The registry: the single place experiments are listed. Add class + entry for a new experiment.
+EXPERIMENTS = {
```
ffont, line 161: Move this to `settings.py` as a single structure that hosts al experiments information and config

ffont (changes requested):
Hi!

I left some comments and I also tested it locally. I was unable to make the `CategoryFilterFeedback` to work (like I don't see a thing). But the other one did work. I think we can for now concentrate on this one, and implement the other one after this works.

In addition to the comments in code, here is some general feedback about the category validation experiment:

* consent forms are not needed (i think we discussed that), as we are already in the context of Freesound website and are covered by the terms. we're colleciting "user actions" data just like any other action in Freesound like a search query or anything. We can do this in accordance to our terms, so we don't need that sheet.

* when making the width small, the "don't ask again dissapears"

* when clicking yes, I would show a message to say "thanks" in the same space that the experiment is taking, but not show a modal asking for more comments as it is quite annoying. only show the modal when user clicks "no". 

* user feedback responses should be always saved via AJAX, not needing to reload the page. I'm a bit confused as I think this is sometimes the case but not always.

* there are other smaller UI concerns that can be addressed when the rest if fixed


There is some re-work needed on the experiment rendering side, but I would day the data model is fine for now. In the future it might be good to have a DB table for `Experiment`, but for now let's simply work with a dictionary in `settings.py` which has a list of available experiments, linking them to their corresponding classes and their sampling rate settings.

To re-work the rendering side of things the idea is to separate as much as possible existing freesound code (views, etc) from the individual experiments. So we should make as minimal changes as possible both in the view and in the template. Using the AJAX post forms will actually help in that, as we don't have the form object in the sound page view. Then having an "experiment" object class (like we already have) that can render "inline" html and can directly call that from the sound page template. All in all it is not so different from what you have already.

Once the category validation case is fully ok, you can move to the other one, which should be similar, maybe simpler as maybe it should not even have a modal

PR #2182, opened 2026-09-16
Save search query records to DB

ffont:
**Description**
This PR adds a new `SearchQuery` model that can be used to store information about every search query made to the database.  It also includes a command to `archive_search_queries` that will dump old `SearchQuery ` entries (older than a number of days that can be specified) into csv files and delete them from the DB. This is because we don't need to keep search query objects forever in the database, but we want to have a historical record.

This is part of an effort to save all research relevant data to the DB instead of relying on graylog so we can eventually get rid of it.

**Deployment steps**:
Migrate DB


ffont:
After discussing with @alastair , we realised that the number of DB rows per day would be really high if we want to store every individual query. Let's have another thought and see in which ways it would make sense to implement this functionality.

PR #2185, opened 2026-09-21
Save records of API sound/pack downloads in DB

ffont:
**Description**
So far we have not been storing information about API downloads in the database (only through logging). This PR adds support for this.

**Deployment steps**:
Migrate DB


apiv2/views.py
```diff
@@ -470,6 +470,11 @@ def get(self, request, *args, **kwargs):
         sound_path, sound_friendly_filename, sound_sendfile_url = prepare_sendfile_arguments_for_sound_download(sound)
         if not os.path.exists(sound_path):
             raise NotFoundException(resource=self)
+
+        if self.client is not None:
+            # Save download record in database
```
ffont, line 475: I'm unsure if we should do checks to avoid duplicate downloads like in the normal `sounds.views.sound_download` view. I guess we should because it works in the same way?

Also we could add already the same download redirect code as in the `sounds.views.sound_download` view and put it under a `USE_CDN_FOR_API_DOWNLOADS` setting so we are ready for the change.

